PR #79, opened 2020-03-02
Fix version printing to the crate version

### LINKED ISSUE #74: Fix version reporting from tremor commandline tools
**Problem**

Tremor CLI tool (tremor-script, tremor-query, tremor-tool) report old versions.

**Steps**

Run commands like `tremor-script --version`.

**Possible Solution(s)**

Version number is hardcoded in places like:
https://github.com/wayfair-tremor/tremor-runtime/blob/v0.7.3/tremor-script/src/main.rs#L70

Can use `CARGO_PKG_VERSION` environment variable to align the version there with that defined in Cargo.toml. Example:
https://github.com/wayfair-tremor/tremor-language-server/blob/v0.7.3/src/main.rs#L27

Licenser:
closes #74

### DIFF
--- a/tremor-query/src/main.rs
+++ b/tremor-query/src/main.rs
@@ -50,7 +50,7 @@ pub fn nanotime() -> u64 {
 #[allow(clippy::too_many_lines)]
 fn main() -> Result<()> {
     let matches = App::new("tremor-query")
-        .version("0.6.0")
+        .version(option_env!("CARGO_PKG_VERSION").unwrap_or(""))
         .about("Tremor interpreter")
         .arg(
             Arg::with_name("SCRIPT")
--- a/tremor-tool/src/cli.yaml
+++ b/tremor-tool/src/cli.yaml
@@ -1,4 +1,4 @@
-name: tremor-cli
+name: tremor-tool
 version: "1.0"
 author: The Tremor Team
 about: Tremor command line interface tool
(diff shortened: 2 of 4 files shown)


tremor-tool/src/cli.yaml
```diff
@@ -1,4 +1,4 @@
-name: tremor-cli
+name: tremor-tool
 version: "1.0"
```
anupdhml: does the version here change anything for the version determination?

tremor-query/src/main.rs
```diff
@@ -50,7 +50,7 @@ pub fn nanotime() -> u64 {
 #[allow(clippy::too_many_lines)]
 fn main() -> Result<()> {
     let matches = App::new("tremor-query")
-        .version("0.6.0")
+        .version(option_env!("CARGO_PKG_VERSION").unwrap_or(""))
```
anupdhml: nice! didn't know about option_env

tremor-tool/src/cli.yaml
Licenser: nope it has no function, good call we might as well kill it

Licenser:
> while we are here, should we also add a `version` flag to tremor-server?

we a;readu got that:
```
     Running `target/debug/tremor-server --version`
tremor version: 0.7.3
rd_kafka version: 0x000000ff, 1.3.0
tremor-runtime 0.7.3
```

PR #88, opened 2020-03-05
GitHub actions macos runner

darach:
Add support for Mac OS X build and test on github


darach:
Renaming actions causes stale ( but passing ) checks to stay incomplete - invoking god mode and rebase+merging

PR #131, opened 2020-03-23
Fixes for local docker image build

anupdhml:


Dockerfile.native
```diff
@@ -15,13 +15,14 @@ RUN apt-get update \
 ENV DEBIAN_FRONTEND=dialog
 
 COPY Cargo.* ./
-COPY .cargo ./.cargo
+
```
Licenser: If we do not copy .cargo in here it'll not build as `cpu-target=native` which was the intent of the secondary image.

anupdhml: good catch! fixed it now

Licenser: 👍

PR #132, opened 2020-03-24
Add features to switch allocators

Licenser:
add the features
* `snmalloc` (defaultly set)
* `jemalloc`
* `mimalloc`

for tremor-server


mjp41:
I hope this doesn't mean that `snmalloc` is not working out for you,.

Licenser:
Oh no, not at all! We added it for easier benchmarking the different allocators and to be able to use memory tracking w/ jemallocs leak checker. :)

Licenser:
My phrasing sucks. I mean: It doesn't mean at all that snmalloc isn't working out for us, it's working wonderfully.

PR #143, opened 2020-04-01
Fix typos in comments, error/warning messages, and corresponding tests

m0n0chr0m3:
I separated it from a future commit with technical changes to keep the reviewing effort reasonable, especially for that other commit.

This PR needs little to no technical expertise to review, the only parts that need more than just English language skills are:
- confirming that changes to the way error messages are produced, corresponds to changes in expected error messages in tests,
- ensuring that renaming a test case from `umatched_parenthesis` to `unmatched_parenthesis` doesn't break other tooling.


tremor-script/src/ast.rs
```diff
@@ -113,6 +113,10 @@ where
     operators: Vec<OperatorDecl<'script>>,
     scripts: Vec<ScriptDecl<'script>>,
     aggregates: Vec<InvokeAggrFn<'script>>,
+    // TODO: Users of the `warnings` field might be helped if `warnings` were a Set. Right now,
```
Licenser, line 116: Good catch :) 👍 for changing this to a set!

Licenser:
Sorry this takes so long, we messed up the CI. We're working on fixing it and will merge once we worked through it.

PR #164, opened 2020-04-10
Subrange-as-slice (+ bugfixes) in interpreter::resolve

m0n0chr0m3:
Applies the subrange-as-slice changes as first shown in https://github.com/m0n0chr0m3/tremor-runtime/commit/67a8201c960cb431efa14e1798681a550f82b32d, but also handles two previously unchecked error conditions, based on feedback on https://github.com/m0n0chr0m3/tremor-runtime/commit/1ce2d96945f0b5608064555b54fa1d12bc12a85c.

Concretely, this branch now offers the following changes vs. master:
- `tremor_script::interpreter::resolve` uses subslices (instead of `usize`s and `unsafe` indexing) to track the range of values;
- when an array index range's end is smaller than its start (e.g., `arr[5:3]`), an error is produced by Tremor Script (instead of crashing at runtime); and
- when an array index range's start is a negative number (e.g., `arr[-25:25]`), a reasonable error is produced (instead of "got integer but expected integer").
-  Additionally, some error messages (and their test cases) were changed:
    - array index out of bounds error now mentions the valid range; and
    - array index range out of bounds error had off-by-one error (as far as I can see...).
- Finally, 4 new test cases were added:
    - `empty-subslice` checks that a range whose start and end are the same, correctly yields an empty subslice;
    - `subslice-repeated` checks that taking a subslice of a subslice of a subslice of ... (x9) behaves as expected;
    - `subslice_end_lt_start` checks that index ranges whose end is smaller than its start, are properly reported as invalid (instead of crashing at runtime); and
    - `subslice_neg_start` checks that index ranges whose start is a negative number, are correctly identified as invalid.
      
Questions that need answers still:
 1. Do the new error messages capture what you were going for in https://github.com/m0n0chr0m3/tremor-runtime/commit/1ce2d96945f0b5608064555b54fa1d12bc12a85c#commitcomment-38283863?
 2. What is the ideal place for the code [extracting the base value to resolve from](https://github.com/m0n0chr0m3/tremor-runtime/commit/21bcf3be14a2d4060a9c76e1f687fe757f93af41#diff-6be2dfa903f8904e1805cab6c1900a62R392)? Where it is now, or as a method on `Path`, or somewhere else still?
 3. What amount of effort should be done to prevent the "expected integer, got integer" error message when an integer does not fit a `usize`? See also [this comment block](https://github.com/m0n0chr0m3/tremor-runtime/commit/21bcf3be14a2d4060a9c76e1f687fe757f93af41#diff-6be2dfa903f8904e1805cab6c1900a62R481). In the code of this commit, only `is_i64` is checked. On, e.g., 32-bit systems, it might make sense to also check whether `val.is_u64()`, and checking whether `val.is_u128() || val.is_i128()` would in theory be even better, but the current implementations for 128-bit values seems to just sign-extend from 64-bit values anyway.
 4. Can you double-check that the error message for array index out of bounds is correct now (i.e., that there used to be an off-by-one error)?

If that's all figured out, and you're all happy with the state of this PR, I'll squash/rebase and remove this PR's "draft" status.

Thanks all.

r?


m0n0chr0m3:
Okay, question 5: what did I break?

m0n0chr0m3:
Question 5 solved: [this fix](https://github.com/wayfair-tremor/tremor-runtime/pull/164/commits/1ce2d96945f0b5608064555b54fa1d12bc12a85c#diff-c4178d4ef3caecfc6871174e162c63c5R478) was not yet applied to the expected output of the integration tests.

It looks like [this fix](https://github.com/wayfair-tremor/tremor-runtime/pull/164/commits/1ce2d96945f0b5608064555b54fa1d12bc12a85c#diff-c4178d4ef3caecfc6871174e162c63c5R473) does not impact any test, but I think that's fine: it only affects the `description` of an `Error`, which [is deprecated](https://doc.rust-lang.org/std/error/trait.Error.html#method.description) anyway.

tremor-script/src/errors.rs
```diff
@@ -36,7 +36,7 @@ use serde_json;
 pub use simd_json::ValueType;
 use simd_json::{prelude::*, BorrowedValue as Value};
 use std::num;
-use std::ops::{Range as IRange, RangeInclusive};
+use std::ops::{Range as RangeExclusive, RangeInclusive};
```
Licenser, line 39: Much better naming 👍

tremor-script/src/errors.rs
```diff
@@ -811,17 +830,77 @@ pub(crate) fn error_array_out_of_bound<'script, T, O: BaseExpr, I: BaseExpr>(
...
     path: &ast::Path<'script>,
-    r: IRange<usize>,
+    r: RangeExclusive<usize>,
+    len: usize,
     meta: &NodeMetas,
 ) -> Result<T> {
     let expr: Range = outer.extent(meta);
+    // FIXME: Why match on `path` when all arms do the same?!
+    // -> ideally: put the `path` into the `ErrorKind::ArrayOutOfRange`, handle in display
+    //        but: not trivial: `Path` is parametric in non-'static lifetime 'script
```
Licenser, line 840: Perhaps introducing a PathType enum would collapse this and allow generalizing it it some places?

```rust
enum PathType {
Local, 
Const,
Meta,
Event,
State,
}
impl Path {
  fn path_type(&self) -> PathType {...}
}
```

tremor-script/src/interpreter.rs
```diff
@@ -388,8 +388,9 @@ where
     'script: 'event,
     'event: 'run,
 {
-    let mut subrange: Option<(usize, usize)> = None;
-    let mut current: &Value = match path {
+    // Fetch the base of the path
+    // TODO: Extract this into a method on `Path`?
```
Licenser, line 392: 👍 seems sensible

tremor-script/src/interpreter.rs
```diff
@@ -434,48 +439,100 @@ where
...
+            } => {
+                if let Some(a) = current.as_array() {
+                    let range_to_consider = subrange.unwrap_or_else(|| a.as_slice());
 
-                    if let Some(c) = a.get(idx) {
-                        current = c;
-                        subrange = None;
-                        continue;
-                    } else {
+                    // Helper closure for evaluating an expression to an `usize` index
```
Licenser, line 475: We have a `eval_to_string` function in https://github.com/wayfair-tremor/tremor-runtime/blob/master/tremor-script/src/interpreter/imut_expr.rs#L54 would it make sense to pull this out into `eval_to_usize`?

That might reduce the code in this function a bit which was always too large I think.

Licenser:
> Do the new error messages capture what you were going for in  m0n0chr0m3@1ce2d96#commitcomment-38283863?

They are only thought when reading them, perhaps we should spell out that the range is max exclusive. I'm thinking of a support person who might use ranges in tremor script for the first time and got a pager with

```
Array index out of bounds, got index 2 but expected an index in the range 0:2
```

That could be confusing, perhaps spelling it out in a NOTE or just the error itself makes sense?

i.e. (in the error)
```
Array index out of bounds, got index 2 but expected an index in the range 0:2 (max exclusive)
```

in a note
```
Array index out of bounds, got index 2 but expected an index in the range 0:2
NOTE: Ranges are max-exclusive so the range 0:7 covers the indexes 0 to 6
```

> What is the ideal place for the code extracting the base value to resolve from? Where it is now, or as a method on Path, or somewhere else still?

Having it in Path isn't a bad idea, there might be a chance to reduce the copy&pasta between resolve and assign https://github.com/wayfair-tremor/tremor-runtime/blob/3e54fd4458768411a2b1c444e5a7b887128e0b44/tremor-script/src/interpreter/expr.rs#L313 worth considering while in this realm.

> What amount of effort should be done to prevent the "expected integer, got integer" error message when an integer does not fit a usize? See also this comment block. In the code of this commit, only is_i64 is checked. On, e.g., 32-bit systems, it might make sense to also check whether val.is_u64(), and checking whether val.is_u128() || val.is_i128() would in theory be even better, but the current implementations for 128-bit values seems to just sign-extend from 64-bit values anyway.

`is_*128` depends on the feature sets selected for `simd-json`. We're currently not using the 128-bit extension since it does degrade parsing performance and we simply haven't had the need :D. Checking for `is_*128` still isn't a bad idea as it will seamlessly work with swapping features. But taking your comment I'm think it'd make sense to extend the ValueTrait that defines the whole `is_*` codes by `is_integer`, `is_float` and `is_number` (it's a simple extension I'll do that after this comment).

Generally, I am not sure how likely it is to see tremor running on a 32-bit system, the likeliest would be an ARM system and that's still going to be a while until the NEON intrinsics make it to stable rust, and even then it'd only apply to armv7, not aarch64 and we'd need to invest some serious effort to make sure it works well. And then still it'd only be hit if someone tries to access an array with more then 4294967295 elements - taking an optimistic estimate that the array is only 64 bit per element (the Value appear will add more!) that'd be 32G of memory which wouldn't work on a 32bit system anyway.

> Can you double-check that the error message for array index out of bounds is correct now (i.e., that there used to be an off-by-one error)?

They seem to be right!

Licenser:
as of `v0.1.8` value-trait supports `is_number`, `is_integer` and `is_float` that might simplify the checks a bit

tremor-script/src/interpreter.rs
m0n0chr0m3: That's a good idea.

I think it might help to do a small refactor to group the context of the interpreter into a struct, since there's a lot of code duplication due to dragging the `opts`, `env`, `event`, `state`, `meta`, and `local` in and out of each method there. Tiny refactor. Really, no biggie refactor. Oh well, https://github.com/m0n0chr0m3/tremor-runtime/commit/fc921c5fa151d4e2cc6756583f5a66c586907234, "306 additions and 451 deletions", reducing code duplication achieved.

Licenser: hah :D 
* for some definitions of tiny
looks good general I left one comment on it where I would be a bit concerned if we allocate and copy the whole struct for every time we switch between  with and without result.

tremor-script/src/errors.rs
m0n0chr0m3: Extracted it into a new issue: https://github.com/wayfair-tremor/tremor-runtime/issues/168

m0n0chr0m3:
Oh, right, pushing to a branch that's part of a PR triggers GitHub Actions on the upstream repo. My apologies.

tremor-script/src/interpreter.rs
m0n0chr0m3: Resolved: `opts` are no longer in the context `struct`.

tremor-script/src/interpreter.rs
m0n0chr0m3: It won't be so straightforward to make this abstract enough that both `resolve` and `present` can use it, since in case of `present` sometimes `Cow::Borrowed(&FALSE)` has to be returned in cases where `resolve` would return an `Err(_)`. That `Cow` makes sense for neither the `Ok(_)`, nor the `Err(_)` branch.

I see one architecturally clean option for deduplicating code here and lower in the functions `resolve` and `present` (and perhaps `assign`) but that solution is way out of scope for this PR: if the interpreter's errors were restructured such that all `ErrorKind`s which have to do with the non-presence of a value are grouped into a new variant `ErrorKind::NotPresent(NotPresentError)`, then the abstracting methods could behave ± like the code in `resolve`, and `present` could simply

```rust
match res {
  Ok(res) => res,
  Err(ErrorKind::NotPresent(_np)) => return Ok(Cow::Borrowed(&FALSE)),
  Err(other) => return Err(other)
}
```

In any case: I won't handle this `TODO` in this PR.

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -60,12 +62,37 @@ where
         meta: &'run Value<'event>,
         local: &'run LocalStack<'event>,
     ) -> Result<Cow<'event, str>> {
-        match stry!(self.run(opts, env, event, state, meta, local)).borrow() {
-            Value::String(s) => Ok(s.clone()),
-            other => error_need_obj(self, self, other.value_type(), &env.meta),
+        match stry!(self.run(opts, env, event, state, meta, local)).into_owned() {
```
Licenser, line 65: I think this is worth refactoring especially to move the into_owned into the inner so we don't clone potentially large structs.

i.e. if for bad reasons the `run` evaluates to the entire event we'd clone the whole thing before we error while we don't need to.

```rust
let value = stry!(self.run(opts, env, event, state, meta, local));
if let Some(s) = value.as_str() {
  Ok(s.into_owned())
} else {
  error_need_str(self, self, value.value_type(), &env.meta)
}
```

That would also get rid of  the deconstruction and depending on the enum representation.

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -181,21 +208,13 @@ where
...
-                    self,
-                    local,
-                    &env.meta,
-                    expr.key_id,
-                    Value::String(k)
-                ));
-                stry!(set_local_shadow(self, local, &env.meta, expr.val_id, v));
+
+            'comprehension_outer: for (k, v) in target_map {
+                let k = Value::String(k.clone());
```
Licenser, line 215: using `Value::from()` is the more idiomatic approach here as it avoids the internal representation
```suggestion
                let k = Value::form(k.clone());
```

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -60,12 +62,37 @@ where
...
+        state: &'run Value<'static>,
+        meta: &'run Value<'event>,
+        local: &'run LocalStack<'event>,
+        path: &'script Path,
+        array: &[Value],
+    ) -> Result<usize>
+    where
+        Expr: BaseExpr,
+    {
+        let val = self.run(opts, env, event, state, meta, local)?;
```
Licenser, line 92: this should use  `stry!` instead of `?`

Licenser: there are a few more places about this I'll call out for simplicty.
```suggestion
        let val = stry!(self.run(opts, env, event, state, meta, local));
```

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -181,21 +208,13 @@ where
...
-                    local,
-                    &env.meta,
-                    expr.key_id,
-                    Value::String(k)
-                ));
-                stry!(set_local_shadow(self, local, &env.meta, expr.val_id, v));
+
+            'comprehension_outer: for (k, v) in target_map {
+                let k = Value::String(k.clone());
+                set_local_shadow(self, local, &env.meta, expr.key_id, k)?;
```
Licenser, line 216:
```suggestion
                stry!(set_local_shadow(self, local, &env.meta, expr.key_id, k));
```

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -181,21 +208,13 @@ where
...
+                set_local_shadow(self, local, &env.meta, expr.val_id, v.clone())?;
```
Licenser, line 217:
```suggestion
                stry!(set_local_shadow(self, local, &env.meta, expr.val_id, v.clone()));
```

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -213,21 +232,10 @@ where
...
-                    self,
-                    local,
-                    &env.meta,
-                    expr.key_id,
-                    count.into()
-                ));
-                stry!(set_local_shadow(self, local, &env.meta, expr.val_id, x));
+            'comp_array_outer: for x in target_array {
+                set_local_shadow(self, local, &env.meta, expr.key_id, count.into())?;
+                set_local_shadow(self, local, &env.meta, expr.val_id, x.clone())?;
```
Licenser, line 238:
```suggestion
                stry!(set_local_shadow(self, local, &env.meta, expr.key_id, count.into()));
                stry!(set_local_shadow(self, local, &env.meta, expr.val_id, x.clone()));
```

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -367,108 +377,95 @@ where
...
-                        let (start, end) = if let Some((start, end)) = subrange {
-                            // We check range on setting the subrange!
-                            (start, end)
+                        let array = subrange.unwrap_or_else(|| a.as_slice());
+                        let start_idx = range_start.eval_to_index(
+                            self, opts, env, event, state, meta, local, path, &array,
+                        )?;
+                        let end_idx = range_end.eval_to_index(
+                            self, opts, env, event, state, meta, local, path, &array,
+                        )?;
```
Licenser, line 429:
```suggestion
                        let start_idx = stry!(range_start.eval_to_index(
                            self, opts, env, event, state, meta, local, path, &array,
                        ));
                        let end_idx = stry!(range_end.eval_to_index(
                            self, opts, env, event, state, meta, local, path, &array,
                        ));
```

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -367,108 +377,95 @@ where
...
+                    } else {
+                        // FIXME: Indexing into something that isn't an array: should this return
+                        // false, or return an error? (Should probably have the same answer as the
+                        // same question a couple of lines higher in this function.)
                         return Ok(Cow::Borrowed(&FALSE));
+                    }
+                }
+                // Next segment is an expression: run `expr` to know which key it signifies at runtime
+                Segment::Element { expr, .. } => {
+                    let key = expr.run(opts, env, event, state, meta, local)?;
```
Licenser, line 451:
```suggestion
                    let key = stry!(expr.run(opts, env, event, state, meta, local));
```

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -367,108 +377,95 @@ where
...
+
+                    let next = match (current, key.borrow()) {
+                        // The segment resolved to an identifier, and `current` is an object: lookup
+                        (Value::Object(o), Value::String(id)) => o.get(id),
+                        // If `current` is an array, the segment has to be an index
+                        (Value::Array(a), idx) => {
+                            let array = subrange.unwrap_or_else(|| a.as_slice());
+                            let idx = value_to_index(self, segment, idx, env, path, array)?;
```
Licenser, line 459:
```suggestion
                            let idx = stry!(value_to_index(self, segment, idx, env, path, array));
```

tremor-script/src/interpreter/imut_expr.rs
m0n0chr0m3: I had presumed that type errors are rare, and never on a "hot" path. I'll keep the performance impact on the error branches in mind. Thanks.

It's a bit sad that `eval_to_string` returns `Result<Cow<'event, str>>` instead of `Result<Cow<'run, str>>`, but from its use-sites it's clear that that's probably reasonable. Still, that means that `eval_to_string` _always_ returns an owned copy of the string, instead of being able to propagate the (possibly borrowed) result of `run`.

Side-note: the code in the comment doesn't seem to work as-is: in `Ok(s.into_owned())`, `s` is of type `&str`, and no implementation of `str::into_owned() -> Cow<'event, str>` exists. I'll adapt it slightly to
```rust
        let value = stry!(self.run(opts, env, event, state, meta, local));
        if let Some(s) = value.as_str() {
            Ok(Cow::from(s.to_owned())) // <-- changed
        } else {
            error_need_str(self, self, value.value_type(), &env.meta)
        }
```

Licenser: Ja was written from memory :) tweaks are likely needed.

m0n0chr0m3:
Squash/rebase-time?

Licenser:
Go fir it :)

PR #180, opened 2020-04-16
Print http errors, fixes #165

### LINKED ISSUE #165: logging error response codes in replies from REST offramp
Right now Tremor is not logging any error message in case there is a bad response from the offramp REST endpoints. This makes little difficult for troubleshooting any request failure.

Log the HTTP request & response in case the REST endpoints return a bad response code. It is desirable if the REST offramp can be configured to accept  a list of HTTP codes as good response. By default it may consider any code 400+ as bad response.

It is highly desirable to have  these unsuccessful messages routed to a different output . It will help in future investigation and reprocessing of the data.

**Notes**

Licenser:


Licenser:
@dependabot merge

PR #186, opened 2020-04-20
Adds unit tests for the types in `tremor-script::ctx`

### LINKED ISSUE #183: Extra empty path segments in `EventOriginUri`s
**Problem**
The `path` method on `EventOriginUri`s behaves different from what I would have expected, and yields different results from what the `uri` crate's `path_segments` yields: invoking `path()` on paths of the form `"scheme://host.name/"` yields a vector of _two_ empty strings, where I'd expect only one.

**Steps**
Consider the behavior at https://github.com/m0n0chr0m3/tremor-runtime/commit/3798d4383a7f8488bbf16042e927ce74ae703dc9#diff-bdd33b2308718bd57f53842f6df9fce6R166. The assertions as written there succeed in the current implementation of `tremor-script`. I'd expect the version in the comment below it. Similarly for the rest of that test case.

For reference, the behavior of the `url` crate's `path_segments` method on those inputs is as follows:
```
"protocol://the.host.name": None
"protocol://the.host.name:8080": None
"protocol://the.host.name/": Some([""])
"protocol://the.host.name/some/path/segments": Some(["some", "path", "segments"])
"protocol://the.host.name/some/path/segments/": Some(["some", "path", "segments", ""])
"protocol://host.names.are.%F0%9F%94%A5": None
```
(as demonstrated [on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b47018761b61d7ecfba898edb07b7e70))

As far as I see, none of the unit tests currently check any of the [origin namespace](https://docs.tremor.rs/tremor-script/functions/origin/) Tremor functions. The Tremor test harness for unit tests runs all scripts/pipelines/etc. in a context where the origin is not set, so all functions would return `null` anyway. I hence cannot gauge the impact of 'fixing' the `path` method.

**Possible Solution(s)**
1. If the behavior is correct as-is, close this bug report.
2. If the behavior is wrong, and the path segments for`"scheme://host/"` should be `vec![""]`, and the path segments for `"scheme://host"` should be `vec![""]` too, tell me and I'll open a PR.
3. If the behavior is wrong, and the path segments for`"scheme://host/"` should be `vec![""]`, and the path segments for `"scheme://host"` should be `vec![]`, tell me and I'll open a PR.
4. If the behavior is wrong, and the path segments for`"scheme://host/"` should be `Some(vec![""])`, and the path segments for `"scheme://host"` should be `None`, tell me and I'll open a PR.
5. If the behavior is wrong, but none of the above, tell me what's expected and I'll look into it. ;)

**Notes**

Output of `tremor-server --version`:
```
tremor version: 0.7.3
rd_kafka version: 0x000000ff, 1.3.0
tremor-runtime 0.7.3
```

m0n0chr0m3:
Adds unit tests for the types in `tremor-script::ctx`.

Other minor refactorings or fixes are:
- Changed what constitutes a path segment as discussed with the core devs in the comments on issue #183.
- Removed the `PartialOrd` bound on `EventOriginUri` and `EventContext`, since it seems not to be in use.
- Changed the return type of `EventOriginUri::port` from `&Option<u16>` to `Option<u16>`, which is more idiomatic for such small `Copy` types.
- Changed the return type of `EventOriginUri::path` from `&Vec<String>` to `&[String]`, which is more idiomatic since a `&Vec<_>` is more constraining to the API maintainer but offers no benefit to the API user.
- Removed the possibility for `EventOriginUri::display` to fail to write the scheme and/or host, yet to try to write the port anyway.
- Made the URL-parse-error use the inner error's `Display` instead of its `Debug`, i.e., it now shows `"Url Parse Error: invalid port number"` instead of `"Url Parse Error: InvalidPort"`, and `"Url Parse Error: invalid domain character"` instead of `"Url Parse Error: InvalidDomainCharacter"`, and so on (though only those two cases are tested).

Let me know if any of those changes are unwanted, or if other changes are needed.

I'm not sure what the preferred granularity of `#[test] fn` per test case is. I grouped them into what I consider logical clusters in a way that is consistent with some of the other test cases in the code base, but feedback is welcome.

Fixes #183. r?


Licenser:
I think that needs a `cargo update` (my fault, sorry)

Licenser:
seems like it wasn't just me, oh dependencies :sob: I'll make aPR to master

Licenser:
Oh my that was painful - the errors should be fixed once https://github.com/wayfair-tremor/tremor-runtime/pull/187 is merged

Licenser:
got just a clippy issue to fix :)

m0n0chr0m3:
> Add more commits by pushing to the tests-for-script-ctx branch on m0n0chr0m3/tremor-runtime.

...it says, but that's a lie, since I _did_ push to [that branch](https://github.com/m0n0chr0m3/tremor-runtime/commits/tests-for-script-ctx) with [the clippy fix](https://github.com/m0n0chr0m3/tremor-runtime/commit/02086cee6796aa14352126063fa097573b4640a6#diff-bdd33b2308718bd57f53842f6df9fce6R50), but GitHub thinks I pushed that 3 days ago so it's not updating here. Uh-oh, I broke GitHub.

Licenser:
I give up on github :joy: merged!

PR #217, opened 2020-05-04
Add exit offramp for integration testing

darach:
Adds an exit offramp that responds to events of the form `{"exit": <u32> }` and terminates
the runtime with the status code provided. The offramp is intended for use with for integrating testing purposes but its use is unconstrained.


src/offramp/exit.rs
```diff
@@ -0,0 +1,68 @@
...
+        }))
+    }
+}
+
+impl Offramp for Exit {
+    fn on_event(&mut self, _codec: &Box<dyn Codec>, _input: String, event: Event) -> Result<()> {
+        for (value, _meta) in event.value_meta_iter() {
+            if let Some(Value::Static(StaticNode::I64(status))) = value.get("exit") {
+                #[allow(clippy::cast_possible_truncation)]
+                std::process::exit(*status as i32);
```
Licenser, line 47: This will satisfy the pedantic mode :D
```suggestion
                // ALLOW: this is the supposed to exit
                std::process::exit(*status as i32);
```

PR #219, opened 2020-05-05
Add url en/de-code functions to std lib

darach:
Adds url encode and decode functions to handle percent encoded
strings.

Example:

```
use std::url;

url::decode(url::encode("foo bar baz"))
```

### DIFF
--- a/tremor-script/src/std_lib/url.rs
+++ b/tremor-script/src/std_lib/url.rs
@@ -20,8 +20,7 @@ use simd_json::BorrowedValue;
 pub fn load(registry: &mut Registry) {
     registry
         .insert(tremor_fn! (url::decode(ctx, s: String) {
-            let ss = s.to_string();
-            let ds = percent_decode_str(&ss).decode_utf8();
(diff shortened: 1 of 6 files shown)


tremor-script/src/std_lib/url.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use crate::registry::Registry;
+use crate::tremor_fn;
+use percent_encoding::{percent_decode_str, utf8_percent_encode, NON_ALPHANUMERIC};
+use simd_json::BorrowedValue;
+
+pub fn load(registry: &mut Registry) {
+    registry
+        .insert(tremor_fn! (url::decode(ctx, s: String) {
+            let ss = s.to_string();
```
anupdhml, line 23: ```percent_decode_str(&s).decode_utf8();``` should work here? `ss` assignment can be removed then.

darach: Well spotted. Not needed in this case

PR #230, opened 2020-05-08
Make members of tremor script doc structs public

anupdhml:
This is for use with tremor-language-server v0.8: https://github.com/wayfair-tremor/tremor-language-server/pull/33

Also improve display of function signature doc when info on return type is absent.


anupdhml:
Ignoring the integration test failures (for `query-exit`) since that is being tracked as part of: https://github.com/wayfair-tremor/tremor-runtime/pull/226

Issue #235, opened 2020-05-11
Tremor "statsd" codec validation throws error for right format payload


anuprout:
**Problem**

The statsd codec validation is not able to parse payload with valid statsd format. The rate field should be optional while it expects that. Even though by passing the rate filed, it errors out .Below are the sample payloads and corresponding errors Tremor throws.

sam:7|c
ERROR tremor_runtime::onramp::prelude - [Codec] Invalid statsd metric

sam:7|c|@0.1
ERROR tremor_runtime::onramp::prelude - [Codec] invalid float literal

**Steps**

1. starts a Tremor service UDP onramp with statsd codec
2. Send the above UDP payload to the Tremor service

**Possible Solution(s)**

**Notes**

Output of `rustup --version`:
Output of `rustup show`:
Output of `tremor-server --version`: 0.7.3

anupdhml:
Closing this out, since this turned out to be a case of payloads ending in a new line, which can be handled with the lines preprocessor: https://github.com/wayfair-tremor/tremor-runtime/pull/244#issuecomment-640784326

PR #244, opened 2020-05-14
fix for issue #235

anuprout:
This is the fix for https://github.com/wayfair-tremor/tremor-runtime/issues/235.


src/codec/statsd.rs
```diff
@@ -163,14 +163,14 @@ fn decode<'input>(data: &'input [u8], _ingest_ns: u64) -> Result<Value<'input>>
...
-                let v: f64 = s.parse()?;
+                let v: f64 = s.trim().parse()?;
                 m.insert("sample_rate".into(), Value::from(v));
             } else {
                 return Err(ErrorKind::InvalidStatsD.into());
             }
         }
         None => (),
-        _ => return Err(ErrorKind::InvalidStatsD.into()),
+        _ => (),//return Err(ErrorKind::InvalidStatsD.into()),
```
Licenser, line 173: It could turn out problematic to silently drop invalid responses, that might make it hard to find mistakes, is there a  reason that's not visible from the snippet that needs to be commented out?

anuprout: Yeah. It is basically failing at new line character. So changed the code to ignore the new line char.

Licenser:
👍 there are no changes left, should we close this in favour of #281?

anuprout:
closing this PR since this can be achieved by lines preprocessor.

PR #274, opened 2020-06-03
Support musl builds along with packaging

anupdhml:
Package formats supported currently are archive (tar.gz) and deb.

Some documentation at: https://github.com/wayfair-tremor/tremor-runtime/tree/musl_plus_packaging/packaging


Licenser:
Currenty the build fails for musl with

```
error[E0308]: mismatched types
   --> /cargo/registry/src/github.com-1ecc6299db9ec823/simd-json-0.3.9/src/lib.rs:216:86
    |
216 | fn please_compile_with_a_simd_compatible_cpu_setting_read_the_simdjonsrs_readme() -> ! {}
    |    ----------------------------------------------------------------------------      ^ expected `!`, found `()`
    |    |
    |    implicitly returns `()` as its body has no tail or `return` expression
    |
    = note:   expected type `!`
            found unit type `()`
```

Which is a bit surprising since based on the output the flags are set correctly.

packaging/cross_build.sh
```diff
@@ -0,0 +1,99 @@
...
+###############################################################################
+
+# install cross if not already there (helps us build easily across various targets)
+# see https://github.com/rust-embedded/cross
+#
+# currently need to install it from a personal fork for builds to work against custom targets
+# (eg: x86_64-alpine-linux-musl which we use for generating working musl binaries right now)
+if ! command -v cross > /dev/null 2>&1; then
+  echo "Installing cross..."
+  cargo install --git https://github.com/anupdhml/cross.git --branch custom_target_fixes
```
Licenser, line 51: Can we upstream that?

packaging/cross_build.sh
```diff
@@ -0,0 +1,99 @@
...
+  CUSTOM_RUSTFLAGS+=("-C" "target-cpu=native")
+fi
+
+if [[ "$TARGET" == *"alpine-linux-musl"* ]]; then
+  # force static binaries for alpine-linux-musl targets (since we are choosing this
+  # target specifically to produce working static musl binaries). Static building
+  # is the default rustc behavior for musl targets, but alpine disables it by
+  # default (via patches to rust).
+  echo "Ensuring static builds for alpine-linux-musl targets..."
+  CUSTOM_RUSTFLAGS+=("-C" "target-feature=+crt-static")
```
Licenser, line 81: This overwrites the target-features set before
```suggestion
  CUSTOM_RUSTFLAGS+=("-C" "target-feature=+crt-static,+avx,+avx2,+sse4.2")
```

anupdhml: good catch!

packaging/cross_build.sh
anupdhml: will do. I wanted to try out some other things on that branch before filing a PR -- will test them out soon and submit it upstream.

packaging/cross_build.sh
Licenser: 👍 I like the storing it as list approach a lot better :D great ide!

packaging/cross_build.sh
anupdhml: got some time today to finally make that upstream PR!

https://github.com/rust-embedded/cross/pull/431

packaging/builder-images/Dockerfile.x86_64-unknown-linux-musl
```diff
@@ -0,0 +1,29 @@
...
+    DL_CMD="curl -C - -L -o" \
+    OUTPUT=/usr/local/ \
+    TARGET=x86_64-linux-musl \
+  && rm -rf $temp_dir
+
+ENV CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER=x86_64-linux-musl-gcc \
+    CC_x86_64_unknown_linux_musl=x86_64-linux-musl-gcc \
+    CXX_x86_64_unknown_linux_musl=x86_64-linux-musl-g++
+
+RUN rustup target add x86_64-unknown-linux-musl
```
anupdhml, line 26: filed an issue on cross upstream to see if we can bump up the gcc version on the default cross images:
https://github.com/rust-embedded/cross/issues/432

If we can do that, we can derive this image from ` rustembedded/cross:x86_64-unknown-linux-musl` and not have to install musl ourselves.

rust-toolchain
```diff
@@ -0,0 +1 @@
+1.43.1
```
Licenser, line 1: Got a new rust version :D
```suggestion
1.44.0
```

PR #281, opened 2020-06-06
Line preprocessor fix to handle datagrams without buffering

anuprout:
This branch has changes to process datagrams in lines preprocessor without buffering them.
It also has few changes for better logging of the statsd codec.


Licenser:
This seems to mix with changes in #244 can we split the changes apart in two separate PR's since they tackle different issues?

anupdhml:
Closed out in favor of https://github.com/wayfair-tremor/tremor-runtime/pull/285.

PR #328, opened 2020-06-30
Add alex CI action

Licenser:
Groundwork on setting up alex for CI (and locally). Currently there is too much missing in Alex / the CI to make this work open issues are:

* [ ] support for ignore and globs (so we can include .rs files) https://github.com/get-alex/alex/issues/287
* [ ] support for `.alexignore` in CI https://github.com/brown-ccv/alex-recommends/issues/3
* [ ] support for additional filetype in CI https://github.com/brown-ccv/alex-recommends/issues/2


Licenser:
The alex folks suggested that using alex for rust code isn't a good idea.

Licenser:
I will close this since the Alex folks said not to use alex with rust :(

PR #356, opened 2020-07-15
updated the tremor script integer parse function documentation

rajasimha1991:
The integer parse function has a typo. Updating the return type from `float` to `integer`.


anupdhml:
Thanks for the fix, Raj!

Licenser:
Thank you!

PR #357, opened 2020-07-15
Implement initial offramp for newrelic

jaysonsantos:
This has gzip, insert and license key and, multiple region support
Sample offramp config:
```yaml
offramp:
  - id: newrelic
    type: newrelic
    config:
      license_key: deadbeef
      compress_logs: true
      region: europe
```


jaysonsantos:
For what it is worth, this sends the data to newrelic successfuly (newrelic returns 202 and gives a requestId).
I will try and figure out with their team why I cannot see the logs on their interface.

Licenser:
Thanks for the great contribution! :+1:

PR #375, opened 2020-08-04
#225 (patch in place issue)

### LINKED ISSUE #225: Patch-in-place optimization misbehaves when patch clauses use patched object
**Context**
Tremor Script features an optimization for `patch` operations by which expressions like
```tremor
let obj = patch obj of ... end;
```
can omit some copying by mutating the patched object in-place.

**Problem**
While semantically the assignment back to the original object only happens at the end of the `patch`, the `patch_in_place` optimization materializes these changes immediately. This distinction can be observed when the object being patched is used in the patch clauses.

Consider the Tremor program
```
let obj = {};
let obj = patch obj of
   insert "a" => obj,
   insert "b" => obj,
   insert "c" => obj
end;

obj
```
It should be equivalent to
```tremor
let obj = {};
let temp = patch obj of
   insert "a" => obj,
   insert "b" => obj,
   insert "c" => obj
end;

let obj = temp;

obj
```
However, the latter correctly returns `{"a": {}, "b": {}, "c": {}}`, while the former erroneously reveals the interim steps, returning `{"a": {}, "b": {"a": {}}, "c": {"a": {}, "b": {"a": {}}}}`.

**Steps**
Run the two Tremor scripts described above, and compare their outputs.

**Possible Solution(s)**
The `patch_in_place` optimization is only executed when [two conditions are met](https://github.com/wayfair-tremor/tremor-runtime/blob/8b7ee71a749957bb9240e352399b72d8a3bc4064/tremor-script/src/interpreter/expr.rs#L151-L156):
 1. the patch is on the RHS of an assignment
 2. the path of the assigned value and the path of the patched expression are identical

A third condition could be added:
 3. the patched value is not used anywhere inside the patch clauses

This would be a conservative over-approximation, but might allow you to keep the performance boost in the majority of actual uses.

Licenser:
Fixes #225


darach (approved):
LGTM

tremor-script/src/interpreter.rs
```diff
@@ -674,7 +674,55 @@ where
     Ok(())
 }
 
+/// enum mirroring `PatchOperation` carrying evaluated elements
+/// as we need to evaluate expressions in patch operations
+/// in one go, before we do the actual in-place manipulations
+/// in order to not expose temporary states of the patched object to intermittent operations
+/// example:
```
m0n0chr0m3, line 681: This example isn't very clear. It does not even appear to be valid Tremor Script syntax according to https://docs.tremor.rs/tremor-script/#patch.

tremor-script/src/interpreter.rs
```diff
@@ -693,99 +741,138 @@ where
     'event: 'run,
 {
     let patch_expr = expr;
+    let mut evaluated: Vec<PreEvaluatedPatchOperation> = Vec::with_capacity(expr.operations.len());
```
m0n0chr0m3, line 744: Is there a specific reason you chose to go for
```rust
let mut coll : Vec<_> = Vec::with_capacity(a.len());
for op in &a {
   coll.push( /* transform op */ );
}
```
instead of
```rust
let coll : Vec<_> = a.iter()
   .map(|op| /* transform op */)
   .collect();
```
?

tremor-script/src/interpreter.rs
```diff
@@ -693,99 +741,138 @@ where
...
+    // first pass over the operations, evaluating them
+    // and (IMPORTANT) get it into an owned, possibly cloned value, so we reference
+    // the target value in the state before any patch operation has been executed.
     for op in &expr.operations {
-        // NOTE: This if is inside the for loop to prevent obj to be updated
-        // between iterations and possibly lead to dangling pointers
+        evaluated.push(match op {
```
m0n0chr0m3, line 749: Perhaps you can extract the conversion from `PatchOperation` into `PreEvaluatedPatchOperation` as a method on either type. That might also fix of the clippy warning suppressed on line 725. The main blocker would be the need to pass in `opts`, `env`, `event`, `state`, `meta`, and `local` to another function. On the other hand, the Tremor code base doesn't shy away from using explicit `inline` annotations, so it might clean up the code base without runtime overhead.

tremor-script/src/interpreter.rs
```diff
@@ -693,99 +741,138 @@ where
...
+            },
+            PatchOperation::TupleMerge { expr } => PreEvaluatedPatchOperation::TupleMerge {
+                merge_value: stry!(expr.run(opts, env, event, state, meta, local)).into_owned(),
+            },
+        })
+    }
+
+    // second pass over pre-evaluated operations
+    // executing them against the actual target value
+    for const_op in evaluated.drain(..) {
```
m0n0chr0m3, line 788: Why `drain(..)` instead of moving the `evaluated` into the `for` loop (implicitly turning it `into_iter`)?

mfelsche: good catch, the vec is going to be dropped after this function anyways :)

tremor-script/src/interpreter.rs
mfelsche: good catch, i was just quickly sketching that down.

tremor-script/src/interpreter.rs
mfelsche: the reason is the error handling we would need to put into map. So it is a convenience thing, otherwise we would have to map every element to a result and error on the first one we find. With a for loop we can error on first occurrence.

tremor-script/src/interpreter.rs
mfelsche: ✅

m0n0chr0m3 (approved):
LGTM

Licenser (commented):
:+1:

darach (approved):
LGTM

PR #409, opened 2020-08-28
Fix server routes due to new path variable syntax in tide.

mfelsche:
I tried to use the api to get an onramp i deployed to the repository and it responded with 404 although a `GET /onramp` returned the exact artefact id being used. It turned out the artefact id was always empty, because the tide path variable syntax changed: https://docs.rs/tide/0.13.0/tide/struct.Server.html#method.at

This should fix it. I was not able to test this with coyote yet. I still need to look into it. Currently it gives like 21 errors or so.

I hope this is fine as is, any hints/tips are very welcome!


Licenser (approved):
This looks great, thank you!

darach:
⭐ 👍 LGTM

Issue #426, opened 2020-09-14
Unified wishlist for 0.9.x


darach:
Wish list of unified client work to be done:
* [x] generating cli doc (9.1?)
* [x]  improve tremor test integration ... so assert pass/fail status is obvious in ci ( no colors means not obvious without text prompt/hint )
* [x]  Patch CI to adopt new binary for api curl based tests
* [x]  Patch CI to adopt new binary for api tremor cli based tests
* [x]  Extend CI to support tremor stdlib unit style tests
* [x]  Extend CI to support tremor cli self hosted tests
* [ ]  Figure out code coverage against unified client

Wish list for cb/wal/ack/gd:

* [ ] A Qos operator for acks, and gated acknowledgement continuations in tremor-script+query
* [ ] An operator to expose circuit-breaker events to tremor-script+query
* [ ] An operator to expose WAL events to tremor-script+query
* [x] Document usage of in-memory WAL verses persistent
* [ ] Consider whether qos ack/cb/wal operators are distinct, a single operator or some mix thereof ...
* [ ] Example of ack/cb/wal qos in concert with state mechanism
* [ ] Raw vs best effort ( in mem + persistent ) vs lossless/guaranteed ( recoverable ) stream detection & reporting in runtime
* [ ] Look into different WAL flavours (ringbuffer etc)
* [ ] Look into more efficient WAL encoding

mfelsche:
Closing this, as all open issues are tracked as new issues now.

PR #430, opened 2020-09-15
Mob branch for v0.9 release changes

### LINKED ISSUE #225: Patch-in-place optimization misbehaves when patch clauses use patched object

darach:
Log of changes to prepare for release:

Must do's:
* [x] Update versions of all crate'd sub projects in cargo
* [ ] publish tremor-script 0.9
* [x] merge linked transports
* [ ] publish new language server and plugin
* [x] fix #225 via #375 
* [x] linked transport docs
* [x] unified cli docs
* [x] Document usage of in-memory WAL verses persistent
* [x] consider WAL ingest ns change (should we update it?)
* [x] Ensure errors are propagated with LT instead of the connection left hanging open (see #489 for 0.9.1)
* [x] Add codecs and preprocessors for WS linked offramp
* [x] update WS linked offramp to use reply channel code ( we can do that in 0.9.1 see #490)

Linked transport todo's:
* [ ] integration tests
* [ ] benchmarks
* [x] move current examples to workshops in docs repo (rest api, rest proxy app, rest proxy passthrough, ws echo server, ws proxy app, ws proxy passthrough)
* [x] add workshop/examples for load balancing (and failover of) rest offramps, http to websocket bridging
* [x] address the outstanding review comments from https://github.com/tremor-rs/tremor-runtime/pull/431

Wish list of unified client work to be done (copied from #426):
* [x] generating cli doc (9.1?)
* [x] improve tremor test integration
* [x] Port over api curl based tests
* [x]  Patch CI to adopt new binary for api curl based tests
* [x] Port over api based tremor cli tests
* [x]  Patch CI to adopt new binary for api tremor cli based tests
* [x]  Extend CI to support tremor stdlib unit style tests (add tests to stdlib modules)
* [x]  Extend CI to support tremor cli self hosted tests
* [ ]  Figure out code coverage against unified client
* [ ] Location reporting of script runtime errors from `run` (an example in comparison to old tremor-script binary is at https://github.com/tremor-rs/tremor-runtime/issues/445)
* [ ]  Flag to include emit port name in the script `run` output (parity with old tremor-script binary) 

Wish list for cb/wal/ack/gd:

* [ ] A Qos operator for acks, and gated acknowledgement continuations in tremor-script+query
* [ ] An operator to expose circuit-breaker events to tremor-script+query
* [ ] An operator to expose WAL events to tremor-script+query
* [ ] Consider whether qos ack/cb/wal operators are distinct, a single operator or some mix thereof ...
* [ ] Example of ack/cb/wal qos in concert with state mechanism
* [ ] Raw vs best effort ( in mem + persistent ) vs lossless/guaranteed ( recoverable ) stream detection & reporting in runtime
* [ ] Look into different WAL flavours (ringbuffer etc)
* [ ] Look into more efficient WAL encoding

Wish list for linked transport:
* [ ] introduce peer artefact/implement source and sink trait to make a linked ramp
* [ ] iron out circuit-breaker/guaranteed-delivery overlap
* [ ] cycle detection for links in binding [ DE/HG - this should wait until we kill the yaml with fire ]

Wish list for packaging:

* [ ] Fix static linking/segfault for musl binaries (reproduce via `make package-x86_64-alpine-linux-musl`). Disable musl packaging if we push this for later.
* [ ] Package validation tests as part of CI release process (installation plus running binary)
* [ ] Enable deb packaging (resolve [openssl linking](https://github.com/tremor-rs/tremor-runtime/blob/main/Makefile#L22-L28)) for it


Licenser (approved):
LGTM

darach (commented):
LGTM. 👍

anupdhml:
LGTM

mfelsche:
Fucking finally! 🎉 🥃

PR #435, opened 2020-09-16
Allow escaping of curly braces in string literals.

### LINKED ISSUE #406: escaping { inside of strings
split out from #402 & with content copied from @anupdhml post:

Yes, let's improve the error message there -- ideally it can mention both that `{` is unterminated and that it is used for interpolation.

---

Right now, escaping `{` is not supported and this issue was for mainly fixing that.

>The left curly brace marks the start of the string interpolation and there does not seem to be a way to escape it in string literals when we don't want to interpolate -- we should allow for that.

I forgot to include the error for that earlier, with `\{`:

```
Error: 
    1 | let test = "\{
      |            ^^ An unexpected escape code '{' was found
```

Should be a small fix in the lexer: https://github.com/wayfair-tremor/tremor-runtime/blob/v0.8.0/tremor-script/src/lexer.rs#L1533

mfelsche:
Fixes #406

### DIFF
--- a/tremor-script/src/lexer.rs
+++ b/tremor-script/src/lexer.rs
@@ -1532,6 +1532,8 @@ impl<'input> Lexer<'input> {
             Some((_, '\'')) => Ok(Some('\'')),
             Some((_, '"')) => Ok(Some('"')),
             Some((_, '\\')) => Ok(Some('\\')),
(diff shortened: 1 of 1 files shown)


Licenser (approved):
:+1: LGTM nicely done :)

PR #442, opened 2020-09-21
Fix WAL issue, not advancing events.

mfelsche:
still needs some testing, but the workshop example worked fine with this fix.


darach:
LGTM once 💅 clippy checks are fixed

Licenser (approved):
Looks good 👍

PR #515, opened 2020-10-20
Fix array pattern matching reporting false positives

### LINKED ISSUE #512: Fix element matching for tremor-script array patterns
In tremor-script, using array patterns to match for elements within an array produces false positives.

Eg: the following snippet should match the default case but we get `"contains zero's"`.

```
let a = [1, 2, 3, 4, 5, 6, 7, 8, 9];
match a of
  case %[ 0 ] => "contains zero's"
  default => "does not contain zero's"
end;
```

This is a variation on the [documented example](https://docs.tremor.rs/tremor-script/#matching-array-patterns).

mfelsche:
When an array pattern (e.g. `%[ 0 ]`) did not match any element of the array, it was reported as matched.

This PR fixes this.

It also now allows using the `_` expression to match any element. Thus `%[ _ ]` will match a non-empty array.

Also some short-circuiting was added to exit faster if we detect a non-match or a match if the result is not needed.

Fixes #512


Licenser (approved):
LGTM :+1:

darach (approved):
LGTM 👍

PR #518, opened 2020-10-20
Docker improvements

### LINKED ISSUE #492: strip binary for docker builds
It would be good to strip the final binary in our docker builds so the size is a bit smaller

### LINKED ISSUE #511: update docker push CI
1) automatically upload v... tag
2) do not upload latest, do this manually as part of the release

Licenser:
* update release config to only publish tags and not the "latest". fixes #511 
* strip docker builds. fixes #492 
* use thin lto for docker (~2% perf decrease but docker builds are for compatibility not highest perf) might fix docker builds

### DIFF
--- a/.github/workflows/publish-tags.yaml
+++ b/.github/workflows/publish-tags.yaml
@@ -1,18 +1,21 @@
-name: Publish Docker - Release
+name: Publish to Registry - Release
 on:
+  release:
+    types: [published]
   push:
-    tags:
-      - v*
+    branches:
+      - master
+  schedule:
+    - cron: "0 2 * * 0" # Weekly on Sundays at 02:00
 jobs:
-  build:
+  update:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
       - name: Publish to Registry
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-tags.yaml
```diff
@@ -1,18 +1,21 @@
...
-  build:
+  update:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
       - name: Publish to Registry
         uses: elgohr/Publish-Docker-Github-Action@master
         with:
-          name: tremorproject/tremor
+          name: myDocker/repository
```
darach: This should be `tremorproject/tremor` right

mfelsche: Does this refer to some other config value, like a secret or is this a copy paste artefact?

darach (approved):
LGTM 👍

.github/workflows/publish-tags.yaml
Licenser: already changed :)

Licenser: fixed!

PR #520, opened 2020-10-20
Remove generic and associated type from SourceReply.

### LINKED ISSUE #486: Remove generic from SourceReply
SourceReply includes a generic that is only used inside ws.rs this should be contained intr that file not spread to all sources.

see `TODO 002`

mfelsche:
And internalize the stuff that was there to the ws source, the only place where it was needed.

Fixes #486


Licenser (approved):
LGTM 👍

PR #534, opened 2020-10-22
Fix dockerfile entrypoint arg quoting bug.

mfelsche:
# Pull request

The current docker image was unusable because it was quoting all args, so `exec` was treating all of them as one.

## Checklist

* [x] The code is tested


Licenser (approved):
LGTM :+1:

PR #536, opened 2020-10-22
Disable tokio dependency for rdkafka

mfelsche:
# Pull request

Disable tokio dependency for rdkafka and also remove explicit dep to `rdkafka-sys` by ugrading to a git tag: https://github.com/fede1024/rust-rdkafka/commit/2901a4b9027c3cce6fcae48c124fe5296397704a slightly beyond release 0.24.

## Description

We include tokio for rdkafka, which we dont need. We still need tokio for `tokio-postgres`, but it still makes sense.

## Checklist

* [x] The code is tested

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3803,7 +3803,6 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
- "tokio 0.2.22",
 ]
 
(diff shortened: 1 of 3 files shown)


Licenser:
Let's wait with this untill we can pull a package and not a git hash

PR #574, opened 2020-11-06
Remove quite a few array access code

Licenser:
# Pull request

Ensure we don't panic due to use of `[]` 

## Description

using `[]` to access data can lead to rust panicing, we want to forbid this.

## Checklist

* [x] The code is tested


.vscode/settings.json
```diff
@@ -1 +1,68 @@
-123,125
\ No newline at end of file
+{
```
mfelsche, line 1: is this added intentionally?

Licenser: ja

mfelsche (approved):

PR #576, opened 2020-11-10
Add docker image and action for training docker image

darach:

# Pull request

Add a docker image and action suitable for learning tremor that derives from base image

## Description

Add a docker image and action suitable for learning tremor that derives from base image

## Checklist

* [x] The code is tested


.github/workflows/publish-courseware.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Publish Tremor Courseware Docker image to Registry
+        uses: elgohr/Publish-Docker-Github-Action@master
+        with:
+          name: tremorproject/tremor
```
Licenser, line 14: can we change this to a own image name that way we can tag it and make it clear this is not a production image and shouldn't be used there?

.github/workflows/publish-courseware.yaml
```diff
@@ -0,0 +1,18 @@
+name: Publish Docker - learn
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
```
Licenser, line 9: If insteat of using its own workflow we add this step after the release we can guarnatee we get the right version

Dockerfile.learn
```diff
@@ -0,0 +1,42 @@
...
+    OPENSSL_STATIC=1
+
+# Switch back to dialog for any ad-hoc use of apt-get
+ENV DEBIAN_FRONTEND=dialog
+ENV RUSTFLAGS="-C target-feature=+avx,+avx2,+sse4.2"
+
+RUN curl https://sh.rustup.rs -sSf | \
+    sh -s -- --default-toolchain stable -y
+ENV PATH=/root/.cargo/bin:$PATH
+RUN cargo install --features=ssl websocat
```
Licenser, line 38: we should clean up /root/.cargo after this step sop we don't ship the entire intermediate artifacts and rust itself.

Licenser (changes requested):

Dockerfile.learn
darach: The binary is now moved and the `.cargo` folder removed.

anupdhml (approved):

Issue #506, opened 2020-10-19
Use common error format and function for building errors


mfelsche:
**Describe the problem you are trying to solve**

We currently create error events in onramp/offramps mainly for making linked transport usable and have more observability of internal errors (e.g. codec failures). But we do not necessarily follow a common format and do not use common codepaths to create errors.

**Describe the solution you'd like**

We should document and follow a common format and create common functions to create error events.
This will help users handle errors. Maybe a separate section about handling errors and connecting the `err` ports for getting good feedback is also helpful.

PR #581, opened 2020-11-11
Improve Websocket Sink

### LINKED ISSUE #489: handle encoder errrors in linked ws sink correctly
we right now don't return a nice error for a codec / postprocessor error instead we just log to the console. This isn't nice behaviour we should fix that.

see: `TODO 001`

### LINKED ISSUE #490: update WS linked sink to use reply channel code
The WS sink, when used linked, can introduce an up to 100ms delay to requests. This can be solved by using the same reply channel approach we used in the rest sink.

mfelsche:
# Pull request

Improve Websocket Sink

## Description

* Handle serialization of outgoing events and deserialization of incoming messages completely on the connection task that is spawned for each tcp connection (by default only 1 for the configured url) to increase parallelism between connections.
* Make use of reply_channel wherever possible to have quick acks and responses
* websocket connection lifecycle events leading to a CB close or open are handled every time an event or a signal comes in. This bumps us up to max 100ms delay in the worst case, but only for CB::close/open events, nothing else.

Fixes #489 
Fixes #490 

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/ws.rs
+++ b/src/sink/ws.rs
@@ -20,20 +20,18 @@ use async_channel::{bounded, unbounded, Receiver, Sender};
 use async_tungstenite::async_std::connect_async;
 use futures::SinkExt;
 use halfbrown::HashMap;
-use std::boxed::Box;
 use std::time::Duration;
+use std::{boxed::Box, sync::Arc};
 use tremor_pipeline::OpMeta;
+use tremor_script::LineValue;
 use tungstenite::protocol::Message;
 use url::Url;
 
-type WsAddr = Sender<(Ids, OpMeta, WsMessage)>;
-type WsReceiver = Receiver<(Ids, OpMeta, WsMessage)>;
 type WsUrl = String;
-
-enum WsMessage {
-    Binary(Vec<u8>),
-    Text(String),
-}
+type WsConnectionHandle = (
+    Option<Sender<ConnectionTaskMsg>>,
+    task::JoinHandle<Result<()>>,
+);
 
 // TODO once we get rid of link option in offramp config,
 // we can just align this with offramp Config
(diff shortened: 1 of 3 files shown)


src/sink/ws.rs
```diff
@@ -42,138 +40,273 @@ struct WsMessageMeta {
...
+                                            op_meta.clone(),
+                                        )
+                                        .await?;
+
+                                        // close connection explicitly
+                                        ws_stream.close(None).await?;
+                                        connection_lifecycle_tx
+                                            .send(WsConnectionMsg::Disconnected(url.clone()))
+                                            .await?;
+                                        break 'recv_loop; // exit recv loop in order to reconnect
```
Licenser, line 206: Wouldn't we need to fail the remaining messages at this point?

src/sink/ws.rs
```diff
@@ -195,17 +394,18 @@ impl offramp::Impl for Ws {
             let (reply_tx, _) = bounded(1);
 
             Ok(SinkManager::new_box(Self {
-                sink_url: TremorURL::from_onramp_id("ws")?, // dummy value
-                codec: Box::new(crate::codec::null::Null {}),
-                preprocessors: vec![],
-                postprocessors: vec![],
-                is_linked: false,
-                connections: HashMap::new(),
+                sink_url: TremorURL::from_onramp_id("ws")?,  // dummy value
```
Licenser, line 397: should we populate this with a bit more details?

src/sink/ws.rs
```diff
@@ -214,123 +414,48 @@ impl offramp::Impl for Ws {
...
-            );
-        }
-        Ok(events)
-    }
-
-    fn create_error_response(event_id: &Ids, e: String) -> Event {
+    fn create_error_response(event_id: &Ids, e: &str, origin_uri: &EventOriginUri) -> Event {
         let mut error_data = simd_json::value::borrowed::Object::with_capacity(2);
-        error_data.insert_nocheck("error".into(), Value::from(e.clone()));
+        error_data.insert_nocheck("error".into(), Value::from(e.to_string()));
```
Licenser, line 450: Would `message` be a better name here? `error.error` might look a bit odd

src/sink/ws.rs
mfelsche: this is overwritten with the actual instance url on `init`, so i think we are good here.

src/sink/ws.rs
mfelsche: everywhere else we use `error`. I think we have an issue for standardizing error messages. Would you be fine doing it as part of that issue? --> https://github.com/tremor-rs/tremor-runtime/issues/506

src/sink/ws.rs
mfelsche: Good catch, i am going to send only one ack if all messages coming from an event succeeded, and a fail otherwise as soon as it happens. We might have send some messages, before we failed, but at least there is only 1 ack or fail flowing backwards per event.

src/sink/ws.rs
Licenser: good call!

src/sink/ws.rs
Licenser: 👍

src/sink/ws.rs
Licenser: 👍 it's in #583 and #584

Licenser (approved):
LGTM 👍

PR #585, opened 2020-11-12
Fix slow kafka sink when queue.buffering.max.ms is set to > 0.

### LINKED ISSUE #532: Investigate kafka egress issue
Get details on the issue

mfelsche:
# Pull request

## Description

Our kafka sink was not only enqueueing every kafka record coming from an event, but also waiting for its delivery before returning. The setting `queue.buffering.max.ms` determined how the rdkafka broker thread waited upon receiving the message in its queue until it sends it off to the broker. The goal here is to possibly batch messages for more efficient transport. But we handle each sink in its own task, and this one task is blocked until message N is delivered. So there is no other message which can arrive before `queue.buffering.max.ms` expires. Ergo, we not only always waited for `queue.buffering.max.ms` (if set to > 0), we never applied any kafka internal batching and always sent out 1 message to the broker. Yes, indeed!

Now we only enqueue all messages coming from an event and then, in another spawned task, wait for the delivery to happen to send out acks/fails.

## Related

* Related Issues: fixes #532 

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/kafka.rs
+++ b/src/sink/kafka.rs
@@ -30,7 +30,7 @@ use rdkafka::{
     error::KafkaError,
     producer::{FutureProducer, FutureRecord},
 };
-use std::fmt;
+use std::{fmt, time::Duration};
 
 #[derive(Deserialize)]
 pub struct Config {
@@ -313,5 +313,17 @@ impl Sink for Kafka {
     }
     fn auto_ack(&self) -> bool {
         false
+    }
+    async fn terminate(&mut self) {
+        if self.producer.in_flight_count() > 0 {
+            // wait a second in order to flush messages.
+            let wait_secs = 1;
+            info!(
+                "[Sink::{}] Flushing messages. Waiting for {} seconds.",
+                wait_secs, &self.sink_url
+            );
+            self.producer.flush(Duration::from_secs(1));
+            info!("[Sink::{}] Terminating.", &self.sink_url);
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


Licenser (dismissed):
LGTM 👍

Licenser (approved):

src/sink/kafka.rs
```diff
@@ -60,11 +59,15 @@ pub struct Config {
...
+
+        // ENABLE LIBRDKAFKA DEBUGGING:
+        // - set librdkafka logger to debug in logger.yaml
+        // - configure: debug: "all" for this onramp
         let producer_config = producer_config
             .set("client.id", &format!("tremor-{}-{}", self.hostname, 0))
             .set("bootstrap.servers", &self.brokers.join(","))
             .set("message.timeout.ms", "5000")
-            .set("queue.buffering.max.ms", "0");
+            .set("queue.buffering.max.ms", "0"); // set to 0 for sending each message out immediately without kafka client internal batching --> low latency, busy network
```
mfelsche, line 70: QUESTION: @Licenser we wanna keep this as a default setting? Or rather use the librdkafka default (which is 5 now)?

Licenser: I was thinking to comment on this :D but I figured it's better in a separate ticket tomorrow. Basically the story is that we added those defaults at a time when we had no way to configure librdkafka. I think it would make sense to scrap all defaults we set (other than client ide and bootstrap servers) but I figure it might be worth a own ticket since it has user impact and someone might go "noo please don't!"

PR #595, opened 2020-11-20
Fix errors around string and heredoc interpolation

### LINKED ISSUE #402: Print proper error for unterminated strings inside of string interpolation
**Problem**

Right now, string literals in tremor-script don't allow `{` as a character.

This gets in the way of representing things like json string literals directly in tremor-script.

**Steps**

```
$ cat test.tremor 
let test = "{";

$ tremor-script test.tremor 
Error: 
    1 | let test = ";
      |              ^^ It looks like you forgot to terminate a string with a closing '"'
```

**Possible Solution(s)**

The left curly brace marks the start of the [string interpolation](https://docs.tremor.rs/tremor-script/#string-interpolation) and there does not seem to be a way to escape it in string literals when we don't want to interpolate -- we should allow for that.

**Notes**

For symmetry, we may also want to allow escaping for `}` (right now, `let test = "}";` is valid tremor-script), and throw an error when the right brace is used by itself in a string and unbalanced.

side note: I was not aware that the curly brace had so many names 😄  https://en.wikipedia.org/wiki/Bracket#Names_for_various_bracket_symbols

### LINKED ISSUE #594: Errors for unexpected escape codes show messed up code sample
**Problem**

The following script:

```tremor
"before \z after"
```

Shows the following error when run:

```
Error: 
    1 |  after"before \z
      | ^^^^^^^^^ An unexpected escape code 'z' was found
```

Not only is the string totally messed up with the content after the wrong escape showing up before the actual string, but also the callout points to the whole string, not the wrong escape.

**Possible Solution(s)**

The highlighter only checks by line, if it reached the point where to print the unfinished string token.
In this case it somehow prints the tokens from the line minus the actual string token, which is embedded in the Error, as the lexer couldnt finish it. Maybe it should check by `Location`, if it can now print the error token.

Or we avoid the token field in those Errors completely and just have the location-range in the Error. Then in the Highlighter we try to highlight as much as we can up until the error, and we just display the rest of the line content without highlighting, instead of manually crafting and transporting a token that is wrong in most cases.

mfelsche:
# Pull request

Errors around string and heredoc interpolation presented themselves with wrong source examples and wrong error locations.
And in some cases misleading error messages.

The root issue was that the highlighter relies on the tokens delivered from the tokenizer/preprocessor to be correct (in terms of their content and location info) in order to display the source correctly AND position the error callout correctly within that source. Whenever we have an issue with string or heredoc, e.g. an unterminated interpolation or really anything else, the string token cannot be used for displaying with the highlighter, as the token production rule just errored.
The displayed unfinished token must be chosen carefully to match with the previous tokens, the actual error location, line numbers to not scramble the displayed source context etc. escpecially when it comes to multi-line tokens like heredocs.

This PR fixes lots of discovered issues around string/heredoc interpolation and error reporting in general:

Example of fixed error message from #402:

```
Error: 
    1 | let test = "{";
      |             ^ It looks like you forgot to terminate a string interpolation with a closing '}'
      |               NOTE: Did you mean to write a literal '{' ? Escape it as '\{' or '{{'.
```

Fixed along the way:

* removed need for injecting whitespace after script content
* fix heredoc content duplications
* updated rust to 1.48.0

The code is a total desaster! Maniac tirefire! The sane way here would actually be to only rely on the tokens for highlighting, not for displaying the source context.

## Related

* Related Issues: 
  * fixes #402 
  * fixes #594 

## Checklist

* [x] The code is tested


src/codec/binflux.rs
```diff
@@ -30,9 +30,11 @@ const TYPE_FALSE: u8 = 4;
 pub struct BInflux {}
 
 impl BInflux {
+    #[allow(clippy::map_err_ignore)]
     pub fn encode(v: &simd_json::BorrowedValue) -> Result<Vec<u8>> {
         fn write_str<W: Write>(w: &mut W, s: &str) -> Result<()> {
             w.write_u16::<BigEndian>(
+                #[allow(clippy::map_err_ignore)]
```
Licenser, line 37: I don't think we need to allow this, including the error into the error will give a more descriptive error.

src/lib.rs
```diff
@@ -23,7 +23,7 @@
     clippy::unnecessary_unwrap,
     clippy::pedantic
 )]
-
+#![warn(clippy::option_if_let_else)]
```
Licenser, line 26: isn't that included in the defaults?

src/preprocessor.rs
```diff
@@ -123,6 +123,7 @@ impl SliceTrim for [u8] {
             c != b'\t' && c != b' '
         }
 
+        #[allow(clippy::option_if_let_else)]
```
Licenser, line 126: we don't want to allow clippy lints if not 100% required.

src/sink/rest.rs
```diff
@@ -65,6 +65,7 @@ impl Endpoint {
         self.as_url_with_base("http://localhost")
     }
 
+    #[allow(clippy::map_err_ignore)]
```
Licenser, line 68: see above, can we include the error instead?

src/source.rs
```diff
@@ -204,6 +204,7 @@ where
         }
     }
 
+    #[allow(clippy::option_if_let_else)]
```
Licenser, line 207: see above

src/source/kafka.rs
```diff
@@ -350,6 +350,7 @@ impl Source for Int {
             // but only commit the offsets explicitly stored via `consumer.store_offset`.
             .set("enable.auto.offset.store", "true");
 
+        #[allow(clippy::option_if_let_else)]
```
Licenser, line 353: see above

src/source/rest.rs
```diff
@@ -343,6 +343,7 @@ fn make_response(
 impl Source for Int {
     // TODO possible to do this in source trait?
     async fn pull_event(&mut self, id: u64) -> Result<SourceReply> {
+        #[allow(clippy::option_if_let_else)]
```
Licenser, line 346: see above

src/source/tcp.rs
```diff
@@ -77,6 +77,7 @@ impl Source for Int {
         &self.onramp_id
     }
 
+    #[allow(clippy::option_if_let_else)]
```
Licenser, line 80: see above

src/utils.rs
```diff
@@ -16,6 +16,7 @@
 #[must_use]
 #[cfg(not(tarpaulin_include))]
 pub fn hostname() -> String {
+    #[allow(clippy::map_err_ignore)]
```
Licenser, line 19: see above

tremor-cli/src/completions.rs
```diff
@@ -59,6 +59,7 @@ fn generate_for_shell(mut app: clap::App, shell: &str) -> Result<()> {
     }
 }
 
+#[allow(clippy::map_err_ignore)]
```
Licenser, line 62: see above

tremor-cli/src/main.rs
```diff
@@ -22,6 +22,8 @@
     clippy::unnecessary_unwrap,
     clippy::pedantic
 )]
+// in most cases, map_or_else is less readable
+#![allow(clippy::option_if_let_else)]
```
Licenser, line 26: see above

tremor-cli/src/util.rs
```diff
@@ -58,6 +58,7 @@ impl TremorApp {
             .and_then(|s| Ok(url::Url::parse(s)?))
     }
 
+    #[allow(clippy::map_err_ignore)]
```
Licenser, line 61: see above

tremor-cli/src/util.rs
```diff
@@ -67,6 +68,7 @@ impl TremorApp {
         })
     }
 
+    #[allow(clippy::map_err_ignore)]
```
Licenser, line 71: see above

tremor-cli/src/util.rs
```diff
@@ -77,6 +79,7 @@ impl TremorApp {
...
```
Licenser, line 82: see above

tremor-script/src/ast/raw.rs
```diff
@@ -1919,6 +1919,7 @@ impl<'script> Upable<'script> for LocalPathRaw<'script> {
                 vec![]
             };
             rel_path.push(id.to_string());
+            #[allow(clippy::option_if_let_else)]
```
Licenser, line 1922: see above

tremor-script/src/lib.rs
```diff
@@ -25,6 +25,8 @@
 )]
 // TODO We need this for simd-json-derive
 #![allow(clippy::missing_errors_doc)]
+// often does not work because borrow checker
+#![allow(clippy::option_if_let_else)]
```
Licenser, line 29: Can we put this around the cases where the borrow check fails not as a generic over the whole file?

Licenser (changes requested):
Let's not resolve clippy complaints by adding clippy allow

src/lib.rs
mfelsche: default is `deny`. this lets it get through, only emitting a warning.

src/sink/rest.rs
mfelsche: in this case there is no error, only a `()`. :( shitty Url lib.

rust-toolchain
```diff
@@ -1 +1 @@
-1.46.0
+1.48.0
```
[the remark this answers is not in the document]
mfelsche: good idea. manually updated for now.

Licenser (approved):
LGTM ❤️  👍

PR #597, opened 2020-11-23
Add support for requiring err before sending

### LINKED ISSUE #505: Assert that linked ramps check all ports are connected before consuming data
When starting a source we wait until `out` is connected before consuming data.

For linked ramps, we should extend this to `in`. The reason is that otherwise, we might run into a race condition where we get a first event before our `in` port is connected and lose the reply.

We also should allow the user to specify to wait for `err` this way it can be ensured that all errors are captured. This could not be mandatory since a user might be OK with dropping errors?

Licenser:

## Description

Introduces the `err_required` field to sources to allow users to require both `out` and `err` to be connected before
the source is started.

## Related

* Related Issues: closed #505 
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/55)

## Checklist

* [x] The code is tested


src/config.rs
```diff
@@ -50,6 +50,8 @@ pub struct OnRamp {
     #[serde(rename = "linked", default = "Default::default")]
     // TODO validate that this is turned on only for supported onramps (rest, ws)
     pub(crate) is_linked: bool,
+    #[serde(rename = "error_required", default = "Default::default")]
```
mfelsche, line 53: can we stick to `err` instead of `error`? It might be confusing for people which one to use.

src/onramp.rs
```diff
@@ -43,17 +43,18 @@ pub enum Msg {
 
 pub type Addr = async_channel::Sender<Msg>;
 
+pub(crate) struct OnrampConfig<'cfg> {
```
mfelsche, line 46: 🤝

src/repository/artefact.rs
```diff
@@ -448,7 +449,7 @@ impl Artefact for OnrampArtefact {
                 expect_answers -= 1;
             }
 
-            info!("Onramp {} unklinked.", id);
```
mfelsche, line 451: `unklinked` also has a ring to it, i have to admit :)

src/config.rs
Licenser: good catch!

mfelsche (approved):

PR #608, opened 2020-11-26
Update to new simd-json API for reused allocations

Licenser:
# Pull request

Update json codec to take advantage of @CJP10's improved deserialiser API.

## Description

The updated API does avoid two allocations per decoding cycle as it keeps buffers as part of the codec. This leads to a ~3% performance gain on the `realworld-througput-json` benchmark.

before: Throughput: `426.2 MB/s`
after: Throughput: `439.0 MB/s`

## Checklist

* [x] The code is tested


src/lib.rs
```diff
@@ -16,7 +16,7 @@
 
 #![forbid(warnings)]
 #![deny(missing_docs)]
-#![recursion_limit = "1024"]
+#![recursion_limit = "2048"]
```
mfelsche, line 19: lol, feeling the pain that got you here

darach (approved):
LGTM 👍

darach (approved):

src/lib.rs
Licenser: ja and good catch lets see if that can be reverted now

Licenser: yes it can! Thanks for catching this!

mfelsche (approved):

Issue #647, opened 2020-12-11
Monitoring : Pipeline metrics are not generated.


nagaprabhu:
**Problem**
Metrics for the pipeline `"measurement":"events"` are missing. (tremor version: `0.9.3`)

My Yaml file has onramp and offramp. I have a trickle file for pipeline. 
In the metrics, I am able to see `{"measurement":"ramp_events"}`  ==> This ones specified in the yaml file.
However, the metric for the pipeline is missing ==> The ones specified in a Trickle file. 

### simple.trickle
```
#!config metrics_interval_s = 5
select event from in into out;
```
### main.yaml
```
binding:
  - id: example                                     # The unique name of this binding template
    links:
      '/onramp/metronome/{instance}/out':            # Connect the inpunt to the pipeline
       - '/pipeline/simple/{instance}/in'
      '/pipeline/simple/{instance}/out':           # Connect the pipeline to the output
       - '/offramp/stdout-output/{instance}/in'
      '/pipeline/system::metrics/system/out':
       - '/offramp/stdout-output/{instance}/in'

onramp:
  - id: metronome
    metrics_interval_s: 5
    type: metronome
    config:
      interval: 1000

mapping:
  /binding/example/passthrough:
    instance: "passthrough"

offramp:
  - id: stdout-output
    type: stdout
```
### Log
```
Dec 10 18:36:36 MyHost tremor[3234]: tremor version: 0.9.3
Dec 10 18:36:36 MyHost tremor[3234]: tremor instance: tremor
Dec 10 18:36:36 MyHost tremor[3234]: rd_kafka version: 0x000000ff, 1.5.0
Dec 10 18:36:36 MyHost tremor[3234]: allocator: snmalloc
Dec 10 18:36:36 MyHost tremor[3234]: Listening at: http://0.0.0.0:9898
Dec 10 18:36:37 MyHost tremor[3234]: {"measurement":"ramp_events","tags":{"ramp":"tremor://localhost/onramp/metronome/passthrough/out","port":"in"},"fields":{"count":0},"timestamp":1607643397487725499}
Dec 10 18:36:37 MyHost tremor[3234]: {"onramp":"metronome","ingest_ns":1607643397487722637,"id":0}
Dec 10 18:36:37 MyHost tremor[3234]: {"measurement":"ramp_events","tags":{"ramp":"tremor://localhost/onramp/metronome/passthrough/out","port":"out"},"fields":{"count":0},"timestamp":1607643397487725499}
[...]

PR #648, opened 2020-12-11
Fix visibility of pipeline metrics along with reserved/duplicate node name detection

anupdhml:
# Pull request

Fix generation of pipeline metrics (as reported in https://github.com/tremor-rs/tremor-runtime/issues/647) and improve handing of cases when a pipeline node is created with the same name as an existing one.

## Description

Ensured consistent id for metrics node when it's added to the pipeline graph (as `"metrics"`). Added tests to verify pipeline and ramp metrics as well.

Also throw error and terminate pipeline creation, if a user happens to use id `metrics` for one of their trickle operators (earlier, this was allowed and we would stop getting metrics too) -- this is solved as part of general handling described in https://github.com/tremor-rs/tremor-runtime/issues/650.

For later, we should rename the `events` measurement to reflect that it's pipeline specific (or merge metrics in `ramp_events` under `events`.)

## Related

* Related Issues: fixes https://github.com/tremor-rs/tremor-runtime/issues/647, https://github.com/tremor-rs/tremor-runtime/issues/650

## Checklist

* [x] The code is tested

### DIFF
--- a/tremor-pipeline/src/query.rs
+++ b/tremor-pipeline/src/query.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::errors::{Error, ErrorKind, Result};
-use crate::op::prelude::{IN, OUT};
+use crate::op::prelude::{ERR, IN, METRICS, OUT};
 use crate::op::trickle::select::WindowImpl;
 use crate::{
     common_cow, op, ConfigGraph, NodeConfig, NodeKind, Operator, OperatorNode, PortIndexMap,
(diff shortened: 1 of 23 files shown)


mfelsche:
I think we should definitely throw an error if a user names a node `in`, `out`, `err` or `metrics` as these are reserved port, and thus pipeline operator names.

anupdhml:
Opened https://github.com/tremor-rs/tremor-runtime/issues/650 to generalize error reporting when a node already exists with the given name.

tremor-pipeline/src/query.rs
```diff
@@ -174,51 +183,38 @@ impl Query {
...
-            ..NodeConfig::default()
-        });
-        nodes.insert(OUT, id);
-        *uid += 1;
-        // ALLOW: id is created by inserting above, we also have no other way to access, thanks petgraph
-        let op = pipe_graph[id].to_op(*uid, supported_operators, None, None, None)?;
-        pipe_ops.insert(id, op);
-        outputs.push(id);
-        let mut port_indexes: PortIndexMap = HashMap::new();
+        for (name, node_kind) in &BUILTIN_NODES {
```
mfelsche, line 186: I am not familiar enough with the DAG form yet, but is it possible to be a problem that we insert the `METRICS` node here already instead of after processing the query, as we did before?

mfelsche (approved):
LGTM in general, I especially like the nice errors. 
I just had a tiny question.

tremor-pipeline/src/query.rs
anupdhml: From the perspective of query processing here, we are adding each node independently of another so the order of metrics node addition should not matter. I checked if there's anything in the way the metrics node is accessed later (as part of a running pipeline), that would benefit it from being inserted at the end, and I couldn't find anything suggesting that, so I think this move should be safe to do.

PR #651, opened 2020-12-14
Use new EventId struct instead of Ids

### LINKED ISSUE #584: Ensure that we only emit 1 ack/fail per event, also for batched events.
**Problem**

We need to make sure for all:
* sinks
* operators that might emit acks/fails

that we only ever emit 1 ack/fail per event, this includes batched events.

**Sinks**

- [ ] blackhole
- [ ] debug
- [ ] elastic
- [ ] exit
- [ ] file
- [x] kafka
- [ ] newrelic
- [ ] postgres
- [ ] rest
- [ ] stderr
- [ ] stdout
- [ ] tcp
- [ ] udp
- [ ] ws

**Operators**

TBD

mfelsche:
# Pull request

## Description

This started out as an endeavor to check if all sinks and operators correctly handle CB messages (i.e. only emit 1 ack/fail per event, be it batched, or might it produce multiple protocol data units at the sink layer).

The new EventId struct introduces the stream_id, which is currently unused, but will be put to use in later PRs. The Ids struct was allocating a Vec of 4 elements everytime. Now we get away without additional allocations in the "normal" case (no batching, no grouping, no WAL).

Also make the elastic sink emit insights immediately, not on the next signal.

## Related

* Related Issues: fixes #584

## Checklist

* [x] The code is tested


mfelsche:
@Licenser could you run your benchmark suite on this branch?

And also have a look at the FIXME?

tremor-pipeline/src/op/qos/wal.rs
```diff
@@ -248,12 +248,14 @@ impl WAL {
         Ok(events)
     }
 
-    fn store_event(&mut self, uid: u64, mut event: Event) -> Result<()> {
+    fn store_event(&mut self, source_id: u64, mut event: Event) -> Result<()> {
         let id = self.wal.generate_id()?;
         let write: [u8; 8] = unsafe { mem::transmute(id.to_be()) };
-        event.id.add_id(uid, id); // = Ids::new(uid, id);
+        // TODO: figure out if handling of separate streams makes sense here
+        // FIXME: this should go into op_meta, shouldnt it?
```
mfelsche, line 255: cc @Licenser

src/sink/elastic.rs
```diff
@@ -74,14 +74,14 @@ impl offramp::Impl for Elastic {
...
+            let (tx, _rx) = bounded(1); // dummy value
 
             Ok(SinkManager::new_box(Self {
+                sink_url: TremorURL::from_offramp_id("elastic")?, // just a dummy value, gonna be overwritten on init
                 postprocessors: vec![],
                 client,
                 queue,
-                rx,
-                tx,
+                insight_tx: tx,
```
mfelsche, line 84: Oops, i moved elastic from the old `drain_insights` pattern to emit insights immediately.

tremor-pipeline/src/op/qos/wal.rs
Licenser: The WAL acknowledges the event downstream (to the source) so it becomes a 'new' event the moment it enters the WAL That was the reason it got a new ID

mfelsche: makes sense, now i know what to do here :)

Licenser (approved):
:+1: LGTM great job on the coverage and perf :D

PR #676, opened 2021-01-06
Fix missing newline in error messages for some sources.

mfelsche:
# Pull request

## Description

If the source line with the error didnt end with a newline (e.g. last line), we missed to add one, so the callout starts on its own line. This is fixed here (with some horrible workaround for term highlighting).

## Checklist

* [x] The code is tested


Licenser (changes requested):
Aside from some tests and a CHANGELOG entry 👍

Licenser (approved):
Perfect 👍

PR #709, opened 2021-01-22
Workflow for adding packages to releases

Licenser:
# Pull request

This adds a workflow to automatically generate packages for releases

## Description

This changes the default LTO type from `true` to `thin` which is a slightly less optimized version. There are three reasons for this:

1) the CI can't build `lto=true` as it consumes too much memory
2) we have been publishing `lto="thin"` as part of all the docker builds already. Docker builds so far have been our release artifact, so this has no performance impact on endusers.
3) It standardizes performance between benchmarks published artifacts

## Checklist

* [x] The code is tested


.github/workflows/release.yaml
```diff
@@ -0,0 +1,22 @@
...
+  package:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Build releases
+        run: make builds
+      - name: Creating release tarballs
+        run: make archives
+      - name: Creating release packages
+        run: make packages
```
anupdhml, line 17: `make packages` already builds the archives too so we can remove the archive step

anupdhml (changes requested):

anupdhml (approved):

PR #720, opened 2021-01-27
Fix ssl issue and improve musl builds

Licenser:
# Pull request

## Description

This adds improvements to the package build process to statically link libssl to make the binaries more portable.

## Checklist

* [x] The code is tested

### DIFF
--- a/.depignore
+++ b/.depignore
@@ -1,0 +1,1 @@
+openssl
--- a/packaging/builder-images/Dockerfile.x86_64-unknown-linux-musl
+++ b/packaging/builder-images/Dockerfile.x86_64-unknown-linux-musl
@@ -2,7 +2,7 @@ ARG RUST_VERSION
 
 # update the version here as needed
 # via https://hub.docker.com/_/rust
-FROM rust:${RUST_VERSION}-buster
+FROM rustembedded/cross:x86_64-unknown-linux-musl-0.2.1
 
 # install musl
 # this setup borrowed from https://github.com/rust-embedded/cross/blob/v0.2.0/docker/musl.sh#L44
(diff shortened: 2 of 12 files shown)


.depignore
```diff
@@ -0,0 +1 @@
+openssl
```
mfelsche, line 1: what is this file for?

packaging/builder-images/Dockerfile.x86_64-unknown-linux-musl
```diff
@@ -16,21 +16,33 @@ RUN temp_dir=$(mktemp -d) \
...
+ENV CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER=x86_64-linux-musl-gcc \
+  CC_x86_64_unknown_linux_musl=x86_64-linux-musl-gcc \
+  CXX_x86_64_unknown_linux_musl=x86_64-linux-musl-g++
 
-COPY shared/install_dependencies_debian.sh /
-RUN /install_dependencies_debian.sh
 
-COPY shared/entrypoint.sh /
-ENTRYPOINT [ "/entrypoint.sh" ]
+# COPY shared/entrypoint.sh /
```
mfelsche, line 47: commented out on purpose?

mfelsche (approved):
This is hard to verify for me, but overall it looks legit :)

Crazy magic stuff.

.depignore
Licenser: we check for deps that are unused to purn the dependency graph, that works on a heuristic that the dependency needs to be used in our code. openssl isn't used i our code directly but we need to include it as a dependency for static linking.

To solve the problem of `openssl` being rightfully claimed as a unused dependency but still needed to be a dependency I added the `.depignore` file to ignore dependencies that are in there.

packaging/builder-images/Dockerfile.x86_64-unknown-linux-musl
Licenser: Ja I didn't wanted to loose those lines, right now musl is blocked on the snmalloc issue: https://github.com/SchrodingerZhu/snmalloc-rs/issues/102

the Dockerfile isn't working as it is right now and it's unclear what is needed to get it working once the snmalloc issue is resolved so I wanted to keep the 'old' docker bits in there as a comment in case it turns later out we needed them after all.

PR #729, opened 2021-01-29
String interpolation new

### LINKED ISSUE #726: Issues with 0.9 and earlier string interpolation
**Problem**

There is an issue with string interpolation that does fail to create some strings sensibly. In addition, having a `{` and `}` as the last element of  a string fails

```tremor
# this fails
"\}"
# this fails too in different ways
"\{\}"
```

**Steps**

1. Create a string with an escaped `{` in it but no interpolation
2. Create a string with an escaped `{` in it but with interpolation
3. The escaped `{` on both is handled different

**Possible Solution(s)**

Implement the string interpolation RFC https://github.com/tremor-rs/tremor-rfcs/pull/34

Licenser:
# Pull request

## Description

This implements the RFC for string interpolation

## Related

* [RFC](https://github.com/tremor-rs/tremor-rfcs/pull/34)
* Related Issues: fixes #726
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/71)

## Checklist

* [x] The code is tested


mfelsche:
Question that comes to mind is: What error will a string like `"#{}"` give?

tremor-script/src/lexer.rs
```diff
@@ -1824,300 +1824,270 @@ impl<'input> Lexer<'input> {
...
-                    } else {
-                        has_escapes = true;
-                        heredoc_content.push(c);
-                        end = e;
-                        end.shift(c);
-                    }
+                        Token::RBrace => {
+                            pcount -= 1;
+                        }
+                        Token::LBrace | Token::Interpol => {
```
mfelsche, line 1872: do we really need `Token::LBrace` here anymore? `Interpol` should suffice, no?

mfelsche (approved):
The test error messages look good, so just one issue with me understanding the logic at one point. Otherwise LGTM!

tremor-script/src/lexer.rs
Licenser: We do since we need to count them in something like:

```tremor
"""
this is a #{  { "here": "doc" } }!
"""
```

Licenser:
> 
> 
> Question that comes to mind is: What error will a string like `"#{}"` give?

Good question :D and great catch I didn't think about that, will add a test tomorrow!

Licenser:
I added a new (nicer) error for the "#{}" case :+1: good catch @mfelsche !

tests/script_errors/string_interpolation_empty/error.txt
```diff
@@ -0,0 +1,4 @@
+Error: 
+    1 | "#{}"
+      |  ^^^ You have an interpolation without content.
```
mfelsche, line 3: 😻

tests/script_runtime_errors/recursion_limit/error.txt
```diff
@@ -0,0 +1,5 @@
+Error: 
+    4 | end;
+    5 | 
+    6 | boom(1)
```
mfelsche, line 4: 💥

mfelsche (approved):

PR #731, opened 2021-01-29
Emit tumbling window based on size immediately when full.

### LINKED ISSUE #728: Tumbling windows based on size should emit when full
**Problem**

Tumbling windows based on size only emit when the next event comes in after they are full.

**Steps**

1. Define a window based on size with `size = 3`
2. Pump 3 events through it
3. It doesnt emit
4. Pump another event through it.
5. It now emits.

**Possible Solution(s)**

A Sized based tumbling window can already emit, when it is full.

**Notes**

Output of `rustup --version`:
Output of `rustup show`:
Output of `tremor --version`:

mfelsche:
# Pull request

## Description

Emit tumbling window based on size immediately when full, do not wait for the next event.

## Related

* Related Issues: fixes #728 
* Related docs PRs: 
  * https://github.com/tremor-rs/tremor-www-docs/pull/73
  * https://github.com/tremor-rs/tremor-www-docs/pull/66

## Checklist

* [x] The code is tested


tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                        let result = stmt.target.run(
+                            opts,
+                            &env,
+                            unwind_event,
+                            state,
+                            event_meta,
+                            &local_stack,
+                        )?;
+
+                        if let Some(guard) = &stmt.maybe_having {
```
Licenser, line 721: could we extract this into a function?

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                            let test =
+                                guard.run(opts, &env, &result, state, &NULL, &local_stack)?;
+                            if let Some(test) = test.as_bool() {
+                                if !test {
+                                    continue;
+                                }
+                            } else {
+                                let s: &Select = &stmt;
+                                return tremor_script::errors::query_guard_not_bool(
```
Licenser, line 730: could we add a query_guard_not_bool_err function that returns the Error instead of Result to avoid return ...?;?

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                    let this_groups = &mut window.dims.groups;
+                    let last_groups = &mut window.last_dims.groups;
+                    let window_impl = &window.window_impl;
                     let mut idgen = self.event_id_gen;
-                    let (_, prev_group) = prev_groups
+                    let (_, this_group) = this_groups
                         .raw_entry_mut()
                         .from_key(&group_str)
                         .or_insert_with(|| {
+                            // This is sound since we only add mutability to groups
```
Licenser, line 691: I don't think we need this as there is no unsafe code here.
```suggestion
```

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                    //   init
+                    // accumulate
+                    // if emit && include
+                    //   push
+                    //   init
+
+                    // ALLOW: we verified that an element exists
+                    let window = &mut self.windows[0];
+                    set_window(consts, Value::from(window.name.to_string()))?;
+                    // This is sound since we only add mutability to groups
```
Licenser, line 682: I don't think we need this as there is no unsafe code here.
```suggestion
```

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                            OUT,
+                            Event {
+                                id: this_group.id.clone(),
+                                ingest_ns: event.ingest_ns,
+                                // TODO: this will blindly chose the origin uri of the last event arriving at the window, which is at least misleading and might be just wrong
+                                // e.g. in case of having two source input events into a stream we select from
+                                // having a single origin_uri doesnt make sense here anymore, if this info is valuable, we should track multiple origins for such events
+                                //origin_uri: event.origin_uri.clone(),
+                                origin_uri: None,
+                                is_batch: event.is_batch,
```
Licenser, line 746: I think this might be off, since it's not strictly batched by a batching function and thus might not be able to be unbatched, we should set this to false

mfelsche: yep, we should consider this a totally new event. Shake off your heritage, little event. be free! go brrrr!

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
```
Licenser, line 839: see above

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                                let s: &Select = &stmt;
+                                return tremor_script::errors::query_guard_not_bool(
+                                    s, guard, &test, &node_meta,
+                                )?;
+                            }
+                        }
+                        let result = result.into_owned();
+                        events.push((
+                            OUT,
+                            Event {
```
Licenser, line 831: would it make sense to extract this into a function?

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                                        window: window_impl.clone(),
+                                        aggrs: aggregates.clone(),
+                                        group: group_value.clone_static(),
+                                        id: idgen.next_id(), // after all this is a new event
+                                    }),
+                                )
+                            });
+
+                        if window_event.include {
+                            // add this event to the aggr state **BEFORE** emit and propagate to next windows
```
Licenser, line 1020: I'm a big confused by this, what's scratch?

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -646,221 +648,647 @@ impl Operator for TrickleSelect {
...
+                                //          swap(scratch, aggrs)
+                                std::mem::swap(&mut scratch, &mut this_group.aggrs); // store current state before init
+                                                                                     //          aggrs.init()
+                                for aggr in &mut this_group.aggrs {
+                                    aggr.invocable.init();
+                                }
+                            } else {
+                                // not first window
+                                //          local_scratch = aggregates.clone()
+                                let mut local_scratch = aggregates.clone(); // FIXME: avoid clone and just add another scratch value?
```
Licenser, line 1148: cloning aggregates every time will be a significant hit on perf :(

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -552,79 +599,32 @@ impl Operator for TrickleSelect {
         }
 
         let group_values: Vec<Value> = group_values.into_iter().map(Value::Array).collect();
+        let mut scratch = aggregates.clone(); // TODO: move to statement level to avoid overhead per event
```
Licenser, line 602: cloning aggregates on every emit seems extremely expensive

tremor-pipeline/src/query.rs
```diff
@@ -113,7 +114,19 @@ fn window_decl_to_impl<'script>(
                         )
                     },
                     |interval| {
-                        Ok(TumblingWindowOnTime::from_stmt(interval, ttl, script, stmt).into())
+                        let emit_on_tick = d
+                            .params
+                            .get("emit_on_tick")
+                            .and_then(Value::as_bool)
+                            .unwrap_or_else(|| script.is_none());
```
Licenser, line 121: it feels wrong that the default is different based on script being defined, should it perhaps always be false? Or simply forbidden ith scripts?

tremor-script/src/std_lib/stats.rs
```diff
@@ -89,6 +89,7 @@ impl TremorAggrFn for Sum {
             },
             |v| {
                 self.0 += v;
+                dbg!(self.0);
```
Licenser, line 92: :hocho:

tremor-pipeline/src/op/trickle/select.rs
mfelsche: It is a `Vec<InvokeAggrFn>`, a scratch value for the aggregation states.
When we have multiple windows, when the current window emits, we cannot just simply merge the local aggr states into the next window, as this window might need to emit itself before it incorporates the new state. Complicated shizzle. So the only safe way i found is to have basically a tmp value around, where we store the current windows state, before we initialize for the next window. When the next window is ready, it merges the state from the scratch value.
We only ever need max 2 scratch "registers" for the whole select statement execution at any time, i am thinking about putting it into the `SelectStatement` struct, but wanted to ensure my approach flies first.

tremor-pipeline/src/op/trickle/select.rs
mfelsche: This is on every event only :P
But yeah, this caused me lots of pain and will be fixed.

tremor-script/src/std_lib/stats.rs
mfelsche: 😱

tremor-pipeline/src/op/trickle/select.rs
mfelsche: we need the `?` as the error types differ i think.

tremor-pipeline/src/query.rs
mfelsche: if the script is not defined we definitely use the `ingest_ns`. In that case the tick-time equals the `ingest_ns` time. We can be sure that when a tick exceeded the time for this window, no event will ever come for that window again. This is why i chose `true` for the case of `script.is_none()` here. I hope my assumption is correct.

tremor-pipeline/src/op/trickle/select.rs
Licenser: `.into()` should do the same and hopefully be more readable.

tremor-pipeline/src/query.rs
Licenser: It shouldn't hold off a merge but it's worth thinking if we perhaps just scrap it? I see that if `script.is_none()` we always ignore it (in the other code too) so the only time this has an effect if is `script.is_some()` and it is set to `false`.


My main concern is that this setting has so many interactions with different parts that it'll be hard to explain to operators, I've a hard time wrapping my head around it but the best explanation I came up with is:


> For size-based windows it has no effect. If a script is not set it has no effect. If a script is set and it is set to false it has no effect. If a script is set and it is set to true it will still emit on ticks.

Is that correct? It feels like a very shoot in the footy kind of setting I don't see a lot of use for, might be worth just leaving it out and have the default behavior of `"If you set a script we can't track your time"` to make it easier to gasp?

tremor-pipeline/src/op/trickle/select.rs
Licenser: Ah I think I get it. Let me try to paraphrase. We clone the value we want to merge so we can reinitialize the old window (kill it's state) emit the pre-merge state of the next without?

tests/query.rs
```diff
@@ -80,8 +80,10 @@ macro_rules! test_cases {
                 assert_eq!(results.len(), out_json.len(), "Number of events differ error");
                 for (_, result) in results {
                     for value in result.value_iter() {
+                        let serialized = sorted_serialize(value)?;
+                        println!("{}", &serialized);
```
Licenser, line 84: :hocho:

tremor-pipeline/src/query.rs
mfelsche: I would have explained it as such in the docs:

```
Windows that use the internal event ingest timestamp as their time source can be tracked and emit with a 100ms resolution if no events arrive. Windows using `script` to configure a custom timestamp cannot be tracked, as the given timestamp might not correlate with the wall-clock time used for the ingest timestamps. If your custom `script` timestamp can be tracked with wall-clock, you can enable the window by setting `emit_on_tick = true`. That makes sense if the timestamp represents a monotonic timestamp that might be off by some constant from the event ingest timestamp.
```

It requires a lot of explanation tbh.
We can totally strip that thing.

Licenser: 👍 let's get rid of it!

Licenser (approved):
LGTM! Darn good work :+1:

PR #740, opened 2021-02-05
Ensure Cargo.lock is up to date

Licenser:
# Pull request

## Description

Ensure that the `Cargo.lock` is always up to date in a pull request

## Checklist
* [x] The code is tested


mfelsche:
What if we need to change the lockfile? That is when we add a dependency etc. This might break CI for dependabot too, no?

I do love the clippy part though ❤️

mfelsche:
Ah no, got it. The build and test run should not change the Cargo.lock from what is checked into git. all good.

mfelsche (approved):

Licenser:
I tested with an outdated `Cargo.lock` and it failed as expected, to keep the history clean I squashed that before the merge.

darach (approved):
LGTM 👍

PR #762, opened 2021-02-17
Add hygenic errors for server run -f

### LINKED ISSUE #761: No hygenic errors in `tremor server run`
**Problem**

when providing a pipeline to `tremor server run` that failed to be parsed or compiled we do not get a hygenic error. Instead, we get a generic error that describes what is wrong but not where or even in which file.

```
[2021-02-17T13:18:38Z ERROR tremor::server] error: Found the token `` but expected one of `!`, `\``, `$`, `(`, `+`, `-`, `<<`, `<ident>`, `[`, `absent`, `args`, `bool`, `event`, `float`, `for`, `group`, `heredoc`, `int`, `match`, `merge`, `nil`, `not`, `patch`, `present`, `recur`, `state`, `window`, `{`
error: Found the token `` but expected one of `!`, `\``, `$`, `(`, `+`, `-`, `<<`, `<ident>`, `[`, `absent`, `args`, `bool`, `event`, `float`, `for`, `group`, `heredoc`, `int`, `match`, `merge`, `nil`, `not`, `patch`, `present`, `recur`, `state`, `window`, `{`
```

**Steps**
1. create a bad file (i.e. `bad.trickle`: `select`)
2. execute the file `tremor server run -f bad.trickle`
3. observe the output

**Possible Solution(s)**

Add hygenic errors to server run file loading

Licenser:

# Pull request

## Description

Fixes the missing Hygenic errors for `server run`

## Related

* Related Issues: fixes #761

## Checklist

* [x] The code is tested

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,6 +86,8 @@ pub(crate) use serde_yaml::Value as OpConfig;
 use system::World;
 pub(crate) use tremor_pipeline::Event;
 use tremor_pipeline::{query::Query, FN_REGISTRY};
+use tremor_script::highlighter::Term as TermHighlighter;
+use tremor_script::Script;
 
 /// Default Q Size
 pub const QSIZE: usize = 128;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -159,7 +161,18 @@ pub async fn load_query_file(world: &World, file_name: &str) -> Result<usize> {
...
         &*FN_REGISTRY.lock()?,
         &aggr_reg,
-    )?;
+    );
+    let query = match query {
+        Ok(query) => query,
+        Err(e) => {
+            let mut h = TermHighlighter::stderr();
+            if let Err(e) = Script::format_error_from_script(&raw, &mut h, &e) {
+                eprintln!("Error: {}", e);
```
mfelsche, line 170: should this rather be logged?

Licenser: The error only occurs if the highlighter couldn't highlight, so in any likelihood, if we ever hit this, something is seriously wrong. so stderr is the most likely thing still working.

Licenser: I think the only time this error is hit is when the writer can't allocate memory

mfelsche: Ah yeah! What i actually meant was that the highlighter is writing to stderr always, but we might not have an stderr around, depending on how tremor was started. But as long as we stop the program when this error hits, this is fine.

mfelsche (approved):

PR #764, opened 2021-02-18
Make elastic sink handle empty batches

mfelsche:
# Pull request

## Description

* Make elastic sink handle empty batches
* Send response if we run into issues building ES payload (serialize value)

## Checklist

* [x] The code is tested


tremor-pipeline/src/op/generic/batch.rs
```diff
@@ -151,31 +151,33 @@ impl Operator for Batch {
...
-            || Ok(EventAndInsights::default()),
-            |delay_ns| {
+        Ok(self
+            .max_delay_ns
+            .map_or_else(EventAndInsights::default, |delay_ns| {
                 if signal.ingest_ns - self.first_ns > delay_ns {
                     // We don't want to modify the original signal we clone it to
                     // create a new event.
 
                     let mut data = empty();
```
Licenser, line 161: could we move that down it feels like we don't need to swamp memory and allocate if there is nothing to swap

Licenser (dismissed):
Nice!   :+1:

Licenser (approved):

PR #766, opened 2021-02-18
Modify Kafka sink/source meta to include message headers

### LINKED ISSUE #737: Support headers for kafka source and sink

Newer versions of Kafka support message headers, we should have a way in tremor to inspect, add, remove and modify them.

In line with the REST and ES offramp, hearers should be treated as metadata in their own sub-field. We could use:

```tremor
let $kafka = {"header1": "value", "h2":"..."};
```

**Notes**

* [Kafka Headers](https://kafka.apache.org/20/javadoc/index.html?org/apache/kafka/connect/header/Header.html#:~:text=A%20Header%20is%20a%20key,This%20is%20an%20immutable%20interface.)
* [librdkafka-rs interface](https://docs.rs/rdkafka/0.25.0/rdkafka/message/trait.Message.html#tymethod.headers)

aryan9600:
# Pull request

## Description
Include Kafka message headers as a meta variable for the Kafka source and sink.

## Related

* Related Issues: fixes #737 

## Checklist

* [x] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New features
 
 * Allow using '_' as seperators in numeric literals [#645](https://github.com/tremor-rs/tremor-runtime/issues/645)
+* Add support for Kafka message headers, available through the `$kafka_headers` metadata variable.
 
 ## 0.10.2
 
--- a/src/sink/kafka.rs
+++ b/src/sink/kafka.rs
@@ -28,6 +28,7 @@ use halfbrown::HashMap;
 use rdkafka::config::ClientConfig;
 use rdkafka::{
     error::KafkaError,
+    message::OwnedHeaders,
     producer::{FutureProducer, FutureRecord},
 };
 use std::{fmt, time::Duration};
@@ -246,6 +247,7 @@ impl Sink for Kafka {
             let encoded = codec.encode(value)?;
             let processed = postprocess(self.postprocessors.as_mut_slice(), ingest_ns, encoded)?;
             let meta_kafka_key = meta.get_str("kafka_key");
+            let meta_kafka_headers = meta.get_object("kafka_headers");
             for payload in processed {
                 // TODO: allow defining partition and timestamp in meta
                 let mut record = FutureRecord::to(self.config.topic.as_str());
@@ -255,7 +257,15 @@ impl Sink for Kafka {
                 } else if let Some(kafka_key) = &self.config.key {
                     record = record.key(kafka_key.as_str());
                 }
-
+                if let Some(kafka_headers) = meta_kafka_headers {
+                    let mut headers = OwnedHeaders::new_with_capacity(kafka_headers.len());
+                    for (key, val) in kafka_headers.iter() {
+                        if let Some(val_str) = val.as_str() {
+                            headers = headers.add(key, val_str);
+                        }
+                    }
+                    record = record.headers(headers);
+                }
                 // send out without blocking on delivery
                 match self.producer.send_result(record) {
                     Ok(delivery_future) => {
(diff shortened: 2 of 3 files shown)


src/sink/kafka.rs
```diff
@@ -245,7 +246,8 @@ impl Sink for Kafka {
         for (value, meta) in event.value_meta_iter() {
             let encoded = codec.encode(value)?;
             let processed = postprocess(self.postprocessors.as_mut_slice(), ingest_ns, encoded)?;
-            let meta_kafka_key = meta.get_str("kafka_key");
+            let meta_kafka_key = meta.get("kafka_key").and_then(Value::as_str);
```
mfelsche, line 249: i think we recently added a `.get_str()` method that can replace this pattern.

src/sink/kafka.rs
```diff
@@ -245,7 +246,8 @@ impl Sink for Kafka {
...
+            let meta_kafka_headers = meta.get("kafka_headers").and_then(Value::as_object);
```
mfelsche, line 250: and a `get_object` i think :)

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.10.3
+
+### New features
+
+* Add support for Kafka message headers, available through the `$kafka_headers` metadata variable.
```
mfelsche, line 7: Can we use `$kafka.headers` (a nested object) for the headers in the metadata? The reason is that our goal is to properly namespace all onramp/offramp specific metadata with the onramp/offramp id, so we dont pollute the keyspace here and we are free to add additional metadata values in a clean way into the `$kafka` object next to the `headers`.

mfelsche (commented):
This is a great PR! Love it. I would only ask you to change the key for the headers, otherwise this is good to go :)

Thank you, this is a great contribution!

CHANGELOG.md
mfelsche: I know this introduces an inconsistency between `$kafka_key` and `$kafka_headers`. Feel free to also put the `key` into the `$kafka` object in this PR, otherwise we tackle this with a follow up PR.

src/sink/kafka.rs
Licenser: I found that for objects, `get` followed by `unwrap_or_default()` is the easiest way, preserving the `value` allows to use `get_<type>` afterward and if the result is `default` it's `null` and all successive calls to `get_` will return `None`

CHANGELOG.md
aryan9600: So I was thinking of having the variable as `$kafka_headers` in this PR, and then in a follow-up PR, we could have both `$kafka_key` and `$kafka_headers` inside the `$kafka` variable, which can then be used through `$kafka.headers` and `$kafka.key`, since the issue that PR is tackling is mainly related to message headers.
If it's better to have this change in this PR itself, I'll update it accordingly, no issues! :)

mfelsche: That is also fine, keeping the PRs small.

src/sink/kafka.rs
aryan9600: I don't think we can use `unwrap_or_default()` as `get` returns a `Value`, and I don't think any structs implementing `Value` have `Default` as derivable. We could opt for using `unwrap_or` and return `None`, if there is no `Some` value though. I'm still fairly new to the codebase, so please correct me if I'm wrong here :)

Licenser: If Default isn't implemented that would be an oversight on my part :) Value should default to 'null' that might have slipped away in some copy & pasta. Sorry if it's missing

mfelsche:
Yay, done! Please rebase, and then i will approve :)

Licenser:
Thank you so much :+1: great contribution :) ping me on Discord with your address if you want a tremor sticker or two :)

mfelsche:
Damn, i think you need to rebase again, I just merged another contribution. Sorry for the hassle!

aryan9600:
Haha, no issues! :)

Licenser (approved):

PR #775, opened 2021-02-22
Add KV connector

Licenser:
# Pull request

## Description

This adds a `kv` connector. It allows issuing 'commands' such as `put`, `get`, `delete`, `cas`, and `scan` with replies being sent back `out` of the connector.

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/82)

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/kv.rs
+++ b/src/sink/kv.rs
@@ -1,0 +1,275 @@
+// Copyright 2020-2021, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO: Add correlation of reads and replies.
+
+#![cfg(not(tarpaulin_include))]
+use crate::sink::{prelude::*, Reply};
+use crate::source::prelude::*;
+use async_channel::Sender;
+use halfbrown::HashMap;
+use serde::Deserialize;
+use sled::{CompareAndSwapError, IVec};
+use std::boxed::Box;
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct Config {
+    dir: String,
+}
+pub struct Kv {
+    sink_url: TremorURL,
+    event_origin_uri: EventOriginUri,
+    db: sled::Db,
+}
+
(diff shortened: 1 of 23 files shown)


src/sink/kv.rs
```diff
@@ -0,0 +1,275 @@
...
+use crate::source::prelude::*;
+use async_channel::Sender;
+use halfbrown::HashMap;
+use serde::Deserialize;
+use sled::{CompareAndSwapError, IVec};
+use std::boxed::Box;
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct Config {
+    dir: String,
```
mfelsche, line 28: what about a config like in the `WAL` operator `flush_on_event` that will flush the kv store on every event?

src/sink/kv.rs
```diff
@@ -0,0 +1,275 @@
...
+                    end: s.get_bytes("end").map(|v| v.to_vec()),
+                })
+            } else {
+                error!("failed to decode command: {}", v);
+                None
+            }
+        });
+
+        for c in cmds {
+            println!("command: {:?}", c);
```
mfelsche, line 230: 🙀

src/sink/kv.rs
mfelsche: or an explicit `flush` command?

src/sink/kv.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    fn auto_ack(&self) -> bool {
+        true
+    }
+
+    fn default_codec(&self) -> &str {
+        "json"
+    }
+
+    async fn terminate(&mut self) {}
```
mfelsche, line 274: should we flush the sled db here?

mfelsche (commented):
Awesome! Unfortunately I am in a rather picky mood ;-)

src/sink/kv.rs
Licenser: :+1:

src/sink/kv.rs
Licenser: I'm 99% sure that sled flushes when it's dropped

src/sink/kv.rs
Licenser: I wanted to avoid that complexity as the requirements are not that of a WAL, I fear that will add more confusion and misuse than just giving one default option. We want this to be accessible to people without making them read on the details of the behavior of storage engine flushes, how their particular OS is working and what the effect between those will be.

src/sink/kv.rs
mfelsche: i couldnt find anything in the docs or sled sources. Any pointer for me?

Licenser: It's not spelled out but implied in the examples, also it is what the code does :D (hard to find :sob:) https://docs.rs/sled/0.34.6/src/sled/tree.rs.html#98

We should probably ask it to be mentioned explicitly in the sled-documentation to ensure that it is part of the API contract.

mfelsche (approved):

PR #778, opened 2021-02-23
Emit concurrent request overflow errors to ERR port 

### LINKED ISSUE #752: Make the sinks that limit concurrency emit an error event when they overflow

Some of our sinks/offramps (e.g. rest and elastic) limit the number of concurrent in flight outbound requests by the config parameter `concurrency`. When those sinks receive an event, they check if there is a concurrent slot for them. If so these events are handled in a newly spawned asynchronous task and the next event can be pulled by the task driving the sink/offramp. If the concurrency is exhausted, the event is discarded, an error is logged and a `CB::Fail` event is triggered.
It would be nice if we had error events emitted in that case, so users can react on that in their pipeline/application code.

Make the rest and elastic sink (and all other that limit concurrency) emit an error event to the `err` port when an event failed due to the sink being busy. Have those events have a consistent format, so users can easily match on it and tell them apart from other error events.

aryan9600:

# Pull request

## Description

Emit an error to the ERR port, whenever the concurrency limit of the elastic and rest offramps is exceeded, allowing users to react and recover accordingly.

## Related

* Related Issues: fixes #752 
* Related Docs: https://github.com/tremor-rs/tremor-www-docs/pull/81

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/rest.rs
+++ b/src/sink/rest.rs
@@ -264,7 +264,7 @@ enum CodecTaskInMsg {
         response: Response,
         duration: u64,
     },
-    ReportFailure(EventId, OpMeta, EventOriginUri, Error),
+    ReportFailure(EventId, OpMeta, EventOriginUri, Error, u16),
 }
 
 enum SendTaskInMsg {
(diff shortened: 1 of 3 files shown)


src/sink/rest.rs
```diff
@@ -391,6 +393,27 @@ impl Sink for Rest {
             Ok(None)
         } else {
             error!("Dropped data due to overload");
+            let mut error_ids = EventIdGenerator::new(sink_uid);
```
mfelsche, line 396: can you put this into the struct? if used this way, every error event will have the same id.

mfelsche (commented):
Great contribution, thanks for taking that on!

Just one thing, and then we are good to go, see my comment.

src/sink/rest.rs
aryan9600: oh yes that makes sense, thanks!

src/sink/rest.rs
```diff
@@ -391,6 +395,26 @@ impl Sink for Rest {
             Ok(None)
         } else {
             error!("Dropped data due to overload");
+            let error_event = create_error_response(
+                self.error_id_gen.next_id(),
```
mfelsche, line 399: I am sorry for the back and forth, but upon inspecting the existing code more closely, there is already an `EventIdGenerator` in the codec_task (line 511) and these to would possibly conflict.

I think what should instead be the best solution here is to send `CodecTaskInMsg::ReportFailure` to the `codec_task_channel`. The codec task will then send a `CB::Fail` and an error_response. In this branch here in `on_event` we should not return an `Err` as this would trigger another `CB::Fail` to be sent. Make it an `Ok(None)` in that case.

Again, sorry, I just remembered now how it all works.

mfelsche: The rest sink is a beast!

aryan9600: Hey sorry for the late reply, uni midterms going on 😩 
Sending `CodecTaskInMsg::ReportFailure` to `codec_task_channel` sounds good to me except for the fact that, the implementation of `codec_task()` always sends a 503 as the status code, which is _technically_ wrong here, since [429 is the status code for too many requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429).
Are we fine with sending 503 as of now, or should we think about modifying the `CodecTaskInMsg::ReportFailure` type to have a `u16` for status codes?

mfelsche: Yeah, i think you are right, we should definitely return a 429 to the client.
Adding a status field to the `CodecTaskInMsg::ReportFailure` sounds like the right thing to do :).
Great suggestion.

mfelsche (approved):
Cool!

mfelsche:
this only needs a rebase and a `rustfmt` run and we are good to merge! :)

mfelsche (approved):

PR #779, opened 2021-02-23
Go back to polling for the kafka onramp

mfelsche:
# Pull request

## Description

* Do not use the async StreamConsumer and MessageStream for the kafka onramp anymore. It is hanging if there is no message in the kafka queue, hindering the runtime from consuming control-plane and CB events. This PR uses the `BaseConsumer` and on each `pull_event` it is polling kafka for a new message.
* Make the `poll_interval` configurable for the kafka onramp. People might want to set it to lower values if they want to be informed faster of incoming messages when their queue has been empty before.
* Add the `retry_failed_events` setting (default `true`), that if set to false, will not seek back to the offset of the failed event it received, but keep on going. This means those failed events remain uncommitted and will only ever be retried if tremor restarts.
* Add the `cb` offramp that sends back `CB events` as requested in the event metadata or payload. E.g. `{"cb": "ack"}` Mostly for testing onramp/operator behaviour with CB events.
* Make the `kafka` onramp fail upon initialization if it couldnt connect to the cluster or couldnt subscribe to all configured topics.

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/80)

## Checklist

* [x] The code is tested


CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
 
 * Allow using '_' as seperators in numeric literals [#645](https://github.com/tremor-rs/tremor-runtime/issues/645)
 * Add support for Kafka message headers, available through the `$kafka_headers` metadata variable.
+* Add the `cb` offramp for testing upstream circuit breaker behaviour [#779](https://github.com/tremor-rs/tremor-runtime/pull/779)
```
Licenser, line 9: I think this deserves some docs too :)

src/lib.rs
```diff
@@ -29,8 +29,8 @@ extern crate serde_derive;
 extern crate log;
 #[macro_use]
 extern crate lazy_static;
-#[macro_use]
-extern crate rental;
+//#[macro_use]
+//extern crate rental;
```
Licenser, line 33:
```suggestion
```

CHANGELOG.md
Licenser: I'll move that over to the docks PR :)

Licenser (approved):
I love how this simplifies the kafka source! :heart: :+1:

PR #787, opened 2021-02-26
Fix systemd file

### LINKED ISSUE #784: systemd starting up tremor without any configs loaded
**Problem**
Installed tremor 0.10.2 using rpm and tried to start the service. Initially it did not start because of missing home directory (https://github.com/tremor-rs/tremor-runtime/issues/782). After that is resolved, tremor starts up with only the logging config.

**Steps**

1. Install 0.10.2 version from rpm (haven't tried any other method)
2. mkdir /home/tremor; chown tremor /home/tremor
3. systemctl start tremor

**Possible Solution(s)**
current systemd start command says
```
ExecStart=/usr/bin/tremor server run --logger-config /etc/tremor/logger.yaml
```
If there is a specific default config file, it has to be supplied along with it. Ex:
```
ExecStart=/usr/bin/tremor server run --logger-config /etc/tremor/logger.yaml -f /etc/tremor/config/main.yaml
```

**Notes**

Output of `rustup --version`:
Output of `rustup show`:
Output of `tremor --version`: 0.10.2

Licenser:
# Pull request

## Description

Fix systemd file to load config on startup

## Related

* Related Issues: fixes #784

## Checklist

* [x] The code is tested


jemshad:
LGTM!

mfelsche:
I think the shell globbing will error if there is no .trickle or no .yaml file in that directory, wont it? 

We could either add an option to tremor to use a directory and load all files in it and use this option here

or

we could have a startup shellscript that is called here that enlists all yaml and trickle files with an -f each, and doesnt add anything if no file is present.

I would root for option 1.

Licenser:
That's a good observation, darn.

I'm honestly not a huge fan of adding "load this directory" option. It goes a bit against the Unix philosophy of doing a thing and doing it well. There are existing tools to do directory loading so they should be used. Doing the same thing twice also is somewhat icky, now we have two ways to solve the same problem adding to the complexity of the CLI. Plus adding code to the project "because systemd" also feels fairly wrong.

I'd be in favor of a startup script, as it only needs to exist in combination with systems and can be kept out of the core codebase.

packaging/distribution/etc/systemd/system/tremor.service
```diff
@@ -7,7 +7,7 @@ Requires=network-online.target
 [Service]
 User=tremor
 Group=tremor
-ExecStart=/usr/bin/tremor server run --logger-config /etc/tremor/logger.yaml
+ExecStart=/usr/lib/tremor/tremor.service
```
mfelsche, line 10: do we really wanna name this shellscript `.service` ? this seems confusing to me, as systemd service files already have the `.service` suffix.

mfelsche (commented):
Looking good, except a tiny tiny thing.

packaging/distribution/etc/systemd/system/tremor.service
Licenser: I'm indifferent to the name, it's used in the `service`, it starts the `service` so I figured `service` is a good name :D  but if you got a better name I'm happy to change it.

mfelsche: what about `tremor_systemd_exec_start.sh`? or just `tremor.sh`?

Licenser: changed it to `tremor.sh` :+1: I think I got them all :tm: but if you can look over the files to make sure I didn't missed any renames :D

mfelsche (approved):
Perfection!

PR #800, opened 2021-03-02
No longer panic on bad config files

### LINKED ISSUE #792: Tremor crashes if a ramp can not be sucessfully started.
**Problem**
Tremor crashes if a ramp can not be started (i.e. non-existing Kafka topic is provided.)

**Steps**
1. create a Kafka source
2. add a topic that doesn't exist
3. :boom: 

**Possible Solution(s)**

Tremor should not panic if a ramp can not be created

**Notes**

Output of `tremor --version`: 0.10.1

Licenser:
# Pull request

## Description

This prevents crashing when bad config files are provided. Instead an error is printed and the server starts with the remaining config.

While this fixes the crash issue we should consider if we want to act this way.

An argument can be made that when a file is provided as a command-line argument the server should fail to start as the user should be shown this very explicitly to avoid confusion.

At the same time, an argument can be made that since tremor is a server and configuration can be loaded and unloaded at any point in its runtime, an invalid config is no reason to stop the service and prevent other pipelines/connectors to run.

Perhaps an alternative would be to exit 'nicely' when a config fails to load and just not panic?

## Related

* Related Issues: fixes #792

## Checklist

* [x] The code is tested

## Performance

This change only touches control plane code, there is no impact on the data plane and performance.

### DIFF
--- a/tremor-cli/src/server.rs
+++ b/tremor-cli/src/server.rs
@@ -121,19 +121,28 @@ pub(crate) async fn run_dun(matches: &ArgMatches) -> Result<()> {
     let (world, handle) = World::start(64, storage_directory).await?;
 
     if let Some(config_files) = matches.values_of("artefacts") {
+        let mut yaml_files = Vec::with_capacity(16);
         // We process trickle files first
-        for config_file in config_files.clone() {
-            match get_source_kind(config_file) {
-                SourceKind::Trickle => tremor_runtime::load_query_file(&world, config_file).await?,
-                _ => continue,
+        for config_file in config_files {
+            let kind = get_source_kind(config_file);
+            match kind {
+                SourceKind::Trickle => {
+                    if let Err(e) = tremor_runtime::load_query_file(&world, config_file).await {
+                        error!("Failed to load trickle script file {}: {}", config_file, e);
+                    }
+                }
+                SourceKind::Tremor | SourceKind::Json | SourceKind::Unsupported(_) => {
+                    error!("Unsupported file format {}: {}", kind, config_file);
+                }
+                SourceKind::Yaml => yaml_files.push(config_file),
             };
         }
+
         // We process config files thereafter
-        for config_file in config_files {
-            match get_source_kind(config_file) {
-                SourceKind::Trickle | SourceKind::Tremor => continue,
-                _ => tremor_runtime::load_cfg_file(&world, config_file).await?,
-            };
+        for config_file in yaml_files {
+            if let Err(e) = tremor_runtime::load_cfg_file(&world, config_file).await {
+                error!("Failed to load config file {}: {}", config_file, e)
+            }
         }
     }
 
--- a/tremor-cli/src/server.rs
+++ b/tremor-cli/src/server.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::errors::{Error, Result};
+use crate::errors::{Error, ErrorKind, Result};
 use crate::util::{get_source_kind, SourceKind};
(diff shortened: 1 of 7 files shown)


tremor-cli/src/server.rs
```diff
@@ -121,19 +121,28 @@ pub(crate) async fn run_dun(matches: &ArgMatches) -> Result<()> {
...
-        for config_file in config_files.clone() {
-            match get_source_kind(config_file) {
-                SourceKind::Trickle => tremor_runtime::load_query_file(&world, config_file).await?,
-                _ => continue,
+        for config_file in config_files {
+            let kind = get_source_kind(config_file);
+            match kind {
+                SourceKind::Trickle => {
+                    if let Err(e) = tremor_runtime::load_query_file(&world, config_file).await {
+                        error!("Failed to load trickle script file {}: {}", config_file, e);
```
mfelsche, line 131: We wanna `return Err(...)` here?

tremor-cli/src/server.rs
```diff
@@ -121,19 +121,28 @@ pub(crate) async fn run_dun(matches: &ArgMatches) -> Result<()> {
...
+                    }
+                }
+                SourceKind::Tremor | SourceKind::Json | SourceKind::Unsupported(_) => {
+                    error!("Unsupported file format {}: {}", kind, config_file);
```
mfelsche, line 135: same here?

tremor-cli/src/server.rs
```diff
@@ -121,19 +121,28 @@ pub(crate) async fn run_dun(matches: &ArgMatches) -> Result<()> {
...
+
         // We process config files thereafter
-        for config_file in config_files {
-            match get_source_kind(config_file) {
-                SourceKind::Trickle | SourceKind::Tremor => continue,
-                _ => tremor_runtime::load_cfg_file(&world, config_file).await?,
-            };
+        for config_file in yaml_files {
+            if let Err(e) = tremor_runtime::load_cfg_file(&world, config_file).await {
+                error!("Failed to load config file {}: {}", config_file, e)
```
mfelsche, line 144: same here

mfelsche (commented):
Looks cool. just 2 things:

* Do we wanna `return Err(...)` and not start the world if any file fails to be loaded correctly?
* Can you add an integration test for the case of an invalid config and assert on output and status?

tremor-cli/src/server.rs
Licenser: Specifically not, no, returning an error here is what leads to panicking the runtime.

Licenser:
Before we merge this we need to have a discussion on what behavior we do want, as mentioned in the PR itself both behaviors (exiting and continuing) are justifiable.

mfelsche:
I would argue for any file, be it yaml or trickle, which has been *explicitly* specified via the `-f`, if it fails to be loaded, we log the error, abort and return `1` and not startup tremor. If keeping it running nonetheless is the intended behaviour i would add an additional flag to state that intent explicitly. This way users wont spawn anything incomplete unless they explicitly sign up for that possibility.

What are your thoughts on this?

Licenser:
Ja that makes sense I replaced the stacktrace / panic with a more human friendly error:
```
tremor_1      | 2021-03-04T10:42:21.736104348+00:00 ERROR tremor::server - error: An error occurred while loading the file `/etc/tremor/config/config.yaml`: receiving from an empty and closed channel
tremor_1      | 2021-03-04T10:42:21.736133666+00:00 ERROR tremor::server - We are SHUTTING DOWN due to errors during initialization!
```

mfelsche (dismissed):

mfelsche (approved):

Issue #815, opened 2021-03-08
bench subcommand doesn't mark the benchmarks as failing when they actually fail


humancalico:
**Problem**

On running `cargo run -p tremor-cli -- test bench tremor-cli/tests/bench` Some benchmarks like `bench_pipe_passthrough` and `bench_pipe_passthrough_csv` don't run at all
```
Framework: Finding benchmark test scenarios
  Benchmark: Running bench_pipe_passthrough
    Tags:  bench pipeline passthrough


  Elapsed: 36ms 186us
  Benchmark: Running bench_pipe_passthrough_json
    Tags:  passthrough pipeline json bench

       | Value Percentile TotalCount 1/(1-Percentile)
       |
       | 125439 0.00000          1           1.00
       | 172031 0.25000     188485           1.33
       | 185343 0.50000     370716           2.00
       | 191487 0.62500     453031           2.67
       | 203775 0.75000     541814           4.00
       | 215039 0.81250     586813           5.33
       | 228351 0.87500     632275           8.00
       | 236543 0.90625     654639          10.67
       | 245759 0.93750     676229          16.00
       | 251903 0.95312     687439          21.33
       | 261119 0.96875     699140          32.00
       | 268287 0.97656     704978          42.67
       | 278527 0.98438     710618          64.00
       | 284671 0.98828     712992          85.33
       | 294911 0.99219     715826         128.00
       | 301055 0.99414     717002         170.67
       | 311295 0.99609     718452         256.00
       | 317439 0.99707     719085         341.33
       | 325631 0.99805     719687         512.00
       | 331775 0.99854     720030         682.67
       | 339967 0.99902     720364        1024.00
       | 346111 0.99927     720544        1365.33
       | 354303 0.99951     720691        2048.00
       | 360447 0.99963     720792        2730.67
       | 368639 0.99976     720864        4096.00
       | 376831 0.99982     720910        5461.33
       | 387071 0.99988     720945        8192.00
       | 395263 0.99991     720967       10922.67
       | 413695 0.99994     720989       16384.00
       | 421887 0.99995     721002       21845.33
       | 438271 0.99997     721011       32768.00
       | 446463 0.99998     721019       43690.67
       | 458751 0.99998     721022       65536.00
       | 462847 0.99999     721025       87381.33
       | 497663 0.99999     721028      131072.00
       | 501759 0.99999     721029      174762.67
       | 528383 1.00000     721031      262144.00
[...]

PR #816, opened 2021-03-08
fix(cli): mark the benchmarks as failing when they fail

humancalico:
Signed-off-by: Akshat Agarwal <humancalico@disroot.org>

# Pull request

## Description
Fix a bug in the `test bench` command that was giving false positives when the benchmarks were failing.
This still doesn't give the proper error trace when a benchmark run fails

## Related

* Related Issues: Ref #815

## Checklist

* [x] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 * Do not crash when we can not execute a config file [#792](https://github.com/tremor-rs/tremor-runtime/issues/792)
 * Sort the artefacts while running the benchmarks so that the benchmark run is more deterministic [#825](https://github.com/tremor-rs/tremor-runtime/issues/825)
 * Remove the bench_pipe_passthrough_csv benchmark [#825](https://github.com/tremor-rs/tremor-runtime/issues/825)
+* Fix a bug in the `test bench` command that was giving false negatives when the benchmarks were failing [#816](https://github.com/tremor-rs/tremor-runtime/pull/816)
 
 ## 0.10.2
 
--- a/tremor-cli/src/report.rs
+++ b/tremor-cli/src/report.rs
@@ -16,7 +16,7 @@ use crate::test::stats;
 use std::collections::HashMap;
 
 /// A test run is a collection of test reports that
-/// have executed int he context of a test run
+/// have executed in the context of a test run
 ///
 #[derive(Serialize, Debug, Clone)]
 pub(crate) struct TestRun {
--- a/tremor-cli/src/test.rs
+++ b/tremor-cli/src/test.rs
@@ -71,8 +71,12 @@ fn suite_bench(
                 status::tags(&tags, Some(&matched), Some(&by_tag.1))?;
                 let test_report = process::run_process("bench", base, root, &tags)?;
                 status::duration(test_report.duration, "  ")?;
+                if test_report.stats.is_pass() {
+                    stats.pass();
+                } else {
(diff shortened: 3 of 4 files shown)


Licenser (approved):
LGTM :+1: nice catch

could you rebase and add a line to the changelog then we can merge it

Licenser (dismissed):
:+1: perfect! Thank you :D

anupdhml (commented):
Nice job Akshat!

We now have the bench failures being reported in the overall stats but it would be nice to also mark the failed benches explicitly. Current reporting output:

```
(bench-fix) $ cargo run -p tremor-cli -- test bench tremor-cli/tests/bench --includes csv
    Finished dev [unoptimized + debuginfo] target(s) in 0.26s
     Running `target/debug/tremor test bench tremor-cli/tests/bench --includes csv`
Framework: Finding benchmark test scenarios
    Benchmark: Skipping bench_pipe_passthrough_influx
    Tags:  pipeline influx passthrough bench
    Benchmark: Skipping bench_pipe_passthrough_two_inputs
    Tags:  passthrough 2i bench pipeline
    Benchmark: Skipping bench_pipe_passthrough_binflux
    Tags:  bench pipeline binflux passthrough
    Benchmark: Skipping bench_pipe_passthrough_json
    Tags:  bench pipeline passthrough json
    Benchmark: Skipping bench_pipe_passthrough
    Tags:  pipeline passthrough bench
    Benchmark: Skipping bench_pipe_passthrough_msgpack
    Tags:  bench passthrough pipeline msgpack
    Benchmark: Skipping throughput-logging-msgpack
    Tags:  logging bench throughput msgpack
    Benchmark: Skipping throughput-logging-json
    Tags:  bench throughput logging json
  Benchmark: Running bench_pipe_passthrough_csv
    Tags:  bench pipeline passthrough csv


  Elapsed: 69ms 850us



All Benchmark Stats: Pass 0 Fail 1 Skip 8  Asserts 0
Total Stats: Pass 0 Fail 1 Skip 8  Asserts 0
Total Elapsed: 79ms 478us
1 out of 8 tests failed.
```

An easy way to achieve that right now is to adopt what integration tests do: define `assert.yaml` as part of the bench configs to catch process failures, which will be reported as assert failures right next to the test. Example content for the file:
```
status: 0
name: csv
asserts:
  - source: fg.out.log
    contains:
      - "Throughput:"
```

Since this is something we would need to replicate on every bench test, perhaps we want to report this the same way as cargo reports test/bench failures: for all the failing benches, report the errors at the end. This mechanism could be used then for other tests too and not just for bench.

I am fine with having the quick `assert.yaml` based solution for now and exploring nicer overall reporting as part of another issue later.

mfelsche:
... aaaand we need another rebase :)

humancalico:
Current status: The final report does show the number of benchmarks failing and the output of failing benchmarks is printed to stderr which is not the best way to things.
As @anupdhml like cargo our harness should log the failing tests/benchmarks and print their errors but should also give a detailed report in the end whose level of verbosity could be configured via some flags

anupdhml:
@humancalico let's track the nicer reporting improvements as a separate issue and merge this PR, since after the changes here, we can use the bench command output to identify the exact failing benchmark (error involved is also printed out which is nice).

anupdhml (dismissed):

anupdhml (approved):

PR #819, opened 2021-03-09
Add match groups and optimize some

### LINKED ISSUE #834: Tilde operator not working as match predicate

The tilde test expression is documented to work as match predicate:

```
match foo of
  case ~base64|| => "base64"
  default null
end
```

Unfortunately our grammar doesnt support that yet.

Make the grammar support tilde test expressions.

Licenser:
# Pull request

## Description

This  adds optimization to match statements, they especially target a large number of cases that are quite simple. It also lays the foundation for later optimizations in this space.

## Related

* Related Issues: fixes #834
* [Docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/91)

## Checklist

* [x] The code is tested

## Performance

After discussion, we believe that gaining 200% In complex cases is worth to lose 1 or 2 % of performance on shorter match cases.

### real-workflow-throughput-json
| main | this | delta |
|-|-|-|
|361.6 MB/s|357.3 MB/s| - 1% |

### real-workflow-throughput-json-v0.10
(same logic as `real-workflow-throughput-json` but written in more modern tremor)

| main | this | delta |
|-|-|-|
|368 MB/s|360 MB/s| - 2.2% |

### merge
| main | this | delta |
|-|-|-|
|54.0 MB/s|163.5 MB/s| + 202 % |


packaging/distribution/usr/lib/tremor/tremor.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh
```
mfelsche, line 1: why is this file here?

Licenser: I noticed it was missing on main :(

tremor-script/src/ast/analyzer.rs
```diff
@@ -0,0 +1,109 @@
...
+        let g = if self.guard.is_some() { 10 } else { 0 };
+        g + self.pattern.cost()
+    }
+}
+
+impl<'script> Costly for Pattern<'script> {
+    fn cost(&self) -> u64 {
+        match self {
+            Pattern::DoNotCare | Pattern::Default => 0,
+            Pattern::Expr(_) => 10,
```
mfelsche, line 41: could you extract those costs to some constants with meaningful names, so the reason for the cost estimation becomes clearer? That would help me and other noobs understand. :)

tremor-script/src/ast/analyzer.rs
```diff
@@ -0,0 +1,109 @@
...
+            ArrayPredicatePattern::Tilde(t) => t.cost(),
+            ArrayPredicatePattern::Record(r) => r.cost(),
+        }
+    }
+}
+
+impl<'script> Costly for ArrayPattern<'script> {
+    fn cost(&self) -> u64 {
+        let s: u64 = self.exprs.iter().map(Costly::cost).sum();
+        s * (self.exprs.len() as u64)
```
mfelsche, line 88: wouldnt the sum better represent the cost here? After all we iterate over the array pattern and try every match, no? Sounds like a linear thingy.

tremor-value/src/value.rs
```diff
@@ -82,6 +85,170 @@ pub enum Value<'value> {
...
+            (StaticNode::I64(v1), StaticNode::F64(v2)) => {
+                if v2 < i64::min_value() as f64 {
+                    Ordering::Greater
+                } else if v2 > i64::max_value() as f64 {
+                    Ordering::Less
+                } else {
+                    v1.cmp(&(v2 as i64))
+                }
+            }
+            // This is not great!
```
mfelsche, line 150: why though?

mfelsche (approved):
This is a magic kolossus! ✨ 

I mostly had questions.

tremor-value/src/value.rs
Licenser: while we don't expose it but float doesn't implement Ord since it refuses to compare `Nan==Nan` which makes sense so we kind of cheat around it

mfelsche: got it.

tests/scripts/match_reorder1/script.tremor
```diff
@@ -0,0 +1,62 @@
...
+  case %{k ~= glob|eight*|} => 8
+  default => event.k
+end of
+  case a = _ => a
+end;
+
+match a  of
+  case event => event
+  default => "snot"
+end
```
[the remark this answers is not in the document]
Licenser: :+1:

tremor-script/src/ast.rs
```diff
@@ -1363,10 +1751,40 @@ pub enum Pattern<'script> {
...
     }
+    fn is_exclusive_to(&self, other: &Self) -> bool {
+        match (self, other) {
+            // Two literals that are different are distinct
+            (Pattern::Expr(ImutExprInt::Literal(l1)), Pattern::Expr(ImutExprInt::Literal(l2))) => {
+                l1 != l2
+            }
+            // For record patterns we compare directly
+            (Pattern::Record(r1), Pattern::Record(r2)) => {
+                r1.is_exclusive_to(r2) || r2.is_exclusive_to(r1)
```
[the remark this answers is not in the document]
Licenser: So we can implement exclusivity in one direction instead of both, for example for `extractors` and `==` we only implement the `does an extractor is exclusive to a ==` not `is a == exclusive to an extractor`.

That said I'm not 100% sure where the 'right' place is to do this doubling :( it is in more places then I liked

tremor-script/src/ast.rs
```diff
@@ -1435,6 +1821,91 @@ pub enum PredicatePattern<'script> {
...
+                    lhs: lhs1,
+                    kind: BinOpKind::Eq,
+                    ..
+                },
+                PredicatePattern::Bin {
+                    lhs: lhs2,
+                    kind: BinOpKind::Eq,
+                    ..
+                },
+            ) if lhs1 == lhs2 => true,
```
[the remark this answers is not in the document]
Licenser: good catch yes! even so having the same RHS might be silly it is something we should not change the behaviour on!

tremor-script/src/ast.rs
```diff
@@ -1140,59 +1212,373 @@ pub struct TestExpr {
...
+        }
+    }
+
+    fn replace_last_shadow_use(&mut self, replace_idx: usize) {
+        match self {
+            Self::Simple { patterns, .. } => {
+                for p in patterns {
+                    let mut e = Ex::null_lit();
+                    mem::swap(&mut p.last_expr, &mut e);
+                    p.last_expr = e.replace_last_shadow_use(replace_idx);
```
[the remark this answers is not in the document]
Licenser: That's what I wanted to write, sadly rust says no :(

tremor-script/src/ast/analyzer.rs
Licenser: for each pattern in the array we itterate over the entire target array, so 3 patterns on a 10 element array iterate over 30 elements. We don't know how many elements will be there in the final array but a conservative approximation is that if you write `n` different patterns the array you're looking at will have at least `n` elements.

PS: This should be in a comment :D (will copy it in.

tremor-value/src/value.rs
Licenser: added this as a comment

tremor-script/src/ast/analyzer.rs
Licenser: not sure how much it'll help but I'll add some, at least it'll make adjusting them easier later :D :+1:

tremor-script/src/ast.rs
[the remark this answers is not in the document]
mfelsche: i think it is ok. We can revisit this when needed. I cannot come up with a case where the exclusivity relation is not commutative. So if anyhow, this might only be a problem of doing something twice, but thats ok as we are not in the hot path.

tremor-script/src/ast/analyzer.rs
mfelsche: Got it. I mixed it up with the tuple pattern logic. All good

mfelsche (approved):
Love it!

Licenser:
Question: are we OK with a 2-3% perf drop on the baseline test for the perf gain on more complex match statements?

mfelsche:
the baseline test is a single match arm? or no match at all? Is the performance drop really attributable to your changes? How is the behaviour with middle-sized match statements, say around 10 arms?

Licenser:
https://github.com/tremor-rs/tremor-runtime/blob/main/bench/real-workflow-throughput-json/main.trickle is the script in question, it is written in a fairly outdated style (we had no pattern matches back then) I'll make a `new` variant to see how that would fare

Licenser:
I'm not particularly happy with the decrease in performance for `real-workflow-throughput-json`

mfelsche (approved):
Finally!

PR #824, opened 2021-03-10
Add request info to response event metadata.

### LINKED ISSUE #753: [Rest Sink] Add request metadata to response event

Currently it is very hard to correlate request and response event, that is to know to which request a response belongs.

Insert the `$request` metadata into the response event, that is the endpoint, the method and the headers.
Maybe also include the body, but only if the sink is configured this way, otherwise the events might get too big.
We can also allow add room (underneath `$request.custom` maybe) for users to insert custom data to drag along into the response event.

aryan9600:
# Pull request

## Description

Include the request method, endpoint and headers inside a metadata variable `$request`, to help better correlate a request and response.

## Related

* Related Issues: fixes #753 
* Related docs PR: https://github.com/tremor-rs/tremor-www-docs/pull/87

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/rest.rs
+++ b/src/sink/rest.rs
@@ -261,6 +261,7 @@ enum CodecTaskInMsg {
         id: EventId,
         origin_uri: Box<EventOriginUri>, // box to avoid becoming the struct too big
         op_meta: Box<OpMeta>,
+        request_meta: Value<'static>,
         response: Response,
         duration: u64,
     },
@@ -351,6 +352,7 @@ impl Sink for Rest {
                                 segments.map(String::from).collect()
                             }),
                         };
+                        let request_meta = build_request_metadata(&request)?;
                         // send request
                         match http_client.send(request).await {
                             Ok(response) => {
@@ -362,6 +364,7 @@ impl Sink for Rest {
                                         id,
                                         origin_uri: Box::new(event_origin_uri),
                                         op_meta: Box::new(op_meta),
+                                        request_meta,
                                         response,
                                         duration,
                                     })
@@ -563,6 +566,7 @@ async fn codec_task(
                 id,
                 origin_uri,
                 op_meta,
+                request_meta,
                 mut response,
                 duration,
             } => {
@@ -595,6 +599,7 @@ async fn codec_task(
                         &id,
                         &mut response_ids,
                         origin_uri.as_ref(),
+                        request_meta,
                         response,
                         codec,
                         &mut codec_map,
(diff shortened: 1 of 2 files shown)


src/sink/rest.rs
```diff
@@ -837,30 +842,69 @@ async fn build_response_events(
...
+    endpoint.insert("path", none_if_empty(request.url().path()))?;
+    endpoint.insert("query", request.url().query().map(ToString::to_string))?;
+    endpoint.insert(
+        "fragment",
+        request.url().fragment().map(ToString::to_string),
+    )?;
+    request_meta.insert("endpoint", endpoint)?;
+
     meta.insert("response", response_meta)?;
+    meta.insert("request", request_meta)?;
```
mfelsche, line 907: can we put this underneath the `response` meta, so this lives at `$response.request`?

The reason is that the Rest offramp will interpret the `$request` metadata and creates a HTTP request from it, so depending on how people wire offramps off, it is possible that this request metadata might interfere with what the user has to configure for the Rest offramp. So we want to avoid this interference.

src/sink/rest.rs
```diff
@@ -837,29 +842,68 @@ async fn build_response_events(
...
                 .map(Value::from)
                 .collect();
-            headers.insert(name.to_string(), header_values)?;
+            response_headers.insert(name.to_string(), header_values)?;
         }
     }
-    response_meta.insert("headers", headers)?;
+
+    response_meta.insert("headers", response_headers)?;
+    request.content_type();
```
mfelsche, line 871: this can be removed, right?

src/sink/rest.rs
```diff
@@ -351,6 +352,7 @@ impl Sink for Rest {
                                 segments.map(String::from).collect()
                             }),
                         };
+                        let response_request = request.clone();
```
mfelsche, line 355: what about instead of cloning the whole request we build the request metadata here already into a Value and passing this into the `CodecTaskInMsg::ToEvent`?

mfelsche (commented):
Nice one! :)
I am just worried about the `.clone()` there. Maybe you can get rid of it? That would be awesome!

src/sink/rest.rs
aryan9600: The problem with building the request metadata into a `Value` and then passing it around would be that we would need define a `'static` lifetime in the struct since we are async here, that means when passing it to `CodecTaskInMsg::ToEvent` we would need to `clone_static()` on it, which I presume would be more expensive than caling `.clone()` on `request`. Am I correct here?

mfelsche: My idea here was to create a`Value<'static>` and to achieve this clone request headers, url etc to an owned value (like `String` and then use `Value::from(...)` on that, as all the containing values in that also need to be `'static`. It could be, I am mistaken here. @Licenser do you know if this could fly? Seems to be the most efficient way to extract the needed data from the request and to send it across an async boundary.

mfelsche (approved):
😻

PR #828, opened 2021-03-11
Restrict event references in windowed select queries and add 'emit_empty_windows` setting

### LINKED ISSUE #799: Degradation of handling signals in select statements
**Problem**

In tremor 0.10 `select` statements no longer handle signals gracefully. 

At a first glance, this is caused by emitting on ticks when the `select` references `event`

**Steps**

1. the 11_influx case in the example docs triggers this issue

**Possible Solution(s)**

Fix the handling of single to return to the 0.9 behavior

**Notes**

Output of `tremor --version`: `docker pull tremorproject/tremor:0.10`

### LINKED ISSUE #823: Clarify and codify access `event` in windows and groupings

Inside a window group we can't guarantee that an event is present when we emit, this prevents access to some of the `event` data at this time.

We should ensure that for windowed `selects` we limit access to parts of the event that are either in an aggregate function or part of the group by statement

mfelsche:
# Pull request

## Description

* Forbid references to `event` or `$` in the select target expression (`select --> event <-- from ...`) if this is a select query with a window, because we cannot access a single event in the select target, in that context we operate on the whole window. A freestanding event reference has always been evaluating to the last event of the window, but we have `aggr::win::last(...)` for that, and since `0.10` it is possible to have windows without an event (for time based windows), where query execution would error out.
  - We allow those references inside aggregate functions and if they are used in the group by clause.
  - Example:

```
select {
  "ok": event.foo, // allowed because used in group by
  "ok2": aggr::win::first(event.timestamp), // allowed because in aggregation function
  "bad": event.snot + 2 // not allowed anymore
} from in[my_window]
group by set(event.foo, event.bar)
into out;
```

* Add the setting `emit_empty_windows` for windows, which is `false` if not explicitly set to `true`. If set to `true`, time based tumbling windows will emit a window event on every `interval`  if they received an event or not. Some aggregations might return `null` in this case.

This fixes some unintended edge cases in the window handling changes introduced in `0.10`.

## Related

* Related Issues: fixes #823, closed #799
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/89)

## Checklist

* [x] The code is tested


tests/query_errors/new.sh
```diff
@@ -17,9 +17,10 @@ fi
 cp -r ${BASEDIR}/_template ${TARGET}
 git add ${TARGET}
 
-sed -e "s;//INSERT;//INSERT\n    ${NAME},;" ${BASEDIR}/../query_error.rs > ${BASEDIR}/tmp && mv ${BASEDIR}/tmp ${BASEDIR}/../query_error.rs
+sed -e "/^   \/\/INSERT/a\
```
Licenser, line 20: could we copy those over to all the new.sh's (or extract them into a function :D

tests/query_errors/window_event_in_target/error.txt
```diff
@@ -0,0 +1,3 @@
+Error: 
+    9 |   "min": math::min(event.number + 1, 0),
+      |                    ^^^^^^^^^^^^ References to `event` or `$` are not allowed in this context
```
Licenser, line 3: Can we add a `NOTE` on that explains this a bit more as to why and some explanation on how to get around this? (use grouping and/or aggregate functions)

tremor-cli/tests/integration/window_empty_windows/config.yaml
```diff
@@ -0,0 +1,21 @@
+onramp:
```
Licenser, line 1: since those are purely testing pipelines, can we move that to a `test`? They run significantly faster, are easier to run, are easier to add and maintain, and easier to understand than cli-integration tests. They also add to code coverage :P

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -137,7 +137,9 @@ impl std::default::Default for WindowImpl {
     fn default() -> Self {
         TumblingWindowOnTime {
             size: 15_000_000_000,
+            emit_empty_windows: false,
```
Licenser, line 140: I still think adding this is an extremely bad idea. We have been very careful not to give people the foot-gun of unbounded memory usage and had long discussions with users about why this is and quite a few tickets on `tremor has a memory leak` before we added this restriction to the grouping operator.

I think we shouldn't have it at all as it increases the sufrace and by that complexity for users and no one has ever asked for this. But if we have to have it for some reason this should be bounded the same way as the grouping operator is into something like `max_empty_windows`

tremor-script/src/ast/query.rs
```diff
@@ -298,6 +298,21 @@ pub(crate) enum GroupByInt<'script> {
     },
 }
 
+pub(crate) trait GroupByVisitor<'script> {
```
Licenser, line 301: since the visitors are pretty large and quite distinct from the rest of the at code,  can we pull them out into `ast/visitors.rs`?

tremor-script/src/registry.rs
```diff
@@ -15,7 +15,7 @@
 mod custom_fn;
 pub use self::custom_fn::CustomFn;
 pub(crate) use self::custom_fn::{RECUR_PTR, RECUR_REF};
-use crate::ast::{BaseExpr, NodeMetas};
+use crate::ast::{AstEq, BaseExpr, NodeMetas};
```
Licenser, line 18: same as with the visitor, can we pull AstEq in it's own module to shrink the at file a bit?

Licenser (changes requested):
Nice improvement :) I left a few inline comments

tremor-pipeline/src/op/trickle/select.rs
mfelsche: I am open to both. But i think this means @Licenser and @darach you need to go into the thunderdome and decide how we want this.

Licenser: Your opinion is as valuable as Darachs or mine :D, sorry you can't not have one.

mfelsche: I think it is ok to have this config when the default is to not emit empty windows.

Licenser: that makes 2 vs 1 :P leave it in then, but can we at least instead of making it unbounded bound it the same way we bound other parts growth parts?

mfelsche: i would very much like to avoid to add another config just because the first one is not good enough. The `max_empty_windows` is really too much. Then i would prefer to leave it out completely, honestly.

tremor-cli/tests/integration/window_empty_windows/config.yaml
mfelsche: Turns out we can't as the tests dont create any tick signals that we need here for those time based windows. :(

tests/query_errors/window_event_in_target/error.txt
mfelsche: good point

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -768,7 +823,7 @@ impl Operator for TrickleSelect {
                         aggregates,
                         &group_str,
                         &group_value,
-                    );
+                    )?;
```
Licenser, line 826: since this is the hot path, can we use `stry` here or are the error types different?

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -564,8 +613,14 @@ fn get_or_create_group<'window>(
             });
             e.insert(k, v)
         }
+        RawEntryMut::Vacant(_) => {
```
Licenser, line 616: we probably should add an `error!` here as well to make sure this is not a silent error even if the error pipeline isn't set up.

tremor-pipeline/src/query.rs
```diff
@@ -95,28 +95,41 @@ pub(crate) fn window_decl_to_impl<'script>(
...
-                    },
-                    |interval| {
-                        Ok(TumblingWindowOnTime::from_stmt(interval, ttl, script, stmt).into())
-                    },
+                .unwrap_or(WindowImpl::DEFAULT_MAX_GROUPS);
+            let emit_empty_windows = d
+                .params
+                .get("emit_empty_windows")
+                .and_then(Value::as_bool)
+                .unwrap_or(WindowImpl::DEFAULT_EMIT_EMPTY_WINDOWS);
```
Licenser, line 107: we should emit a warning here if `emit_empty_windows` is set and `max_groups` is not set

tremor-script/src/ast/support.rs
```diff
@@ -24,60 +21,13 @@ impl<'script> fmt::Debug for InvokeAggrFn<'script> {
     }
 }
 
+/// custom impl becauyse field `invocable` does not implement `PartialEq`
```
Licenser, line 24: could a blanket implementation of `impl<T: AstEq> for PartialEq T` work here?

Licenser:
I  love the ast walker :D that's going to be so handy over time!  I left a few comments inline most just questions otherwise :+1:

tests/query.rs
```diff
@@ -48,8 +48,8 @@ macro_rules! test_cases {
                 tremor_runtime::functions::load()?;
                 let query_dir = concat!("tests/queries/", stringify!($file), "/").to_string();
                 let query_file = concat!("tests/queries/", stringify!($file), "/query.trickle");
-                let in_file = concat!("tests/queries/", stringify!($file), "/in.xz");
-                let out_file = concat!("tests/queries/", stringify!($file), "/out.xz");
+                let in_file = concat!("tests/queries/", stringify!($file), "/in");
```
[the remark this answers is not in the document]
mfelsche: this is just for coverage of the `load_events_file` function, so this case is actually used. It will always support both the `.xz` name and the one without and always try both.

tremor-pipeline/src/op/trickle/select.rs
mfelsche: i am gonna stry all the things here!

tremor-pipeline/src/op/trickle/select.rs
mfelsche: this is erroring out of the whole event which will log an error as tested here: https://github.com/tremor-rs/tremor-runtime/pull/828/files#diff-d1682b33030c689487d46d41c5f162f105858e9b3e5611918438a35295adc045R9

is this good enough?

tremor-script/src/ast/support.rs
mfelsche: you mean that `PartialEq` used the `AstEq` logic for everything that also implements `AstEq`?

If so, I would say no, as `AstEq` explicitly ignores the node metadata (`mid`) and thus e.g. the location in the source, while `PartialEq` should take the `mid` into account.

tremor-pipeline/src/op/trickle/select.rs
Licenser: Ja! I didn't look up the chain :) sorry

tremor-script/src/ast/support.rs
Licenser: fair argument!

Licenser (approved):
Woooh! Nice job :D

PR #851, opened 2021-03-18
Emit warning when emit_empty_windows is used unguarded.

mfelsche:
And also rework the warnings api for the parser a bit.

This was an oversight from #828 

# Pull request

## Description

We now create a warning like this when only `emit_empty_windows` is specified without guards like `max_groups` or `eviction_period`:

```
Warning: 
    1 | define tumbling window my_window
    2 | with
    3 |   interval = core::datetime::with_seconds(2),
    4 |   emit_empty_windows = true
    5 | end;
      | ^^^ Using `emit_empty_windows` without guard is potentially dangerous. Consider limiting the amount of groups maintained internally by using `max_groups` and/or `eviction_period`
```

## Checklist

* [x] The code is tested


darach (dismissed):
LGTM

Licenser (approved):
:+1: Nice! I love the const'ed strings!

Licenser (approved):

PR #857, opened 2021-03-19
Fix kafka consumer offset lag of at least 1 by using offset + 1 when committing.

mfelsche:
# Pull request

## Description

We have been doing it wrong all the time. We manually committed offsets when `enable.auto.commit: false` has been configured the wrong way by taking the kafka record offset and using it in the commit call.
I found out one is expected to add 1 to the offset to denote "I consumed this message, don't give it to me again.".

> Note: The committed offset should always be the offset of the next message that your application will read. Thus, when calling commitSync(offsets) you should add one to the offset of the last message processed. 

https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html

There is currently no cool way to test this. We need to setup some more involved end to end tests for this.
But wanted to get this out quickly.

## Checklist

* [ ] The code is tested


darach (approved):
LGTM 🚀  Nice catch!

Licenser (approved):
Darn, that's a hard one to catch! Well done :+1:

Licenser (approved):

PR #866, opened 2021-03-23
Add offramp and onramp for NATS

### LINKED ISSUE #723: Add NATS.io Source and Sink
Add a source that can connect to and consume messages from a [nats-server](https://docs.nats.io/nats-server/installation) and [nats-streaming-server](https://docs.nats.io/nats-streaming-concepts/intro).

Add a sink that can connect to and send messages to a [nats-server](https://docs.nats.io/nats-server/installation) and [nats-streaming-server](https://docs.nats.io/nats-streaming-concepts/intro).

aryan9600:
# Pull request

## Description

Add an onramp and offramp for [NATS](https://nats.io) using the [`nats.rs`](https://github.com/nats-io/nats.rs) crate.

## Related

* Related Issues: closes #723 

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/nats.rs
+++ b/src/sink/nats.rs
@@ -1,0 +1,316 @@
+// Copyright 2020-2021, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![cfg(not(tarpaulin_include))]
+
+use std::iter::FromIterator;
+use std::time::Instant;
+
+use crate::sink::prelude::*;
+use async_channel::{bounded, Receiver};
+use async_nats::Connection as NatsConnection;
+use async_nats::Headers;
+use async_nats::Options as NatsOptions;
+use halfbrown::HashMap;
+use tremor_pipeline::OpMeta;
+
+// struct containing connection options
+#[derive(Clone, Debug, Deserialize, Default, PartialEq)]
+pub struct ConnectOptions {
+    token: Option<String>,
+    username: Option<String>,
+    password: Option<String>,
+    credentials_path: Option<String>,
+    cert_path: Option<String>,
+    key_path: Option<String>,
+    name: Option<String>,
+    #[serde(default = "bool::default")]
+    echo: bool,
+    max_reconnects: Option<usize>,
+    reconnect_buffer_size: Option<usize>,
+    #[serde(default = "bool::default")]
+    tls: bool,
+    root_cert: Option<String>,
+}
+
+impl ConnectOptions {
+    pub fn generate(&self) -> Result<NatsOptions> {
+        let mut options = None;
+        if let Some(token) = &self.token {
+            options = Some(NatsOptions::with_token(token.as_str()));
+        }
+        if let Some((username, password)) = self.username.as_ref().zip(self.password.as_ref()) {
+            options = Some(NatsOptions::with_user_pass(
+                username.as_str(),
+                password.as_str(),
+            ));
+        }
+        if let Some(credentials_path) = &self.credentials_path {
+            options = Some(NatsOptions::with_credentials(credentials_path));
+        }
+        let mut nats_options = options.unwrap_or_default();
+        if let Some((cert_path, key_path)) = self.cert_path.as_ref().zip(self.key_path.as_ref()) {
+            nats_options = nats_options.client_cert(cert_path, key_path);
+        }
+        if let Some(name) = &self.name {
+            nats_options = nats_options.with_name(name.as_str());
+        }
+        if let Some(max_reconnects) = self.max_reconnects {
+            nats_options = nats_options.max_reconnects(max_reconnects);
+        }
+        if let Some(reconnect_buffer_size) = self.reconnect_buffer_size {
+            nats_options = nats_options.reconnect_buffer_size(reconnect_buffer_size);
+        }
+        if let Some(root_cert) = &self.root_cert {
+            nats_options = nats_options.add_root_certificate(root_cert);
+        }
+        if self.echo {
+            nats_options = nats_options.no_echo();
+        }
+        nats_options = nats_options.tls_required(self.tls);
+        Ok(nats_options)
+    }
+}
+
+#[derive(Deserialize)]
+pub struct Config {
+    // list of hosts
+    pub hosts: Vec<String>,
+    // subject to send messages to
+    pub subject: String,
+    // options to use when opening a new connection
+    #[serde(default = "Default::default")]
+    pub options: ConnectOptions,
+    // reply to use for the messages
+    #[serde(default = "Default::default")]
+    pub reply: Option<String>,
+    // headers to use for the messages
+    #[serde(default = "Default::default")]
+    pub headers: HashMap<String, Vec<String>>,
+}
+
+impl Config {
+    fn connection(&self) -> Result<NatsConnection> {
+        let hosts = self.hosts.join(",");
+        task::block_on(async {
+            let connection = self.options.generate()?.connect(&hosts).await?;
+            Ok(connection)
+        })
+    }
+}
+
+impl ConfigImpl for Config {}
+
+pub struct Nats {
+    sink_url: TremorUrl,
+    config: Config,
+    postprocessors: Postprocessors,
+    reply_channel: Sender<sink::Reply>,
+    connection: Option<NatsConnection>,
+    error_rx: Receiver<()>,
+    error_tx: Sender<()>,
+    merged_meta: OpMeta,
+}
+
+impl offramp::Impl for Nats {
+    fn from_config(config: &Option<OpConfig>) -> Result<Box<dyn Offramp>> {
+        if let Some(config) = config {
+            let config: Config = Config::new(config)?;
+            let (dummy_tx, _) = bounded(1);
+            let (error_tx, error_rx) = bounded(crate::QSIZE);
+            Ok(SinkManager::new_box(Self {
+                sink_url: TremorUrl::from_offramp_id("nats")?,
+                config,
+                postprocessors: vec![],
+                reply_channel: dummy_tx,
+                connection: None,
+                error_rx,
+                error_tx,
+                merged_meta: OpMeta::default(),
+            }))
+        } else {
+            Err("Nats offramp requires a configuration.".into())
+        }
+    }
+}
+
+impl Nats {
+    fn handle_connection(&mut self) -> Result<Option<&NatsConnection>> {
+        while let Ok(()) = self.error_rx.try_recv() {
+            self.connection = None;
+        }
+        if self.connection.is_none() {
+            self.connection = Some(self.config.connection()?);
+            return Ok(self.connection.as_ref());
+        }
+        Ok(None)
+    }
+}
+
+#[async_trait::async_trait]
+impl Sink for Nats {
+    #[allow(clippy::cast_possible_truncation)]
+    async fn on_event(
+        &mut self,
+        _input: &str,
+        codec: &mut dyn Codec,
+        _codec_map: &HashMap<String, Box<dyn Codec>>,
+        mut event: Event,
+    ) -> ResultVec {
+        self.handle_connection()?;
+        let ingest_ns = event.ingest_ns;
+        let processing_start = Instant::now();
+        // evaluate here to avoid borrowing again while borrowed.
+        let config_reply = self.config.reply.as_deref();
+        let op_meta = &event.op_meta;
+        self.merged_meta.merge(op_meta.clone());
+        let insight_event = event.insight_ack();
+        if let Some(connection) = &mut self.connection {
+            for (value, meta) in event.value_meta_iter() {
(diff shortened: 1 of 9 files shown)


src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                    let mut message_reply = None;
+                    if let Some(reply) = reply {
+                        if let Some(reply_str) = reply.as_str() {
+                            message_reply = Some(reply_str);
+                        }
+                    } else if let Some(config_reply) = &self.config.reply {
+                        message_reply = Some(config_reply.as_str());
+                    }
+                    // prepare message headers
+                    let mut key_val: Vec<(&str, &str)> = Vec::new();
```
Licenser, line 177: We probably want to use `with_capacity` here to preallocate the size

src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                    if let Some(headers) = headers {
+                        if let Some(headers_obj) = headers.as_object() {
```
Licenser, line 179: you could unnest this a bit by calling `get_object` in Line 163

src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                            }
+                        }
+                    } else {
+                        for (key, val) in &self.config.headers {
+                            for ele in val.iter() {
+                                key_val.push((key.as_str(), ele.as_str()));
+                            }
+                        }
+                    }
+                    let key_val_len = key_val.len();
```
Licenser, line 197: usually `key_val.is_empty()` is the more ideomatic aproach

src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                                    for ele in array.iter() {
+                                        if let Some(ele) = ele.as_str() {
+                                            key_val.push((key, ele));
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    } else {
+                        for (key, val) in &self.config.headers {
```
Licenser, line 191: Should we always set those headers and then only overwrite them on the metadata?

src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                let mut reply = None;
+                if let Some(meta_data) = meta_data {
+                    headers = meta_data.get("headers");
+                    reply = meta_data.get("reply");
+                }
+                for payload in processed {
+                    // prepare message reply
+                    let mut message_reply = None;
+                    if let Some(reply) = reply {
+                        if let Some(reply_str) = reply.as_str() {
```
Licenser, line 170: you can unnest this as `get_str` in L163

src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                            }
+                        }
+                    }
+                    let key_val_len = key_val.len();
+                    let nats_headers = Headers::from_iter(key_val);
+                    let message_headers = if key_val_len > 0 {
+                        Some(&nats_headers)
+                    } else {
+                        None
+                    };
```
Licenser, line 203: You could possibly sijmplify this as:

```suggestion
                    let message_headers = if key_val.is_empty() {
                        None
                    } else {
                        Some(Headers::from_iter(key_val))
                    };
```

and then call `message_headers.as_ref()` in the call to publish

src/sink/nats.rs
aryan9600: Right now, we use the headers from the metdata if they _exist_, otherwise we use the headers specified in the config. Isn't that what we should be doing?

Licenser: There are arguments for both, but the big one IMHO is that for rest, [where we also allow setting headers](https://github.com/tremor-rs/tremor-runtime/blob/main/src/sink/rest.rs#L813) in the config and via metadata, we do use the config headers + the metadata headers. So being consistent will help users to avoid unexpected behavior.

If we did an `either / or` in rest I'd say to stick to that but in the existing code we do a `and` so I think we should stick to this behavior.

src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                        }
+                    }
+                }
+            }
+        }
+        Ok(None)
+    }
+
+    async fn on_signal(&mut self, signal: Event) -> ResultVec {
+        if self.connection.is_none() {
```
mfelsche, line 239: How could we get here?
After what I see, we do fail `init` if we can't connect, so we will only get active if we established a connection, and there is no place where we set the connection to `None`.

Should a background task call `self.connection.flush()` in a regular manner and set the connection to `None` if this fails? Should we try and reconnect when a publish request fails?

Anyways if you send those restores/triggers, you need to put an op_meta into the insight event. Please see the Ws sink and how it handles it with the `merged_meta` field.

src/source/nats.rs
```diff
@@ -0,0 +1,178 @@
...
+        } else {
+            nc.subscribe(self.config.subject.as_str()).await?
+        };
+        self.connection = Some(nc);
+        self.subscription = Some(sub);
+        Ok(SourceState::Connected)
+    }
+
+    async fn terminate(&mut self) {
+        if let Some(connection) = &self.connection {
```
mfelsche, line 172: should we drain the subscription here?

src/source/nats.rs
```diff
@@ -0,0 +1,178 @@
...
+                    data,
+                    meta: Some(nats_meta_data),
+                    codec_override: None,
+                    stream: 0
+                })
+            } else {
+                Ok(SourceReply::Empty(0))
+            }
+        } else {
+           Ok(SourceReply::StateChange(SourceState::Disconnected))
```
mfelsche, line 151: when are we getting here and how can we recover from this?

mfelsche (commented):
Hi @aryan9600 thanks for working on this, much appreciated!

I have some remarks with regards to connection and insight handling though.

src/sink/nats.rs
```diff
@@ -0,0 +1,298 @@
...
+                        .publish_with_reply_or_headers(
+                            self.config.subject.as_str(),
+                            message_reply,
+                            message_headers,
+                            payload,
+                        )
+                        .await;
+                    match publish_result {
+                        Ok(()) => {
+                            self.reply_channel
```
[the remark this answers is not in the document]
aryan9600: I see. So just logging the error should be enough here?

src/sink/nats.rs
aryan9600: Looking at the current implementation, it looks like `on_signal()` will be called every 100 ms, so instead of checking `connection` is `None`, we could  call `connection.flush()`, try to restore the connection and set it to `None` if connecting fails? Regarding, reconnecting if a publish request fails, since we are going to be running this very frequently, is that really required? Please let me know :D
I'll take a look at the websocket sink, to get a better idea on how to handle `op_meta`.

src/source/nats.rs
aryan9600: I was in two minds about this, but decided to not do it, after going through the implementation of the [`close()`](https://github.com/nats-io/nats.rs/blob/master/src/client.rs#L222-L255) of the `nats.rs` crate, which does drains all subscriptions.

src/sink/nats.rs
mfelsche: Calling flush here is not a good idea, as this will `await` until it was able to do the `ping - pong` roundtrip, i think this can take quite long and in this time the sink cannot progress otherwise, which is not what we want. Also, i don't know if it is possible to share a connection across tasks, so a background task might not be feasible.

aryan9600: I think we can do this in a background task. If you take a look at the `nats.rs` source code, a [`Connection` seems to be an `Arc` wrapping another struct](https://github.com/nats-io/nats.rs/blob/master/src/lib.rs#L307), which should be safe to pass across tasks, if I am not mistaken.

mfelsche: yes, in this case this should be safe. I only hope this won't block publishing events in some way.

src/source/nats.rs
aryan9600: Okay, so I think I messed up here. `sub.next()` is a blocking call, which returns `None` if the subscription isn't valid somehow. So we should be returning `Ok(SourceReply::StateChange(SourceState::Disconnected))` instead of `Ok(SourceReply::Empty(0))`. Returning `Ok(SourceReply::Empty(0))` doesn't actually make sense because `sub.next()` is a blocking call (the `async_nats` crate doesn't provide a non blocking method).
I don't think we need to do anything in this `else` block, since `self.subscription` is never being set to `None` anywhere. Regarding recovery, I can't seem to think how to do that since we seem to be just returning an `Ok`. Please do let me know if I'm wrong about this.

src/sink/nats.rs
aryan9600: I was playing around with this, and I think it makes more sense to recover a dead connection when we fail to publish a message rather than call `connection.flush()` in a background task in `on_signal()` (similar to how we handle this in kafka). Do let me know your thoughts about this.

mfelsche: The simpler the better! :+1: And it will work greatly if you can somehow distinguish between connection errors and non-connection errors.

darach (dismissed):
LGTM @aryan9600 👍

src/sink/nats.rs
```diff
@@ -0,0 +1,324 @@
...
+            options = Some(NatsOptions::with_token(token.as_str()));
+        }
+        if let Some(username) = &self.username {
+            if let Some(password) = &self.password {
+                options = Some(NatsOptions::with_user_pass(
+                    username.as_str(),
+                    password.as_str(),
+                ));
+            }
+        }
```
Licenser, line 60: This can likely be de-nested using Option::zip (I learned that today :D)
```suggestion
        if let Some((username, password)) = &self.username.zip(&self.password) {
            options = Some(NatsOptions::with_user_pass(
                &username,
                &password,
            ));
        }
```
(some for the `cert_client` version)

src/sink/nats.rs
```diff
@@ -0,0 +1,324 @@
...
+                let headers = meta.get("nats").and_then(|v| v.get_object("headers"));
+                let reply = meta.get("nats").and_then(|v| v.get_str("reply"));
+                for payload in processed {
+                    // prepare message reply
+                    let mut message_reply = None;
+                    if let Some(reply) = reply {
+                        message_reply = Some(reply);
+                    } else if let Some(config_reply) = &self.config.reply {
+                        message_reply = Some(config_reply.as_str());
+                    }
```
Licenser, line 195: The more ideomatic way to write this is:
```suggestion
                    let message_reply = if let Some(reply) = reply {
                        message_reply = Some(reply);
                    } else if let Some(config_reply) = &self.config.reply {
                        message_reply = Some(config_reply.as_str());
                    } else {
                        None
                    };
```

src/sink/nats.rs
```diff
@@ -0,0 +1,324 @@
...
+                    // prepare message headers
+                    let mut key_val: Vec<(&str, &str)> =
+                        Vec::with_capacity(self.config.headers.len());
```
Licenser, line 198: It would be possible including the length of custom headers (+/- the map function) :
```suggestion
                    let mut key_val: Vec<(&str, &str)> =
                        Vec::with_capacity(self.config.headers.len() + headers.map(HashMap::len).unwrap_or_default());
```

src/sink/nats.rs
```diff
@@ -0,0 +1,324 @@
...
+                        for (key, val) in headers.iter() {
+                            if let Some(array) = val.as_array() {
+                                for ele in array.iter() {
+                                    if let Some(ele) = ele.as_str() {
+                                        key_val.push((key, ele));
+                                    }
+                                }
+                            }
+                        }
+                    }
```
Licenser, line 214: This could possibly be de-nested and simplifed using filter_map:
```suggestion
                    if let Some(headers) = headers {
                        for (key, val) in headers.iter().filter_map(|(k,v)| Some((k, v.as_array()?)) {
                            for ele in array.iter() {
                                if let Some(ele) = ele.as_str() {
                                    key_val.push((key, ele));
                                }
                            }
                        }
                    }
```

src/source/nats.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[serde(default = "Default::default")]
+    pub options: ConnectOptions,
+}
+
+impl ConfigImpl for Config {}
+
+impl Config {
+    async fn connection(&self) -> Result<NatsConnection> {
+        let hosts = self.hosts.join(",");
+        let connection = self.options.generate()?.connect(hosts.as_str()).await?;
```
Licenser, line 39: Where a `&str` is required and you have a `String` most of the time just using `&` works fine and is the more rusty way :D 
```suggestion
        let connection = self.options.generate()?.connect(&hosts).await?;
```

src/source/nats.rs
```diff
@@ -0,0 +1,178 @@
...
+        if let Some(sub) = &self.subscription {
+            if let Some(msg) = sub.next().await {
+                let mut origin_uri = self.origin_uri.clone();
+                origin_uri.path = vec![msg.subject];
+                let data = msg.data;
+                let mut nats_meta_data = Value::object_with_capacity(1);
+                let mut msg_reply = None;
+                if let Some(reply) = msg.reply {
+                    msg_reply = Some(reply);
+                }
```
Licenser, line 121: The more rusty way is using an expression instead of mutating, it generally makes the code easier to reason about:
```rust
                let msg_reply = if let Some(reply) = msg.reply {
                    Some(reply)
                } else {
                    None
                }
```

it also maps it easier to spot that this basically is:
```suggestion
                let msg_reply = msg.reply;
```

Licenser (approved):
Nice :D thank you so much! The last thing to do before this is merged is adding [the docs](https://github.com/tremor-rs/tremor-www-docs) for the nats sink / source :)

Licenser (dismissed):
Woooh!

Licenser (approved):

PR #870, opened 2021-03-24
Fix sources not setting `Event::transactional`

### LINKED ISSUE #868: Sources do not set field Event::transactional
**Problem**

We discovered the kafka onramp not committing offsets although the connected offramp (UDP in that case) received and handled the messages successfully. The kafka onramp was configured with `enable.auto.commit: false`, so it returned `true` from `Source::is_transactional`. The intended logic here is that each event is marked as `transactional` so any downstream operator or sink will know to send a CB `Ack` or `Fail` insight back, as the event origin needs that (in our case the kafka consumer needed it to commit the offset).

It turns out we never set that flag.

Also we found out that we for most offramps send `Ack` or `Fail` regardless of the `transactional` flag of the current event. We can spare some needless contraflow events here when we dont need them.

**Steps**

1. Configure kafka offramp with `enable.auto.commit: false`.
2. Connect via passthrough pipeline to stdout or udp offramp.
3. send messages to kafka, watch them appear at stdout or the udp downstream.
4. Watch the kafka consumer offset to stay the same.

**Possible Solution(s)**

* Set the `transactional` flag based upon the `Source::is_transactional()` method.
* Only send CB `Ack`/`Fail` events if the current event is `transactional`.

mfelsche:
# Pull request

## Description

* Fix sources not setting `Event::transactional`
* Ensure sinks only emit ack/fail insights when the event is transactional, so we spare some processing if it is not needed.
* Fix kafka onramp `origin_url` not containing the partition id, although the docs state it should.

## Related

* Related Issues: fixes #868 
* docs PR: https://github.com/tremor-rs/tremor-www-docs/pull/94

## Checklist

* [x] The code is tested

## Performance

roughly equivalent.


Licenser:
grafik
I love it!!! 🚀 👍

darach:
I love it too. It'd be nicer if the coverage went up though 🤓

Licenser:
All jokes aside, I don't see any code that would explain a 20% perf improvement, any idea how it improved?

darach:
20mb/s in those numbers so 7.8%?

Licenser:
> 20mb/s in those numbers so 7.8%?

Ja too late for math, but I don't see then 7.8%  either so I'm curious :D

mfelsche:
Seems to be local fluctuation due to whatever background noise through other processes on my machine. Also got a run on my branch with 255MB/s. Gonna do some more runs.

mfelsche:
Jep, all noise, consider the throughput roughly the same for this benchmark.

Licenser (approved):
:+1:

PR #872, opened 2021-03-25
Add Update and Delete to ES sink

### LINKED ISSUE #822: Enhance ES sink support

We should support more capabilities of the ES API:

* [ ] Delete
* [ ] Update

TBD

Licenser:
# Pull request

## Description

Add additional operations to elastic search

## Related

* Related Issues: closed #822
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/96)

## Checklist

* [x] The code is tested

## Performance

This PR has no impact on performance outside of the ES source, but the change to rust 1.51.0 is worth tracking:

## real-workflow-throughput-json-v0.10

| main | add-es-option |
|-|-|
| 348 MB/s | 353 MB/s |

marginal improvement but close to noise


src/sink/elastic.rs
```diff
@@ -241,13 +241,36 @@ fn build_event_payload(event: &Event) -> Result<Vec<u8>> {
         if let Some(pipeline) = meta.get_str("pipeline") {
             index_meta.insert("pipeline", pipeline)?;
         };
-        let value_meta = json!({ "index": index_meta });
+        let key = match meta.get_str("es_op") {
```
mfelsche, line 244: i would call it `action` as this is the elasticsearch terminology here: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update

We might want to namespace those metadata values under `elastic` as we do with other sinks, to avoid potential collisions with other `action` metadata values.

Licenser: Good call! I hated the name :D `asction` is much better!

src/sink/elastic.rs
```diff
@@ -202,6 +209,7 @@ fn build_bulk_success_data(
     let mut meta = Object::with_capacity(1);
     let mut es_meta = Object::with_capacity(4);
     es_meta.insert("id".into(), Value::from(item.id().to_string()));
+    es_meta.insert("id".into(), Value::from(item.id().to_string()));
```
mfelsche, line 212: I think here you also need to add `_id` and all.

Licenser: darn I overlooked that

mfelsche (approved):

PR #878, opened 2021-03-27
refactor: improved file path handling

humancalico:
# Pull request

## Related

## Checklist

* [ ] The code is tested


tremor-cli/src/run.rs
```diff
@@ -229,7 +230,7 @@ impl Egress {
 }
 
 fn run_tremor_source(matches: &ArgMatches, src: String) -> Result<()> {
-    let raw = slurp_string(&src);
+    let raw = fs::read_to_string(&src);
```
Licenser, line 233: I would prefer keeping slurp_string as it gives better errors than `fs::read_to_string`. (i.e. tells a user what file couldn't be opend)

tremor-cli/src/test/after.rs
```diff
@@ -36,38 +37,38 @@ impl After {
...
         // This is optional
-        if Path::new(after_str).is_file() {
-            let after_json = load_after(after_str)?;
+        if Path::new(&after_path).is_file() {
+            let after_json = load_after(&after_path)?;
             let after_process = after_json.spawn(root)?;
             if let Some(mut process) = after_process {
-                let after_out_file = format!("{}/after.out.log", root);
-                let after_err_file = format!("{}/after.err.log", root);
+                let after_out_file = root.join(Path::new("after.out.log"));
```
Licenser, line 70: Is `Path::new` needed here? in line 64 it seems to work without.

tremor-cli/src/test/assert.rs
```diff
@@ -107,11 +107,11 @@ pub(crate) struct FileBasedAssert {
...
 pub(crate) type Asserts = Vec<FileBasedAssert>;
 
-pub(crate) fn load_assert(path_str: &str) -> Result<AssertSpec> {
-    let data = slurp_string(path_str)?;
+pub(crate) fn load_assert(path: &Path) -> Result<AssertSpec> {
+    let data = fs::read_to_string(path)?;
     match serde_yaml::from_str::<AssertSpec>(&data) {
         Ok(mut s) => {
-            let base = Path::new(path_str);
+            let base = Path::new(path);
```
Licenser, line 114: would `.clone()` be more ideomatic here?

tremor-cli/src/test/process.rs
```diff
@@ -97,9 +100,9 @@ pub(crate) fn run_process(
     after.spawn()?;
     after::update_evidence(&bench_root, &mut evidence)?;
     // Assertions
-    let assert_str = &format!("{}/assert.yaml", bench_root);
-    let report = if Path::new(assert_str).is_file() {
-        let assert_yaml = assert::load_assert(assert_str)?;
+    let assert_path = bench_root.join(Path::new("assert.yaml"));
+    let report = if Path::new(&assert_path).is_file() {
```
Licenser, line 104: since `assert_path` is a path already do we need `Path::new` here?

tremor-cli/src/run.rs
Licenser: or perhaps even better a `tremor-common` like a variant of read_to_string

https://github.com/tremor-rs/tremor-runtime/blob/main/tremor-common/src/file.rs

tremor-cli/src/test/command.rs
```diff
@@ -136,8 +135,8 @@ pub(crate) fn suite_command(
                             job::TargetProcess::new_with_stderr(&resolved_cmd, &args, &case.env)?;
                         let exit_status = fg_process.wait_with_output();
 
-                        let fg_out_file = format!("{}/fg.{}.out.log", base_str, counter);
-                        let fg_err_file = format!("{}/fg.{}.err.log", base_str, counter);
+                        let fg_out_file = suite_root.join(&format!("fg.{}.out.log", counter));
+                        let fg_err_file = suite_root.join(&format!("fg.{}.out.log", counter));
```
Licenser, line 139: this should probably be `fg.{}.out.log`

tremor-cli/src/test/tag.rs
```diff
@@ -135,16 +134,12 @@ impl TagFilter {
...
         let mut tags = TagFilter::new(vec![], vec![]);
-        tags = tags.join(Some(maybe_slurp_tags(&tags_file)));
-        for dirname in rel.split('/') {
-            base = format!("{}/{}", base, dirname);
-            let tags_file = format!("{}/tags.json", &base);
-            tags = tags.join(Some(maybe_slurp_tags(&tags_file)));
-        }
+        tags = tags.join(Some(maybe_slurp_tags(&base_tags_file)));
+        let other_tags_file = other.join("tags.json");
+        tags = tags.join(Some(maybe_slurp_tags(&other_tags_file)));
```
Licenser, line 142: I think that changes the behaviour. The intention is to find all tag files in the directories between `base` and `other` this code no longer does that from what I can tell.

basically if base is `/a/b/c` and other is `/a/b/c/d/e/f` we want to look for:

- `/a/b/c/tags.json`
- `/a/b/c/d/tags.json`
- `/a/b/c/d/e/tags.json`
- `/a/b/c/d/e/f/tags.json`

Licenser (changes requested):
Nice! Looks really good :+1: just two little things poped up (see comments inline).

tremor-cli/src/test/command.rs
humancalico: Whoops :facepalm:

tremor-cli/src/test/tag.rs
humancalico: I've done some changes but I still don't think that's the right way to do it. We should probably use the globwalk crate to do this. wdyt?

Licenser: Globwalk sounds like a really good idea 👍

tremor-cli/src/test/tag.rs
```diff
@@ -125,31 +124,42 @@ impl TagFilter {
...
-        let tags_file = format!("{}/tags.json", &base);
+    if other.strip_prefix(base).is_ok() {
+        let tags_file = base.join("tags.json");
         let mut tags = TagFilter::new(vec![], vec![]);
         tags = tags.join(Some(maybe_slurp_tags(&tags_file)));
-        for dirname in rel.split('/') {
-            base = format!("{}/{}", base, dirname);
-            let tags_file = format!("{}/tags.json", &base);
+        let depth = other.components().count() - base.components().count();
+        let walker = globwalk::glob_builder("tags.json")
```
Licenser, line 148: I don't thinkt that is going to work the way we want, could we add a test to make sure? I think this would find all tags.json relative to the base name of the executable.

It also would probably find something like /a/b/c/x/tag.json

humancalico: Forgot to add the base :facepalm:

Licenser (changes requested):
I think the glob walker would still act incorrectly. Can we add a test for this?

if the base is `/a/b/c` and other is `/a/b/c/d/e/f` we want to look for:
```
    /a/b/c/tags.json
    /a/b/c/d/tags.json
    /a/b/c/d/e/tags.json
    /a/b/c/d/e/f/tags.json
```

but not for

```
    /a/b/c/x/tags.json
    /a/b/c/y/e/tags.json
    ...
```

humancalico:
> Can we add a test for this?

I am not sure how. I need some help with this

Licenser:
Absolutely,

I'd go about it like this:

Create a sub-folder (for example `tests/fixtures/resolve_tags` ) to hold an example structure:
```
/a/tags.json
/a/b/tagts.json
/a/b/c/tags.json
/a/b/c/d/tags.json
/a/b/c/d/e/tags.json
/a/b/c/d/e/f/tags.json
/a/b/c/x/tags.json
/a/b/c/y/e/tags.json
```

with each `tags.json` using the folder name as a 'tag' so for example

* `/a/tags.json` == `["a"]`
* `/a/b/tags.json` == `["b"]`

and so on.

Then I'd add a few tests using different base and folder fixtures to check if the logic works correctly.

for example (full path omitted for sake of less typing):

```
resolve(".../a/b", ".../a/b/c/d") == ["b", "c", "d"]
```

and so on

humancalico:
Tests passing :thinking: That doesn't look right

Licenser:
You're right, and this is puzzeling, I've tried it out locally and it looks like the test isn't run, I'll investigate this too.

Licenser (approved):
Awesome, this looks great :D thank you so much :rocket:

PR #879, opened 2021-03-28
Add a stdin onramp.

### LINKED ISSUE #509: Additional Linked sources / sinks
There are some additional sources that could take some linked transport goodness:

* [ ] stdin/stdout/stderr source
* [ ] tcp source
* [ ] tcp sink

## stdin/stdout/stderr source

read from stdin write to stdout, errors go to stderr

## tcp source

stream replie to requests over the tcp connection

## tcp sink 

stream replies for requests on a tcp connection

aryan9600:

# Pull request

## Description
Add an onramp which takes inputs from stdin.

## Related

* Related Issues: fixes #509 
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/99)

## Checklist

* [x] The code is tested

### DIFF
--- a/src/source/stdin.rs
+++ b/src/source/stdin.rs
@@ -1,0 +1,105 @@
+// Copyright 2020-2021, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#![cfg(not(tarpaulin_include))]
+
+use crate::{source::prelude::*, url::TremorUrl};
+use async_std::io;
+
+const INPUT_SIZE_BYTES: usize = 8192;
+
+pub struct Stdin {
+    onramp_id: TremorUrl,
+}
+
+pub struct Int {
+    onramp_id: TremorUrl,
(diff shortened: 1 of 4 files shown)


src/source/stdin.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[async_trait::async_trait()]
+impl Source for Int {
+    fn id(&self) -> &TremorUrl {
+        &self.onramp_id
+    }
+
+    async fn pull_event(&mut self, _id: u64) -> Result<SourceReply> {
+        let mut input = String::new();
+        if let Ok(len) = self.stdin.read_line(&mut input).await {
```
Licenser, line 81: We shouldn't split the input into lines here, that's the job of the `lines` preprocessor. That way it mirrors the behaviour of the `stdout` sink where the `lines` postprocessor adds the newlines to the output (if required)

Licenser (changes requested):
👍 nice job, only small thing is not to use `read_line` but just `read` otherwise looks great :D

src/source/stdin.rs
aryan9600: I don't think `read_line()` splits the input into lines. For example: passing an input like `snot \n badger`, would be treated as one single string "snot \n badger", which is similar to how the websocket onramp works for example. Do let me know if I'm missing something here :D

Licenser: The [documentation](https://docs.rs/async-std/0.99.7/async_std/io/struct.Stdin.html#method.read_line) says "Reads a line of input" so I'm fairly sure it does read a line.

The second issue, enforcing UTF8 compliance as it uses a `String` also is relevant.

aryan9600: Done! ✅ @Licenser

Licenser: nice :D thank you!

Licenser (approved):
Thank you! This is great :D

PR #901, opened 2021-04-07
Add non binding UDP sink support

### LINKED ISSUE #900: UDP offramp caching DNS IPs
UDP offramp current configuration caching DNS IPs and no configurable options to send data using round-robin passion.
https://docs.tremor.rs/artefacts/offramps/#udp
 
We would need data to be sent in round robin passion to the destination endpoint at UDP offramp.

**Notes**

If this is an implementation of an [RFC](https://github.com/tremor-rs/tremor-rfcs) provide a URL
to the RFC this enhancement implements.

If this is a major enhancement or contribution an RFC may be required. It is ok to submit an enhancement
first and our core team will assist with major contributions. In general, major contributions should be
discussed with the community before submission.

Licenser:

# Pull request

## Description

Introduce the `bound` option to the UDP sink which can be set to `false` to let the sink re-acquire the connection to a target for every package

## Related

* Related Issues: fixes #900
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/108)

## Checklist

* [x] The code is tested

## Performance

only sink changes

### DIFF
--- a/src/sink/udp.rs
+++ b/src/sink/udp.rs
@@ -88,6 +88,7 @@ impl Sink for Udp {
                     if self.config.bound {
                         socket.send(&processed).await?;
                     } else {
+                        // reaquire the destination to handle DNS changes or multi IP dns entries
                         socket
                             .send_to(
                                 &processed,
--- a/src/sink/udp.rs
+++ b/src/sink/udp.rs
@@ -42,7 +42,14 @@ pub struct Config {
     pub port: u16,
     pub dst_host: String,
     pub dst_port: u16,
+    #[serde(default = "t")]
+    pub bound: bool,
 }
+
+fn t() -> bool {
+    true
+}
+
 impl ConfigImpl for Config {}
 
 impl offramp::Impl for Udp {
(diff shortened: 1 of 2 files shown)


src/sink/udp.rs
```diff
@@ -42,7 +42,14 @@ pub struct Config {
     pub port: u16,
     pub dst_host: String,
     pub dst_port: u16,
+    #[serde(default = "t")]
+    pub bound: bool,
 }
+
+fn t() -> bool {
```
mfelsche, line 49: can we name this `default_bound`?

src/sink/udp.rs
```diff
@@ -78,7 +85,16 @@ impl Sink for Udp {
...
                 for processed in postprocess(&mut self.postprocessors, ingest_ns, raw)? {
                     //TODO: Error handling
-                    socket.send(&processed).await?;
+                    if self.config.bound {
+                        socket.send(&processed).await?;
+                    } else {
+                        socket
+                            .send_to(
+                                &processed,
+                                (self.config.dst_host.as_str(), self.config.dst_port),
```
mfelsche, line 94: would it be faster/more efficient to store a ready SocketAddr in the config instead of having to go through the `into()` dance for this tuple for every package?

mfelsche (commented):
2 tiny questions

src/sink/udp.rs
mfelsche: Ah, i think the into dance includes the address resolving here, so this actually makes sense at this point. Could you add a comment on that?

mfelsche (approved):

PR #910, opened 2021-04-10
Ensure transactional events send acks/fails in every case

mfelsche:
# Pull request

## Description

First stab at fixing an error condition that was triggered by a pipeline from kafka onramp via `qos::wal` operator to UDP offramp. The `qos::wal` operator was quickly growing full as the UDP offramp didnt sent `CBAction::fail` insights back when it was failing (failed to send).

We fixed this for all offramps by sending `fail` insights in every case if we had an error in `on_event`. This might lead to double `fail` insights in some edge cases, but 1) should hurt any operator or onramp and 2) is better than swallowing it completely.

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/elastic.rs
+++ b/src/sink/elastic.rs
@@ -415,7 +415,7 @@ impl Elastic {
                 // send response events
                 for response in responses {
                     if let Err(e) = response_tx.send(response).await {
-                        error!("Failed to send bulk item response: {}", e);
+                        error!("[Sink::ES] Failed to send bulk item response: {}", e);
                     }
                 }
 
--- a/src/sink/elastic.rs
+++ b/src/sink/elastic.rs
@@ -44,9 +44,10 @@ use halfbrown::HashMap;
 use simd_json::json;
 use std::time::Instant;
 use std::{iter, str};
-use tremor_pipeline::{EventId, EventIdGenerator, OpMeta};
+use tremor_pipeline::{EventId, EventIdGenerator};
 use tremor_script::prelude::*;
 use tremor_script::Object;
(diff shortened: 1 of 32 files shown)


src/sink/elastic.rs
```diff
@@ -112,131 +113,66 @@ fn build_source(id: &EventId, origin_uri: Option<&EventOriginUri>) -> Value<'sta
...
-    es_meta.insert("id".into(), Value::from(item.id().to_string()));
-    es_meta.insert("index".into(), Value::from(item.index().to_string()));
-    es_meta.insert("doc_type".into(), Value::from(item.ty().to_string()));
-
-    es_meta.insert("_id".into(), Value::from(item.id().to_string()));
-    es_meta.insert("_index".into(), Value::from(item.index().to_string()));
-    es_meta.insert("_type".into(), Value::from(item.ty().to_string()));
-
-    meta.insert("elastic".into(), Value::from(es_meta));
+    let mut meta = literal!({
```
Licenser, line 123: <3 `literal!`

tremor-pipeline/src/op/generic/batch.rs
```diff
@@ -39,6 +39,7 @@ pub struct Batch {
     /// the resulting id will be a new distinct id and will be tracking
     /// all event ids (min and max) in the batched event
     batch_event_id: EventId,
+    has_transactional: bool,
```
darach, line 42: To my ears `is_transactional` would be a better name than `has_transactional`

darach (approved):
LGTM 🚀

PR #914, opened 2021-04-12
Fix off-by on in qos::wal operator and change config reqs

mfelsche:
# Pull request

## Description

* Change `qos::wal` operator to only require one of `max_elements` or `max_bytes` (using both is still possible).
* Fix `qos::wal` operator never cleaning up the last event from its storage

## Related

* [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/115)

## Checklist

* [x] The code is tested

### DIFF
--- a/tremor-pipeline/src/op/qos/wal.rs
+++ b/tremor-pipeline/src/op/qos/wal.rs
@@ -21,7 +21,7 @@ use std::mem;
 use std::ops::{Add, AddAssign};
 use tremor_script::prelude::*;
 
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Copy, Default, PartialEq)]
 struct Idx([u8; 8]);
 #[cfg(not(tarpaulin_include))]
 impl std::fmt::Debug for Idx {
@@ -209,9 +209,12 @@ op!(WalFactory(_uid, node) {
 
 impl Wal {
     const URI_SCHEME: &'static str = "tremor-wal";
-    const READ: &'static str = "read";
+    // tree names
     const EVENTS: &'static str = "events";
     const STATE: &'static str = "state";
+    // keys in the state tree
+    const CONFIRMED: &'static str = "confirmed";
+    const READ: &'static str = "read";
 
     fn new(id: String, config: Config) -> Result<Self> {
         let wal = if let Some(dir) = &config.dir {
@@ -223,6 +226,7 @@ impl Wal {
         let state_tree = wal.open_tree(Self::STATE)?;
 
         #[allow(clippy::cast_possible_truncation)]
+        let confirmed = state_tree.get(Self::CONFIRMED)?.map(Idx::from);
         let read = state_tree
             .get(Self::READ)?
             .map(Idx::from)
@@ -231,7 +235,7 @@ impl Wal {
             cnt: events_tree.len() as u64,
             wal,
             read,
-            confirmed: None,
+            confirmed,
             events_tree,
             state_tree,
             config,
@@ -267,6 +271,12 @@ impl Wal {
             let mut event = Event::from_slice(e_slice)?;
             event.transactional = true;
             events.push((OUT, event))
+        }
+        if events.len() > 0 {
+            // track the last read event
+            if let Err(e) = self.state_tree.insert(Self::READ, self.read) {
+                error!("WAL: failed to persist read state: {}", e);
+            }
         }
         self.gc()?;
         Ok(events)
@@ -347,7 +357,7 @@ impl Operator for Wal {
                         // ALLOW: we just set it upstairs
                         .expect("`confirmed` not Some, although just set above, weird!")
                 };
-                if let Err(e) = self.state_tree.insert(Self::READ, confirmed) {
+                if let Err(e) = self.state_tree.insert(Self::CONFIRMED, confirmed) {
                     error!("Failed to persist confirm state: {}", e);
                 }
                 if let Some(e) = self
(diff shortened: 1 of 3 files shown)


tremor-pipeline/src/op/qos/wal.rs
```diff
@@ -184,47 +196,60 @@ op!(WalFactory(_uid, node) {
...
+        }
 
     } else {
         Err(ErrorKind::MissingOpConfig(node.id.to_string()).into())
     }
 });
 
 impl Wal {
+    const URI_SCHEME: &'static str = "tremor-wal";
+    const READ: &'static str = "read";
```
Licenser, line 212: while this is changing, should we rename this as 'confirmed' ?

tremor-pipeline/src/op/qos/wal.rs
```diff
@@ -184,47 +196,60 @@ op!(WalFactory(_uid, node) {
...
+        let read = state_tree
+            .get(Self::READ)?
+            .map(Idx::from)
+            .unwrap_or_default();
         Ok(Wal {
             cnt: events_tree.len() as u64,
             wal,
             read,
-            confirmed: read,
+            confirmed: None,
```
Licenser, line 234: I think it would make sense to track the last read element at least if we store it. This would allow us to handle GC (deletion) more gracefully down the road.

tremor-pipeline/src/op/qos/wal.rs
```diff
@@ -300,17 +335,27 @@ impl Operator for Wal {
                         // This is not for us
                         return;
                     };
-                self.confirmed.set(event_id);
-                if let Err(e) = self.state_tree.insert("read", self.confirmed) {
+
+                trace!("WAL confirm: {}", event_id);
+                let confirmed = if let Some(confirmed) = self.confirmed.as_mut() {
```
Licenser, line 340: could we avoid this with 'get_or_insert'?

tremor-pipeline/src/op/qos/wal.rs
```diff
@@ -322,25 +367,30 @@ impl Operator for Wal {
...
                         "trying to fail a message({}) that was already confirmed({})",
                         event_id, c
                     );
-                    self.confirmed.set(event_id);
-                    if let Err(e) = self.state_tree.insert("read", self.confirmed) {
+                    self.confirmed = Some(Idx::from(event_id));
+                    if let Err(e) = self.state_tree.insert(
+                        Self::READ,
+                        // ALLOW: we just set `self.confirmed` above
+                        self.confirmed.expect("we just set this above, weird!?"),
```
Licenser, line 392: could we just use `*confirmed* here and avoid the expectt?

tremor-pipeline/src/op/qos/wal.rs
```diff
@@ -360,15 +410,17 @@ impl Operator for Wal {
...
         let now_full = self.limit_reached()?;
         // If we just became full or we went from full to non full
         // update the CB status
         let insights = if self.full && !now_full {
+            trace!("WAL not full any more.");
             let mut e = Event::cb_restore(signal.ingest_ns);
             e.origin_uri = self.origin_uri.clone();
             vec![e]
         } else if !self.full && now_full {
+            trace!("WAL full.");
```
Licenser, line 423: I love the idea of adding an event here, would it make sense to bump this to `warn` or at least `info` it seems like a significant event.

tremor-pipeline/src/op/qos/wal.rs
mfelsche: i dont think so, it is used for the pointer to the event we have already read and emitted downstream, no?

mfelsche: ah, now i see what you mean, yeah makes sense

tremor-pipeline/src/op/qos/wal.rs
mfelsche: Sooooo much better!

mfelsche: aaah, no, doesnt work, the mut reference makes the borrow checker unhappy.

tremor-pipeline/src/op/qos/wal.rs
mfelsche: Set it to `warn`. Do you mean emitting a real `Event` here? I would not do that. Logging should suffice.

tremor-pipeline/src/op/qos/wal.rs
Licenser: waht's the issue the borrow checker is throwing? something that might be solvable?

tremor-pipeline/src/op/qos/wal.rs
Licenser: oh yes just warning that it's full is fine, logging the event would probably be terrible and blow things up :D

tremor-pipeline/src/op/qos/wal.rs
mfelsche: we get a `&mut` reference back and this is consumed in `self.state_tree.insert(...)` and thus cannot be used in `events_tree.get()` later on.

tremor-pipeline/src/op/qos/wal.rs
mfelsche: I hope what I added is what you meant.

tremor-pipeline/src/op/qos/wal.rs
Licenser: would `let confirmed: & <whatevertypeitis> = confirmed` work to downgrade it from a mutable to a immutable reference?

mfelsche: it does! :heart: Man, this shadowing is really dirty.

Licenser (approved):

darach (approved):
LGTM

Issue #724, opened 2021-01-28
Enhance tremor with connectors for the Google Cloud Platform


darach:
Add support for the GCP suite of APIs to tremor as connectors ( source, sink, linked transport )
minimally to offer support for GCS and PubSub, but with reuse in mind for extending API support
to the GCP suite of APIs in tremor over time.

**Describe the problem you are trying to solve**

Minimally enhance tremor with support for GCS and PubSub.

**Describe the solution you'd like**

Ideally the plumbing and integration for GCP and its suite of APIs would be a
separate rust crate / project usable by the rust community at large. One possible
approach to implementation would be adding a `rust` code generation target to
Google's own API client generation toolkit: https://github.com/googleapis/gapic-generator

**Notes**

Recommended skills - A basic working knowledge of `rust` and `java` with some exposure
to tools and frameworks that generate bindings against declarative API specifications to
produce source code. A basic working knowledge of CI, particulary github actions.

imskr:
I would to work on this issue. Please assign me :)

Licenser:
Hi there :D, thank you so much for your interest!

Feel free to hop by our discord to chat.

imskr:
sure @licenser

SantamRC:
Hello @Licenser @darach,
This is Santam. I have basic understading of Rust and Google Cloud Platform. I am interested to work on this issue.

darach:
Hi @SantamRC,

Thank you for your interest. Feel free to hop by our discord to chat ( https://bit.ly/tremor-discord )

Cheers,

Darach.

PR #917, opened 2021-04-13
Google Cloud Storage connectors

jigyasak05:
# Pull request

## Description

Contributes a Google Cloud Storage linked transport sink to tremor.

## Related

Partial completion of https://github.com/tremor-rs/tremor-runtime/issues/724

## Checklist

* [x] The code is tested

## Performance

No access to the production test environment. Not done.


src/sink/gcs.rs
```diff
@@ -0,0 +1,345 @@
...
+            return Err("Invalid Command, expected `command` field".into());
+        };
+
+        let command = match cmd_name {
+            "fetch" => StorageCommand::Fetch(
+                if let Some(Value::String(bucket_name)) = o.get("bucket") {
+                    bucket_name.to_string()
+                } else {
+                    return Err("Invalid Command, expected `bucket` field".into());
+                },
```
Licenser, line 119: Since this pattern repeats quite a lot here a suggestion how it can be re-written to make it a bit easier to read, especially since it's in a struct field:

```suggestion
                o.get_str("bucket").ok_or("Invalid Command, expected `bucket` field")?.to_string()
```

darach (approved):
LGTM Super awesome work! 🎸

src/connectors/gcp/storage.rs
```diff
@@ -0,0 +1,139 @@
...
+use crate::errors::Result;
+use reqwest::Client;
+use tremor_value::Value;
+
+pub(crate) async fn get_object(
+    client: &Client,
+    bucket_name: &str,
+    object_name: &str,
+) -> Result<Value<'static>> {
+    let url = format!(
```
Licenser, line 26: It would be nicer to use `Url` and build the URL instead of a string format. including variables like that opens this up to possible injection attacks as no escaping takes place. since the gcs code handles secrets we might want to be extra careful here.

(same below for the other `format!` calls)

src/sink/gcs.rs
```diff
@@ -0,0 +1,345 @@
...
+        _input: &str,
+        _codec: &mut dyn Codec,
+        _codec_map: &HashMap<String, Box<dyn Codec>>,
+        mut event: Event,
+    ) -> ResultVec {
+        let remote = if let Some(remote) = &self.remote {
+            remote
+        } else {
+            self.remote = Some(auth::json_api_client(&self.config.pem, &HeaderMap::new()).await?);
+            let remote = self.remote.as_ref().ok_or("Client error!")?;
```
Licenser, line 223: That's nice I dind't knew you could include a `&str` in the `ok_or` I always used Error::from() this is a nice trick. Thank you!

src/sink/gcs.rs
```diff
@@ -0,0 +1,345 @@
...
+                "data": response
+            })
+            .into();
+            if self.is_linked {
+                if let Some(reply_channel) = &self.reply_channel {
+                    reply_channel
+                        .send(sink::Reply::Response(
+                            OUT,
+                            Event {
+                                id: EventId::new(1, 2, 3),
```
Licenser, line 279: this should be a real incremental ID or it will cause problems in the pipeline

src/sink/gcs.rs
```diff
@@ -0,0 +1,345 @@
...
+        self.reply_channel = Some(reply_channel);
+        self.is_linked = is_linked;
+        Ok(())
+    }
+    async fn on_signal(&mut self, signal: Event) -> ResultVec {
+        if self.is_down && self.qos_facility.probe(signal.ingest_ns) {
+            self.is_down = false;
+            // This means the port is connectable
+            info!("Google Cloud Storage -  sink remote endpoint - recovered and contactable");
+            self.is_down = false;
```
Licenser, line 329: We won't need to set this twice :)

src/sink/gcs.rs
```diff
@@ -0,0 +1,345 @@
...
+            })
+            .into();
+            if self.is_linked {
+                if let Some(reply_channel) = &self.reply_channel {
+                    reply_channel
+                        .send(sink::Reply::Response(
+                            OUT,
+                            Event {
+                                id: EventId::new(1, 2, 3),
+                                data: LineValue::new(vec![], |_| ValueAndMeta::from(response)),
```
Licenser, line 280: We'll need to include the corresponding correlation metadata here. the `es` sink has a good example for that

Licenser (commented):
Looks great overall :D that's really good progress! I've just left a few minor comments.

As a not for a next iteration, it'd be nice to treat the request asynchronously so the onramp won't block until it's totally completed but that has no urgency.

src/sink/gcs.rs
darach: Yep - we'll hit these in the next revision as live testing in a real environment will no doubt uncover some enhancements before we mark this as stable ...

src/sink/gcs.rs
darach: This is boilerplate as the failover/detector is yet to be tested

darach (approved):
LGTM 🎸  - Happy days!

darach (approved):
Random CI test failures impacting the merge - manually merging

darach (approved):
It's groundhog day! 👍

jigyasak05:
Indeed!

PR #918, opened 2021-04-13
Add dns sink

Licenser:
# Pull request

## Description

Add a linked sync that allows performing DNS queries

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/116)

## Checklist

* [x] The code is tested

## Performance
Only additional code

### DIFF
--- a/src/sink/dns.rs
+++ b/src/sink/dns.rs
@@ -54,62 +54,101 @@ impl offramp::Impl for Dns {
     }
 }
 
-fn str_to_record_type(s: &str) -> Option<RecordType> {
+fn str_to_record_type(s: &str) -> Result<RecordType> {
     match s {
-        "A" => Some(RecordType::A),
-        "AAAA" => Some(RecordType::AAAA),
-        "ANAME" => Some(RecordType::ANAME),
-        "ANY" => Some(RecordType::ANY),
-        "AXFR" => Some(RecordType::AXFR),
-        "CAA" => Some(RecordType::CAA),
-        "CNAME" => Some(RecordType::CNAME),
-        "HINFO" => Some(RecordType::HINFO),
-        "HTTPS" => Some(RecordType::HTTPS),
-        "IXFR" => Some(RecordType::IXFR),
-        "MX" => Some(RecordType::MX),
-        "NAPTR" => Some(RecordType::NAPTR),
-        "NS" => Some(RecordType::NS),
-        "NULL" => Some(RecordType::NULL),
-        "OPENPGPKEY" => Some(RecordType::OPENPGPKEY),
-        "OPT" => Some(RecordType::OPT),
-        "PTR" => Some(RecordType::PTR),
-        "SOA" => Some(RecordType::SOA),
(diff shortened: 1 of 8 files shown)


src/sink/dns.rs
```diff
@@ -0,0 +1,192 @@
...
+        for e in event.value_iter() {
+            if let Some(lookup) = e.get("lookup") {
+                if let Some(name) = lookup.as_str().or_else(|| lookup.get_str("name")) {
+                    let data = if let Some(record_type) =
+                        lookup.get_str("type").and_then(str_to_record_type)
+                    {
+                        // type lookup
+                        lookup_to_value(
+                            &resolver
+                                .lookup(name, record_type, DnsRequestOptions::default())
```
mfelsche, line 140: what happens on invalid `record_type`? I think it makes sense to check for a valid one here.

src/sink/dns.rs
```diff
@@ -0,0 +1,192 @@
...
+        _codec: &mut dyn Codec,
+        _codec_map: &HashMap<String, Box<dyn Codec>>,
+        event: Event,
+    ) -> ResultVec {
+        let resolver = self
+            .resolver
+            .as_ref()
+            .ok_or_else(|| Error::from("No resolver set"))?;
+        let mut res = Vec::with_capacity(event.len());
+        let correlation = event.correlation_meta();
```
mfelsche, line 129: In case we have a batched event, this will put all the correlation values from each batched event into a single `Value` and below it will be replicated into each response event.

I would suggest using the value_meta_iter() and extract the `$correlation` value from the meta manually for each (possibly) batched event. I added the `Event::correlation_meta()` method for times when i need to drag that on into different scopes etc. across tasks etc.

src/sink/dns.rs
```diff
@@ -0,0 +1,192 @@
...
+                        // generic lookup
+                        lookup_to_value(resolver.lookup_ip(name).await?.as_lookup())
+                    };
+                    let e = Event {
+                        data: (data, correlation.clone().unwrap_or_default()).into(),
+                        origin_uri: Some(self.event_origin_uri.clone()),
+                        ..Event::default()
+                    };
+
+                    res.push(Reply::Response("out".into(), e))
```
mfelsche, line 153: we have some constants for those ports around somewhere.

src/sink/dns.rs
```diff
@@ -0,0 +1,192 @@
...
+                } else {
+                    error!("Invalid DNS request: {}", e)
```
mfelsche, line 155: I think we should add Error events in this case and in all other error cases too and send them out the `err` port.

src/sink/dns.rs
Licenser: It would default to a `A` record

Licenser: it will now error

mfelsche (approved):
fine with me, if CI is on our side.

mfelsche:
I really like how much nicer the code got with the `literal!` and other improvements!

PR #919, opened 2021-04-13
Add tremor_value `literal` macro

darach:
# Pull request

## Description

Adds a `tremor_value::literal` macro which provides a `json!` like macro but for
the tremor_value variant.

## Related

This feature requires at least v0.4.1 of simd-json

## Checklist

* [x] The code is tested

## Performance

The introduction of `tremor_value::literal!` replaces occurances of `simd_json::json!(...).into()` which
first creates an OwnedValue and then transforms these to borrowed values. The new macro avoids the
redundant transient copies ab initio.

### DIFF
--- a/tremor-value/src/macros.rs
+++ b/tremor-value/src/macros.rs
@@ -92,7 +92,8 @@ macro_rules! value_internal {
 
     // Insert the current entry followed by trailing comma.
     (@object $object:ident [$($key:tt)+] ($value:expr) , $($rest:tt)*) => {
-        let _x = $object.insert(($($key)+).into(), $value);
+        // ALLOW: let _
+        let _ = $object.insert(($($key)+).into(), $value);
         value_internal!(@object $object () ($($rest)*) ($($rest)*));
     };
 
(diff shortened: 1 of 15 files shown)


tremor-value/src/macros.rs
```diff
@@ -0,0 +1,331 @@
...
+    // We require two copies of the input tokens so that we can match on one
+    // copy and trigger errors on the other copy.
+    //////////////////////////////////////////////////////////////////////////
+
+    // Done.
+    (@object $object:ident () () ()) => {};
+
+    // Insert the current entry followed by trailing comma.
+    (@object $object:ident [$($key:tt)+] ($value:expr) , $($rest:tt)*) => {
+        let _ = $object.insert(($($key)+).into(), $value);
```
Licenser, line 95: Since you updated to the newest value trait, you can use `try_insert` here and avoid the `let _ ` :D

Licenser: I stand corrected this isn't a `Value::Object` just a `Object` :(

Licenser (approved):
Nice :D :+1:

PR #922, opened 2021-04-14
Add udp metadata

Licenser:
# Pull request

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/118)

## Checklist

* [x] The code is tested

## Performance

Only sink changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add DNS sink
 * Add syslog codec.
 * Add `$udp.host` and `$udp.port` to allow controling udp packet destinations on a per event basis.
+* Depricate `udp.dst_*` config, introduce `udp.bind.*` config instead.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Change `qos::wal` operator to only require one of `max_elements` or `max_bytes` (using both is still possible).
 * Add DNS sink
 * Add syslog codec.
+* Add `$udp.host` and `$udp.port` to allow controling udp packet destinations on a per event basis.
 
 ### Fixes
 
--- a/src/codec/string.rs
+++ b/src/codec/string.rs
@@ -69,4 +69,15 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn test_string_codec2() -> Result<()> {
+        let seed = literal!(["snot badger"]);
+
+        let codec = String {};
+        let as_raw = codec.encode(&seed)?;
+        assert_eq!(as_raw, b"[\"snot badger\"]");
+
+        Ok(())
+    }
 }
--- a/src/codec/string.rs
+++ b/src/codec/string.rs
@@ -56,13 +56,11 @@ impl Codec for String {
 #[cfg(test)]
 mod test {
     use super::*;
-    use simd_json::json;
-    use simd_json::OwnedValue;
+    use tremor_value::literal;
 
     #[test]
     fn test_string_codec() -> Result<()> {
-        let seed: OwnedValue = json!("snot badger");
-        let seed: Value = seed.into();
+        let seed = literal!("snot badger");
 
         let mut codec = String {};
         let mut as_raw = codec.encode(&seed)?;
--- a/src/sink/udp.rs
+++ b/src/sink/udp.rs
@@ -36,14 +36,30 @@ pub struct Udp {
 }
 
 #[derive(Deserialize, Debug)]
+struct Host {
+    host: String,
+    port: u16,
+}
+impl Default for Host {
+    fn default() -> Self {
+        Self {
+            host: String::from("0.0.0.0"),
+            port: 0,
+        }
+    }
+}
+
+#[derive(Deserialize, Debug)]
 pub struct Config {
     /// Host to use as source
-    pub host: String,
-    pub port: u16,
-    pub dst_host: String,
-    pub dst_port: u16,
+    host: String,
+    port: u16,
+    dst_host: Option<String>,
+    dst_port: Option<u16>,
+    #[serde(default = "Host::default")]
+    bind: Host,
     #[serde(default = "t")]
-    pub bound: bool,
+    bound: bool,
 }
 
 fn t() -> bool {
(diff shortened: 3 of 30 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * Change `qos::wal` operator to only require one of `max_elements` or `max_bytes` (using both is still possible).
+* Add `$udp.host` and `$udp.port` to allow controling udp packet destinations on a per event basis.
```
mfelsche, line 8: while i like the terseness of those names, i would think about making it consistent with the config, where it is `dst_host` and `dst_port`. `host` and `port` are the local socket address, this might be confusing.

Licenser: Ja that's a bit of a conflict, on the other hand in the `rest` metadata we call it `host` and `port` so naming it `dst_host` and `dst_port` will confuse people who used this mechanism from rest before :( 

how about we depreciate `dst_host` and `dst_port` (warning for this release) and add a `dst` struct that would be `dst.host`, `dst.port` that way we can keep short names and stay consistent with the config?

```yaml
offramp:
  - id: udp
    type: udp
    config:
      host: 0.0.0.0
      port: 0
      dst:
        host: snot
        port: 42
```

or we ad `bind` for current `host` and `port` to allow specifying the binding locally?


```yaml
offramp:
  - id: udp
    type: udp
    config:
      bind:
        host: 0.0.0.0
        port: 0
      host: snot
      port: 42
```


(I kind of like the second option since the top-level host/port for the binding address always struck me as wrong)

mfelsche: Lets go with the second option. :)
But then we should rename the `bound` config to `connected`, as the setting, if `false`, avoids the `connect` call, not the bind call.

Licenser: Oh that's a good point! If we think further, how about we depricate `bound` altogether with this and enforce re-connects to be done via metadata, that will reinforce the feeling of "this happens on every event since I've to set it on every event"?

mfelsche: kill it with fire! :fire:

src/sink/udp.rs
```diff
@@ -55,7 +71,20 @@ impl ConfigImpl for Config {}
 impl offramp::Impl for Udp {
     fn from_config(config: &Option<OpConfig>) -> Result<Box<dyn Offramp>> {
         if let Some(config) = config {
-            let config: Config = Config::new(config)?;
+            let mut config: Config = Config::new(config)?;
+            if !config.bound {
+                warn!("The `bound` setting of the UDP offramp is depricated, in futur re-binding will only work over the $udp metadata")
```
mfelsche, line 76:
```suggestion
                warn!("The `bound` setting of the UDP offramp is deprecated, in future re-binding will only work over the $udp metadata")
```

src/sink/udp.rs
```diff
@@ -55,7 +71,20 @@ impl ConfigImpl for Config {}
...
+            }
+            if let Some(dst_port) = config.dst_port.take() {
+                warn!("The `dst_port` setting of the UDP offramp is depricated, this will in future be `port` and `port` will be `bind.port`.");
```
mfelsche, line 79:
```suggestion
                warn!("The `dst_port` setting of the UDP offramp is deprecated, this will in future be `port` and current `port` will be `bind.port`.");
```

src/sink/udp.rs
```diff
@@ -55,7 +71,20 @@ impl ConfigImpl for Config {}
...
+                config.bind.port = config.port;
+                config.port = dst_port;
+            }
+            if let Some(dst_host) = config.dst_host.take() {
+                warn!("The `dst_host` setting of the UDP offramp is depricated, this will in future be `host` and `host` will be `bind.host`.");
```
mfelsche, line 84:
```suggestion
                warn!("The `dst_host` setting of the UDP offramp is deprecated, this will in future be `host` and current `host` will be `bind.host`.");
```

src/sink/udp.rs
```diff
@@ -81,19 +110,19 @@ impl Sink for Udp {
...
                 for processed in postprocess(&mut self.postprocessors, ingest_ns, raw)? {
-                    //TODO: Error handling
-                    if self.config.bound {
+                    let udp = meta.get("udp");
+                    if let Some((host, port)) = udp.get_str("host").zip(udp.get_u16("port")) {
+                        socket.send_to(&processed, (host, port)).await?;
+                    } else if self.config.bound {
                         socket.send(&processed).await?;
                     } else {
+                        warn!("using `bound` in the UDP sink is depricated please use $udp.host and $udp.port instead!");
```
mfelsche, line 122:
```suggestion
                        warn!("using `bound` in the UDP sink config is deprecated please use $udp.host and $udp.port instead!");
```

src/sink/udp.rs
```diff
@@ -128,19 +157,21 @@ impl Sink for Udp {
         _reply_channel: Sender<sink::Reply>,
     ) -> Result<()> {
         self.postprocessors = make_postprocessors(processors.post)?;
-        let socket = UdpSocket::bind((self.config.host.as_str(), self.config.port)).await?;
+        let socket =
+            UdpSocket::bind((self.config.bind.host.as_str(), self.config.bind.port)).await?;
         socket
-            .connect((self.config.dst_host.as_str(), self.config.dst_port))
+            .connect((self.config.host.as_str(), self.config.port))
```
mfelsche, line 163: this should also be guarded by `self.config.bound` as it is done in `on_signal`, no?

mfelsche (commented):
some small typos and a question, otherwise LGTM! :rocket:

src/sink/udp.rs
Licenser: Good catch! you're spot on :D

Licenser: extracted the whole bind logic in a function that gets called in both places :+1:

src/codec/string.rs
```diff
@@ -71,4 +69,15 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn test_string_codec2() -> Result<()> {
```
mfelsche, line 74: i see what you did there, sneakily upping the coverage :P

mfelsche (approved):

src/codec/string.rs
Licenser: Ja there was no new uncovered line :( so I had to be creative. string had the lowest coverage I could find so 🗡️ that test!

PR #928, opened 2021-04-15
Fix syslog to default to parsing with UTC timezone if none is provided.

mfelsche:
It was taking local time, which lead to test fails depending on executing machine.

# Pull request

## Checklist

* [x] The code is tested


Licenser (approved):
:rocket:

PR #942, opened 2021-04-20
Allow insights/contraflow events to traverse through multiple connected pipelines.

mfelsche:
# Pull request

## Description

This is still an experimental feature. This fixes the issue of insights/contraflow events not being forwarded across two connected pipelines, this breaks onramps/operators that require ack/fail insights (e.g. kafka onramps with `enable.auto.commit: false`).

## Checklist

* [x] The code is tested

### DIFF
--- a/src/pipeline.rs
+++ b/src/pipeline.rs
@@ -74,8 +74,8 @@ impl Addr {
         Ok(self.addr.send(msg).await?)
     }
 
-    pub(crate) fn try_send(&self, msg: Msg) -> bool {
-        self.addr.try_send(msg).is_ok()
+    pub(crate) fn try_send(&self, msg: Msg) -> Result<()> {
+        Ok(self.addr.try_send(msg)?)
     }
 
     pub(crate) async fn send_mgmt(&self, msg: MgmtMsg) -> Result<()> {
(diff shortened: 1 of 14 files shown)


src/pipeline.rs
```diff
@@ -94,15 +94,26 @@ pub(crate) enum CfMsg {
     Insight(Event),
 }
 
+#[derive(Debug)]
+pub(crate) enum ConnectTarget {
```
darach, line 98: Nice! 💅

src/pipeline.rs
```diff
@@ -207,25 +247,64 @@ async fn handle_insight(
...
-                    .await
-                {
-                    error!("[Pipeline] failed to send to onramp: {} {:?}", e, &o);
+                if let Err(e) = match input {
+                    Input::Onramp(addr) => addr
+                        .send(onramp::Msg::Cb(insight.cb, insight.id))
+                        .await
+                        .map_err(Error::from),
+                    Input::Pipeline(addr) => addr.send_insight(insight).await,
+                    // linked offramps dont support insights yet, although they definitely should
```
darach, line 288: Make this a `TODO`

darach (changes requested):
A few minor tweaks and we're good - loving it!

src/pipeline.rs
mfelsche: Yes, definitely! Will ticket this as well

darach (approved):
LGTM 👍

PR #954, opened 2021-04-23
Trying to fix flaky test with timing issues.

mfelsche:
Bumped up the timeout quite a bit. If it still fails after that timeout, we have some serious problems in our runtime, which is very unlikely.

## Checklist

* [x] The code is tested

### DIFF
--- a/src/pipeline.rs
+++ b/src/pipeline.rs
@@ -691,7 +691,8 @@ mod tests {
         .await?;
 
         // transactional onramp received it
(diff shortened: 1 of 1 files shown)


darach (approved):
LGTM worth a shot 🤞

mfelsche:
I am re-running the CI job a few times and see if it comes back

darach (approved):
LGTM 👍

PR #957, opened 2021-04-26
Add textual-length-prefix pre and postprocessor

murex971:
# Pull request

## Description

add support for the RFC 5425 transport protocol (textual length prefix)

## Checklist

* [x] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Allow insights/contraflow events to traverse through multiple connected pipelines
 
 * Add GCP Cloud Storage linked sink connector.
+* Add textual-length-prefix pre and postprocessor.
 
 ### Fixes
 
--- a/src/postprocessor.rs
+++ b/src/postprocessor.rs
@@ -23,6 +23,7 @@ use tremor_common::time::nanotime;
 pub type Postprocessors = Vec<Box<dyn Postprocessor>>;
 use std::io::Write;
 use std::mem;
+use std::str;
 
 /// Postprocessor trait
 pub trait Postprocessor: Send {
@@ -54,6 +55,7 @@ pub fn lookup(name: &str) -> Result<Box<dyn Postprocessor>> {
         "ingest-ns" => Ok(Box::new(AttachIngresTs {})),
         "length-prefixed" => Ok(Box::new(LengthPrefix::default())),
         "gelf-chunking" => Ok(Box::new(Gelf::default())),
+        "textual-length-prefix" => Ok(Box::new(TextualLength::default())),
         _ => Err(format!("Postprocessor '{}' not found.", name).into()),
     }
 }
@@ -244,6 +246,25 @@ impl Postprocessor for LengthPrefix {
     }
 }
 
+#[derive(Clone, Default)]
+pub(crate) struct TextualLength {}
+impl Postprocessor for TextualLength {
+    #[cfg(not(tarpaulin_include))]
+    fn name(&self) -> &str {
+        "textual-length-prefix"
+    }
+
+    fn process(&mut self, _ingres_ns: u64, _egress_ns: u64, data: &[u8]) -> Result<Vec<Vec<u8>>> {
+        let size = data.len();
+        let mut digits: Vec<u8> = size.to_string().into_bytes();
+        let mut res = Vec::with_capacity(digits.len() + 1 + size);
+        res.append(&mut digits);
+        res.push(32);
+        res.write_all(&data)?;
+        Ok(vec![res])
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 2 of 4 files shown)


src/postprocessor.rs
```diff
@@ -244,6 +246,25 @@ impl Postprocessor for LengthPrefix {
...
+#[derive(Clone, Default)]
+pub(crate) struct TextualLength {}
+impl Postprocessor for TextualLength {
+    #[cfg(not(tarpaulin_include))]
+    fn name(&self) -> &str {
+        "textual-length"
+    }
+
+    fn process(&mut self, _ingres_ns: u64, _egress_ns: u64, data: &[u8]) -> Result<Vec<Vec<u8>>> {
+        let mut res = Vec::new();
```
mfelsche, line 258: please allocate a new array with the known length. with something like: `digits.len() + 1 + data.len()`. This will be faster and avoid copying stuff around when the Vec needs to be resized.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Allow insights/contraflow events to traverse through multiple connected pipelines
 
 * Add GCP Cloud Storage linked sink connector.
+* Add textual-prefix-length pre and postprocessor.
```
mfelsche, line 15:
```suggestion
* Add textual-length-prefix pre and postprocessor.
```

src/postprocessor.rs
```diff
@@ -54,6 +55,7 @@ pub fn lookup(name: &str) -> Result<Box<dyn Postprocessor>> {
         "ingest-ns" => Ok(Box::new(AttachIngresTs {})),
         "length-prefixed" => Ok(Box::new(LengthPrefix::default())),
         "gelf-chunking" => Ok(Box::new(Gelf::default())),
+        "textual-length-prefix" => Ok(Box::new(TextualLength::default())),
```
mfelsche, line 58: maybe also add `textual-length-prefixed` as lookup key here, so we are consistent with the `length-prefixed` postprocessor.

src/postprocessor.rs
```diff
@@ -244,6 +246,25 @@ impl Postprocessor for LengthPrefix {
     }
 }
 
+#[derive(Clone, Default)]
+pub(crate) struct TextualLength {}
+impl Postprocessor for TextualLength {
+    #[cfg(not(tarpaulin_include))]
+    fn name(&self) -> &str {
+        "textual-length"
```
mfelsche, line 254: What do you think about this, to be consistent with the lookup key above?

```suggestion
        "textual-length-prefix"
```

src/preprocessor.rs
```diff
@@ -379,6 +381,54 @@ impl Preprocessor for LengthPrefix {
...
 }
+#[derive(Clone, Default, Debug)]
+pub(crate) struct TextualLength {
+    len: Option<usize>,
+    buffer: BytesMut,
+}
+impl Preprocessor for TextualLength {
+    #[cfg(not(tarpaulin_include))]
+    fn name(&self) -> &str {
+        "textual-length"
```
mfelsche, line 392: same here. :)

src/postprocessor.rs
```diff
@@ -270,4 +291,12 @@ mod test {
 
         assert_eq!(Ok(vec![b"c25vdA==".to_vec()]), post.process(0, 0, b"snot"));
     }
+
+    #[test]
+    fn textual_length_prefix_postp() {
```
mfelsche, line 296: how would this postprocessor handle an empty `data` array? worth adding a test for it?

src/preprocessor.rs
```diff
@@ -400,6 +450,103 @@ mod test {
...
+        let mut in_ns = 0_u64;
+        let mut res = pre_p.process(&mut in_ns, data.as_bytes()).unwrap();
+        assert_eq!(1, res.len());
+        let payload = res.pop().unwrap();
+        assert_eq!(42, payload.len());
+    }
+
+    #[test]
+    fn empty_textual_prefix() {
+        let data = ("").as_bytes();
```
mfelsche, line 543: Seeing this, i also thought about an empty textual-prefix-length, like so `0 `. This should be valid, right? You think this is worth adding a test too?

src/preprocessor.rs
```diff
@@ -400,6 +450,103 @@ mod test {
...
 
+    fn textual_prefix(len: usize) -> String {
+        format!("{} {}", len, String::from_utf8(vec![b'O'; len]).unwrap())
+    }
+
+    use proptest::prelude::*;
+
+    // generate multiple chopped length-prefixed strings
+    fn multiple_textual_lengths(max_elements: usize) -> BoxedStrategy<(Vec<usize>, Vec<String>)> {
+        proptest::collection::vec(".+", 1..max_elements) // generator for Vec<String> of arbitrary strings, maximum length of vector: `max_elements`
```
mfelsche, line 461: Maybe, if we change the regex in the string to `"+*"` we will also get empty strings and could test the preprocessor against that?

mfelsche (commented):
Great stuff! 

I found some minor things too look for, before we get this in.

If you like, you could also add a property based test that tests that the preprocessor can handle everything the postprocessor produces. But this is a purely optional nice-to-have.

:+1: 🚀

Licenser:
This is a great contribution :+1: really nice set of tests too! Great job :D

mfelsche (approved):

PR #961, opened 2021-04-28
ci: deny warnings in ci only

humancalico:
Ref:
https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html

Signed-off-by: Akshat Agarwal <humancalico@disroot.org>

# Pull request

## Related

## Checklist

* [x] The code is tested

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -37,27 +37,27 @@ jobs:
       - name: Build test runner
         run: cargo build -p tremor-cli
         env:
-          RUSTFLAGS: "-C target-feature=${{ matrix.target_feature }}"
+          RUSTFLAGS: -D warnings -C target-feature=${{ matrix.target_feature }}
       - name: Run Integration Tests
         run: cargo run -p tremor-cli -- test integration tremor-cli/tests ${{ matrix.exclude_tests }}
         env:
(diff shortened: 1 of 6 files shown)


Licenser:
I'm not a fan of this change, what's the reasoning for only denying warnings in CI?

humancalico:
> I'm not a fan of this change, what's the reasoning for only denying warnings in CI?

From https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html. 
> By disallowing the compiler to build with warnings, a crate author opts out of Rust's famed stability. Sometimes new features or old misfeatures need a change in how things are done, thus lints are written that warn for a certain grace period before being turned to deny.

Licenser:
I can see that reasoning for crates that are published to creates.io, but most of tremor isn't. How about we remove the deny warning from those parts that are publically shared crates and keep it in the parts that aren't.

We do bind the version of rust tremor compiles with, so the problem that the article mentions isn't really one for us and I would air on the side of strictness over convenience that isn't really a convenience since we have the rust version pinned so updates always force a human to decide on them.

Licenser (changes requested):
as discussed in the comments

Licenser:
Awesome :D thank you for the work and contribution!

Licenser (approved):

PR #969, opened 2021-04-30
Trying to further optimize the record pattern

mfelsche:
# Pull request

## Description

As we have an hashmap available, i tweaked the `KnownKey` API to be able to work on those maps, so we can spare a match.

## Checklist

* [x] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add `op` key to KV offramp responses in order to differentiate responses by the command that triggered them
 * Add `key` to KV offramp responses.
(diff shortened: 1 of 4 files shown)


Licenser (approved):
I love it :D that should be a nice bump :rocket: I'll run benchmarks in a moment :)

Could we also back port that to simd_json's `KnownKey`?

mfelsche:
Will do the backport this weekend :)

mfelsche:
I think there are still a few places left where we can apply those new functions.

Licenser:
Benchmarks say +/- 0% diff as of this entry, but the code changes are still nice :D

Issue #607, opened 2020-11-26
Add git commit hashes to version for non-official release builds


darach:
**Describe the problem you are trying to solve**

Add git commit hashes to version for non-official release builds

**Describe the solution you'd like**

The feature branch name and commit hash should be represented in version information
emitted by tremor in the cli, api and other output for developer builds. This aids in replication
of issues and in identifying heritage of a build when multiple versions of tremor are in use
on the same environment.

**Notes**

If this is an implementation of an [RFC](https://github.com/tremor-rs/tremor-rfcs) provide a URL
to the RFC this enhancement implements.

If this is a major enhancement or contribution an RFC may be required. It is ok to submit an enhancement
first and our core team will assist with major contributions. In general, major contributions should be
discussed with the community before submission.

GaryPWhite:
yea sounds like a good'n I'm gonna give this a swing

GaryPWhite:
close? :)

mfelsche:
C L O S E ! ! !

PR #977, opened 2021-05-03
Provide branch/commit in version output

GaryPWhite:
# Pull request

## Description

Adds explicit version tracking for branches/commits not made on the `main` branch.

## Related

* Related Issues: https://github.com/tremor-rs/tremor-runtime/issues/607

## Checklist

* [x] The code is tested


mfelsche:
@GaryPWhite in case you didnt find i yet, we have the command `make clippy` we run before we push a PR, it will show you all nitpicks from the whole codebase. Hope this helps. Clippy is a really strict master.

Licenser:
I'm not sure this fully resolves the issue, since building tremor without docker / makefile won't include the hash. Would it be possible to leverage the build.rs to intern this natively into the build process?

GaryPWhite:
ah, I assumed the build processes used for any dev work would be through the makefile... I can look into getting the same input to `build.rs`!

Licenser:
Really nice! Only one tiny thing: lets default to something like `"<NOT BUILT FROM GIT>"` (or really any sensible placeholder / warning )" when we can't use got, otherwise tarball builds won't work and I think at least nix depends on that. Otherwise, it'd cause people with tarballs problems :), that aside I love it Darn good work :D

GaryPWhite:
had to update the `dockerfile` -- found that after we include the `build` argument, it would detect all the surrounding folders included in the `rust` image. Instead of building on `/`, it builds now in `/app`

build.rs
```diff
@@ -0,0 +1,60 @@
...
+        .output();
+    if let Ok(branch_output) = branch {
+        let branch_string = String::from_utf8_lossy(&branch_output.stdout);
+        return branch_string
+            .lines()
+            .next()
+            .unwrap_or("unknown_branch")
+            .to_string();
+    } else {
+        panic!("Can not get git branch: {}", branch.unwrap_err());
```
Licenser, line 31: we shouldn't panic in the build script

build.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    if let Ok(commit_output) = commit {
+        let commit_string = String::from_utf8_lossy(&commit_output.stdout);
+        return commit_string
+            .lines()
+            .next()
+            .unwrap_or("unknown_hash")
+            .to_string();
+    } else {
+        panic!("Can not get git commit: {}", commit.unwrap_err());
```
Licenser, line 52: see above

src/version.rs
```diff
@@ -17,31 +17,47 @@ use rdkafka::util::get_rdkafka_version;
...
 pub const DEBUG: bool = true;
 
+#[must_use]
+/// Provides formatting for "long" version name of build
+pub fn long_ver() -> String {
+    #[cfg(not(debug_assertions))]
+    const VERSION_LONG: &str = env!("CARGO_PKG_VERSION");
+    #[cfg(debug_assertions)]
+    const VERSION_LONG: &str = concat!(env!("CARGO_PKG_VERSION"), " (DEBUG)");
+    return match option_env!("VERSION_BRANCH") {
```
Licenser, line 34: This could be un-nested to:

```
    return match option_env!("VERSION_BRANCH") {
        Some("main") => VERSION_LONG.to_string(),
        Some(branch) => {
              // additional version info otherwise
              let commit_hash: &str = option_env!("VERSION_HASH").map_or("", |hash| hash);
              format!("{} {}:{}", VERSION_LONG, branch, commit_hash)
        }
        None => VERSION_LONG.to_string(), // default if option env not set
    };
```

Licenser (changes requested):
Aside from the panics in the `build.rs` this looks great :+1:

optional but not urgent would be, I just realized this yesterday, adding it to the `/version` endpoint in rest but that can be skipped for the PR too and we ticket it for later

src/version.rs
GaryPWhite: ah, clever. Learning Rust and appreciate the guidance :)

build.rs
GaryPWhite: changed to a "zero" value

build.rs
GaryPWhite: changed to "empty" value so it's evident if this check failed, without being disruptive to version output.

Licenser (dismissed):
I love it :rocket: great work :D

mfelsche:
@GaryPWhite Gaaaah! there is some git-mess-up at play here, some rebase conflicts.

I would suggest you do something along those lines:

```sh
git fetch upstream                         # whatever name you gave the tremor-runtime repo locally
git checkout main 
git reset --hard upstream/main    # reset main to the state it currently is in our repo (this is potentially dangerous)
git checkout <YOUR_BRANCH>
git rebase main                             # rebase your branch to the freshly reset main branch, so you get a clean state
git push -f origin                           # let github know
```

There are for sure better ways, but this is how I would resolve something like this.

GaryPWhite:
@mfelsche all set when this is done, gtg!

mfelsche (approved):

mfelsche:
@GaryPWhite it seems we merged something in before this. You'd need to rebase your branch again.

PR #981, opened 2021-05-04
Redesign system_metrics test to be less dependent on timings.

mfelsche:
# Pull request

## Description

The system_metrics integration test failed pretty often due to timing issues. Those should be fixed with this PR.

The test has been redesigned to depend less on timings.

## Checklist

* [x] The code is tested

### DIFF
--- a/tremor-cli/tests/integration/system_metrics/assert.yaml
+++ b/tremor-cli/tests/integration/system_metrics/assert.yaml
@@ -8,4 +8,16 @@ asserts:
   - source: metrics_offramp.log
     equals_file: "expected_metrics_offramp.json"
   - source: events.log
-    equals_file: "expected_events.json"
+    contains:
+      - |
+        {"onramp":"metronome","ingest_ns":null,"id":0}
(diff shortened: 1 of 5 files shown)


Licenser:
I'm curious what's the difference, we have contain for all 5 elements before we had a static check for all 5?

mfelsche:
With the current integration testing tools, we can only match on exact contains of the file. They must be equal. Or we can check if it contains a certain line without order. There is not much more we can do. And this test is not testing event order, but that system metrics get produced according to `metrics_interval_s` config. I just wanted to make sure the events that caused the metrics values are present. As the system metrics can be produced with a slight delay, we might have more events in that file, so checking for exact equivalence is what was causing the instability of the test results.

darach (approved):
LGTM

Licenser (approved):
🚀

darach:
We need a stream oriented expect style set of checks here and graceful shutdown / orderly shutdown to resolve - lets consider these for `0.next`

PR #982, opened 2021-05-04
Fix ws offramp not reconnecting after connection loss.

mfelsche:
# Pull request

## Description

When the `ws` downstream server exits, we get an Io error, attempting to close the connection afterwards did lead to a `return Err(...)` and stopped the connection loop trying to reconnect. Also we cancelled the connection task on every `Disconnect` event, although it would have been reconnecting after sending that event, this PR avoids cancelling those tasks. Now we will always attempt reconnects.

## Checklist

* [x] The code is tested


Licenser (approved):
Nice catch!

PR #987, opened 2021-05-05
Fix errer reporting with heredocs caused by preprocessor handling

mfelsche:
# Pull request

## Description

The preprocessing was messing up the locations of the heredoc itself and of everything after it.
Heredoc tokens havent been printed correctly and re-tokenized afterwards, so the location information was off by some lines.

## Checklist

* [x] The code is tested


Licenser (approved):
Nice Catch! And a good improvement to the lexer! :rocket:

tremor-script/src/pos.rs
```diff
@@ -46,6 +46,15 @@ impl std::ops::Sub for Location {
     }
 }
 
+impl std::ops::Add<char> for Location {
+    type Output = Location;
+    fn add(self, c: char) -> Self::Output {
+        let mut l = self; // copy
+        l.shift(c);
+        l
+    }
```
Licenser, line 55: We can avoid this copy since we take ownership

```suggestion
    fn add(mut self, c: char) -> Self::Output {
        self.shift(c);
        self
    }
```

mfelsche: I am not sure this has the desired effect. I explicitly want a new instance and not mutate self. Honestly i am a bit confused by the copy semantics. Would a call like this: 

```rust
my_function(loc1, loc1 + 'a')
```

Make a copy of loc1 for the second argument?

PR #999, opened 2021-05-10
Packaging fixes

mfelsche:
# Pull request

## Description

* Ensure the whole stdlib is put inside the deb package, we were missing `cncf` module files.
* Add `/etc/tremor/config` to the `TREMOR_PATH` so people can put their module files there (maybe in a subdirectory) and they get picked up properly.

## Checklist

* [x] The code is tested


Licenser:
i would rather see this under `/opt/tremor` or `/usr/share/thremor/ then `/etc` POSIX is quite explicit about what `/etc` is for and what it isn't.

mfelsche:
Then lets add `/usr/share/tremor` and `/usr/local/share/tremor`, ok @Licenser ? 

I could also live with `/usr/lib/tremor` and `/usr/local/lib/tremor`. e.g. ruby and python put their std library there (on OSX).

Licenser:
Sounds perfekt! I'm fine with all of those, they go in line with POSIX on different lines of reasoning, which is all fine :) but let's pick one to prevent confusion for people when there are conflicting places where files could be.

Licenser:
I can see the reasoning behind that, I don't like it :joy: but that really isn't a factor :D

Licenser (dismissed):
:rocket:

PR #1001, opened 2021-05-10
Amqp

manuelstein:
# Pull request

## Description

Adding source and sink for AMQP using lapin.

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/139)

## Checklist

* [X] Any user-facing impact of the changes is reflected in docs.tremor.rs
* [X] The code is tested
* [x] Use of unsafe code is reasoned about in a comment (N/A)
* [X] The performance impact of the change is measured (see below)

## Performance

adding an onramp/offramp wouldn't affect workflow processing, would it?

### DIFF
--- a/src/sink/amqp.rs
+++ b/src/sink/amqp.rs
@@ -257,8 +257,12 @@ impl Sink for Amqp {
     }
     async fn terminate(&mut self) {
         if let Some(channel) = self.channel.as_ref() {
-            let _res_close = channel.close(0, "terminating sink");
-            let _res_confirms = channel.wait_for_confirms();
+            if let Err(e) = channel.close(0, "terminating sink").await {
+                error!("[Sink] Failed to close channel: {}", e);
+            }
+            if let Err(e) = channel.wait_for_confirms().await {
+                error!("[Sink] Failed to close channel: {}", e);
+            };
         }
         /*if self.channel.in_flight_count() > 0 {
             // wait a second in order to flush messages.
--- a/src/sink/amqp.rs
+++ b/src/sink/amqp.rs
@@ -105,9 +105,10 @@ impl Amqp {
                 Ok(channel) => self.channel = Some(channel),
                 Err(error) => return Err(error.into()),
             }
-            return Ok(self.channel.as_ref());
+            return Ok(self.channel.as_ref())
+        } else {
+            return Ok(self.channel.as_ref())
         }
-        Ok(None)
     }
 }
 
--- a/src/sink/amqp.rs
+++ b/src/sink/amqp.rs
@@ -23,8 +23,8 @@ use crate::url::TremorUrl;
 use async_channel::{bounded, Receiver};
 use halfbrown::HashMap;
 use lapin::{
-    options::*, publisher_confirm::Confirmation, BasicProperties, Channel, Connection,
-    ConnectionProperties, PromiseChain,
+    options::BasicPublishOptions, publisher_confirm::Confirmation, BasicProperties, Channel,
+    Connection, ConnectionProperties, PromiseChain,
 };
 use serde::Deserialize;
 use std::{fmt, time::Instant};
@@ -192,15 +192,16 @@ impl Sink for Amqp {
                             }
                         }
                         Confirmation::Nack(err) => {
-                            match err {
-                                Some(e) => error!(
+                            if let Some(e) = err {
+                                error!(
                                     "[Sink::{}] failed to send message: {} {}",
                                     &self.sink_url, e.reply_code, e.reply_text
-                                ),
-                                None => error!(
+                                )
+                            } else {
+                                error!(
                                     "[Sink::{}] failed to send message: unknown error",
                                     &self.sink_url
-                                ),
+                                )
                             }
                             if self.error_tx.send(()).await.is_err() {
                                 error!(
--- a/src/sink/amqp.rs
+++ b/src/sink/amqp.rs
@@ -1,0 +1,268 @@
+// Copyright 2020-2021, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![cfg(not(tarpaulin_include))]
+
+//! # AMQP Offramp
+//!
+//! The `amqp` offramp allows producing events to an amqp broker.
+
+use crate::sink::prelude::*;
+use async_channel::{bounded, Receiver};
+use serde::{Deserialize};
+use crate::url::TremorUrl;
+use lapin::{
+    options::*, Connection, ConnectionProperties, Channel, BasicProperties, publisher_confirm::Confirmation
+};
+use halfbrown::HashMap;
+use std::{
+    fmt,
+    time::{Instant},
+};
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct Config {
+    pub amqp_addr: String,
+    routing_key: String,
+    exchange: String,  // "", by default
+    //publish_options: BasicPublishOptions,
+    #[serde(default = "Default::default")]
+    pub close_on_done: bool,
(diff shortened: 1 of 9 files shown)


src/sink/amqp.rs
```diff
@@ -0,0 +1,268 @@
...
+use std::{
+    fmt,
+    time::{Instant},
+};
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct Config {
+    pub amqp_addr: String,
+    routing_key: String,
+    exchange: String,  // "", by default
```
Licenser, line 38: Adding
```rust
    #[serde(default = "Default::default")]`
```
to this would make it optional and default to `""` when nothing is provided

src/sink/amqp.rs
```diff
@@ -0,0 +1,268 @@
...
+                error_tx,
+            }))
+        } else {
+            Err("Amqp offramp requires a config".into())
+        }
+    }
+}
+
+impl Amqp {
+    fn handle_channel(&mut self) -> Result<Option<&Channel>> {
```
Licenser, line 104: this this could in result of `channel` being async also become async

src/sink/amqp.rs
```diff
@@ -0,0 +1,268 @@
...
+                    postprocess(self.postprocessors.as_mut_slice(), ingest_ns, encoded)?;
+                //let headers = meta.get("nats").and_then(|v| v.get_object("headers"));
+                for payload in processed {
+                    /*
+                    // prepare message reply
+                    let message_reply = reply.or(config_reply);
+                    */
+                    // prepare message headers
+                    let properties = BasicProperties::default();
+                    /*
```
Licenser, line 148: Hades would be really nice, but not a blocker, lets throw in a ticket to add them later  if you're OK with that

src/source/amqp.rs
```diff
@@ -0,0 +1,236 @@
...
+pub struct Config {
+    pub amqp_addr: String,
+    queue_name: String,
+    queue_options: QueueDeclareOptions,
+    routing_key: String,
+    exchange: String,  // "", by default
+    #[serde(default = "Default::default")]
+    pub close_on_done: bool,
+    #[serde(default = "Default::default")]
+    pub sleep_on_done: u64,
```
Licenser, line 42: see the nats src, I think the same applies to this config

src/source/amqp.rs
```diff
@@ -0,0 +1,236 @@
...
+        self.with_ack
+    }
+    fn id(&self) -> &TremorUrl {
+        &self.onramp_id
+    }
+    async fn pull_event(&mut self, _id: u64) -> Result<SourceReply> {
+        if self.consumer.is_none() {
+            return Ok(SourceReply::StateChange(SourceState::Disconnected))
+        }
+        if let Some(delivery) = self.consumer.as_mut().unwrap().next().await {
```
Licenser, line 128: we don't allow unwrap or expect, instead we would prefer errors bubbling up in this cases.

The reasoning here is that if a single source or sink has an issue we'd not want to panic the entire application and potentially take down unrelated and pipelines that themselves are unaffected. This gets more important since we allow loading configuration at run time, and someone making an API call could, if we'd allow unwrap/expect/panic/etc. then crash the application with providing a 'bad' configuration

Licenser (changes requested):
Heya,

thousand thanks for this :D it looks really good there are just a few small ponish things left, I mark them inline to make it easier to track :), if you need any help with them feel free to ping on discord.

src/sink/amqp.rs
Licenser: I've no idea what I mean with this so I'll close it :joy:

src/sink/amqp.rs
```diff
@@ -0,0 +1,268 @@
...
+#[derive(Deserialize, Debug, Clone)]
+pub struct Config {
+    pub amqp_addr: String,
+    routing_key: String,
+    exchange: String,  // "", by default
+    publish_options: BasicPublishOptions,
+    #[serde(default = "Default::default")]
+    pub close_on_done: bool,
+    #[serde(default = "Default::default")]
+    pub sleep_on_done: u64,
```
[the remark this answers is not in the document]
manuelstein: thanks, it's some copy'n'paste residual. removed

src/sink/amqp.rs
manuelstein: that's a great idea, I checked that both routing_key and exchange can be blank in amqp, which is the default and in fanout mode even is of no use (so blank string it is).

src/sink/amqp.rs
```diff
@@ -0,0 +1,268 @@
...
+    // headers to use for the messages
+    #[serde(default = "Default::default")]
+    pub headers: HashMap<String, Vec<String>>,
+}
+
+impl Config {
+    fn channel(&self) -> Result<Channel> {
```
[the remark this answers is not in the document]
manuelstein: now, this was a tough one, but I got there eventually - even learning one or two things about Rust. I hope they stick :) thx!

Licenser:
This looks great :D you just got to rebase then it's good to go :rocket: thank you so much for this!

Licenser (dismissed):

src/source/amqp.rs
manuelstein: as above, removed these two residuals

src/sink/amqp.rs
manuelstein: Headers postponed for now due to lack of time, sry.

src/source/amqp.rs
manuelstein: Resolved it into a match tree

manuelstein:
Thank you for the code review!
Rebased onto main.
I'm still getting sanity errors on let _ variable use when the AMQP connection is terminated :-/
Used vscode auto-formatting and addressed as many clippies as I could, not sure what to do with the remaining.
Didn't update Cargo.lock, which seems to fail tests-linux.

manuelstein:
and thanks for the praise 😊❤️ at least it's a start

src/source/amqp.rs
```diff
@@ -0,0 +1,244 @@
...
+    fn id(&self) -> &TremorUrl {
+        &self.onramp_id
+    }
+    async fn pull_event(&mut self, _id: u64) -> Result<SourceReply> {
+        match self.consumer.as_mut() {
+            None => Ok(SourceReply::StateChange(SourceState::Disconnected)),
+            Some(consumer) => {
+                match consumer.next().await {
+                    Some(delivery) => {
+                        let (_, delivery) = delivery.expect("error in consumer");
```
Licenser, line 139: expect's can't be used but you can write it like this:
```
                        let (_, delivery) = delivery?;
```

Licenser:
Heya I want over the remaining issues and made three commits to go over them one by one, let me quickly post them and walk through them :)

[The `Cargo.lock`](https://github.com/tremor-rs/tremor-runtime/commit/e84e494e4831882a49379f6fe6c2b5c25887c9ee) this is easy every time cargo build is called the lock file is updated

[Safety checks](https://github.com/tremor-rs/tremor-runtime/commit/2c1e75319dc6dee57d4aee87bf19f3241e751bd2) there are two things here.

One is that the CI requires must-use results to be used and something is done with them, that's to avoid silent errors. the piny thing doesn't make it easy to figure out what to do but effectively we look at the results. The good news is this caught a missing `await` too so the code would not have been executed :).

The second one is the `expect` could be replaced with a `?` and we have nice errors now  :rocket: 

Last [Clippy](https://github.com/tremor-rs/tremor-runtime/commit/ccc480bdcd5706911fc9e260e79a44d23b5c43a8) our best friend and worst enemy :joy: the two things it fagged were:

A truncation, which makes sense, usually truncating a 128bit value into a 64 bit value can lead to problems. In this case, we know what it is used for so we allow it and add an explanation on my we allow it that way when in 100 years we stumble over that allow we know why it is there :D

The second thing was just the rust style Clippy enforces it prefers `something.map(some_function)` over `something.map(|v| some_function(v))` they do the same but Clippy keeps our code style consistent so it's nice, even so sometimes annoying ;)

Feel free to cherry-pick the commits or copy&paste the changes into your branch, I just wanted to split them out in a few different units so it's easy to explain what is going on behind the scenes, I hope it helps!

Other than that, all that's needed would be a git-rebase to prevent merge conflicts and :rocket: we're good to go :D

Licenser (approved):

Licenser:
Wooooh :D nice

darach:
I love returning from time off work and finding a wonderful new contribution to tremor. Great work @manuelstein and i'm looking forward to giving this a whirl!

darach (approved):
LGTM 🎸

PR #1017, opened 2021-05-17
Fix contraflow cycle via system::metrics pipeline

mfelsche:
# Pull request

## Description

Before this PR is was possible to trigger a contraflow event cycle when connecting any pipeline to the `system::metrics` pipeline. In that case contraflow messages might be sent via the `system::metrics` pipeline which used to account for all its input pipelines (all pipelines are input to the `system::metrics` pipeline). So the cycle went from offramp to pipeline `X` to `system::metrics` pipeline to all of its inputs (among them pipeline `X`) and to `system::metrics` pipeline and so forth.

This is avoided by having `system::metrics` not account for its input pipelines, so it is a dead end for contraflow messages.

## Checklist

* [x] The code is tested

### DIFF
--- a/src/pipeline.rs
+++ b/src/pipeline.rs
@@ -713,6 +713,31 @@ mod tests {
         // non-transactional did not
         assert!(onramp2_rx.is_empty());
 
+        // send a cb insight
+        let event_id = EventId::from((0, 0, 1));
+        addr.send_insight(Event {
+            id: event_id.clone(),
+            cb: CbAction::Close,
+            ..Event::default()
+        })
+        .await?;
+
+        // transactional onramp received it
+        // chose exceptionally big timeout, which we will only hit in the bad case that stuff isnt working, normally this should return fine
+        match timeout(Duration::from_millis(10000), onramp_rx.recv()).await {
+            Ok(Ok(onramp::Msg::Cb(CbAction::Close, cb_id))) => assert_eq!(cb_id, event_id),
+            Err(_) => assert!(false, "No msg received."),
+            m => assert!(false, "received unexpected msg: {:?}", m),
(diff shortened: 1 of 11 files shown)


src/pipeline.rs
```diff
@@ -709,6 +713,31 @@ mod tests {
...
+        addr.send_insight(Event {
+            id: event_id.clone(),
+            cb: CbAction::Close,
+            ..Event::default()
+        })
+        .await?;
+
+        // transactional onramp received it
+        // chose exceptionally big timeout, which we will only hit in the bad case that stuff isnt working, normally this should return fine
+        match timeout(Duration::from_millis(10000), onramp_rx.recv()).await {
```
Licenser, line 727: would it make sense to const that timeout?

tremor-cli/src/test/assert.rs
```diff
@@ -209,6 +210,7 @@ pub(crate) fn process(
     Ok((s, elements))
 }
 
+#[allow(clippy::clippy::too_many_lines)]
```
Licenser, line 213: perhaps we can split out the different parts into their own functions?

mfelsche: done

src/pipeline.rs
mfelsche: dont know, but i did it :)

Licenser: :+1: you always say that so I wanted to say it once too :joy:

Licenser (approved):
:rocket: great job :D

src/pipeline.rs
mfelsche: that is only fair 😆

Issue #1015, opened 2021-05-17
Wildly unsound use of `transmute` across whole project


Voultapher:
**Problem**

Borderline every time `transmute` is used in this project it's unsound and wrong! Especially in the context of self-referential structs.

I've been going through rental dependents and replacing them with [self_cell](https://github.com/Voultapher/self_cell) eg. https://github.com/matklad/fall/pull/6#issuecomment-841620597 as it's no longer maintained and users are encouraged to switch to another solution.

I started by trying to replace the use for `LineValue`, immediately I noticed the alarming use of `transmute`.

> tremor-script/src/lib.rs:204
```rust
/// Borrow the parts (event and metadata) from a rental.
/// This borrows the data as immutable and then transmutes it
/// to be mutable.
#[allow(mutable_transmutes, clippy::transmute_ptr_to_ptr)]
#[must_use]
pub fn parts<'value, 'borrow>(
    &'borrow self,
) -> (&'borrow mut Value<'value>, &'borrow mut Value<'value>)
where
    'borrow: 'value,
{
    unsafe {
        let data = self.suffix();
        let unwind_event: &'borrow mut Value<'value> = std::mem::transmute(data.value());
        let event_meta: &'borrow mut Value<'value> = std::mem::transmute(data.meta());
        (unwind_event, event_meta)
    }
}
```

Here as you say yourself, you take an immutable self and transmute it to be mutable. Also note this function isn't even marked as *unsafe*. It's called 27 times across the whole project in all kinds of places.

This is *not* the only time you do this. A further example from the same file:

> tremor-script/src/lib.rs:204
```rust
/// Event value forced to borrowd mutable (uses `mem::transmute`)
///
/// # Safety
/// This isn't save, use with care and reason about mutability!
#[allow(mutable_transmutes, clippy::transmute_ptr_to_ptr, clippy::mut_from_ref)]
#[must_use]
pub unsafe fn force_value_mut(&self) -> &mut Value<'event> {
    std::mem::transmute(&self.v)
}
```

Here you mark this function unsafe and remark `This isn't save, use with care and reason about mutability!`.

Let's quickly take a look at what the language specification has to say about this:

https://doc.rust-lang.org/nomicon/transmutes.html

> Transmuting an & to &mut is UB.
>
> - Transmuting an & to &mut is *always* UB.
> - No you can't do it.
> - No you're not special.

So no matter how careful you are when calling these functions, it will always instantly be wrong!

[...]

PR #1019, opened 2021-05-18
untransmutisation

Licenser:
## Description

#1015 brings up the frequent use of `std::mem::transmute`. It is worth doing an inventory and see in which places rust and we have advanced enough to remove some of them. This PR intentionally avoids the realm of `rental` / `self_cell` and focuses on non-destructive restructuring.

## Related

* Related Issues: #1015

## Checklist

* [x] The code is tested

## Performance

As expected, with all changes being just re-structuring there is no measurable performance difference beyond run-on-run variance.

new: ~353 MB/s
main: ~354 MB/s

### DIFF
--- a/.github/checks/safety.sh
+++ b/.github/checks/safety.sh
@@ -15,6 +15,7 @@ code sanity checker
   -e         check for expect
   -d         check for dbg!
   -t         check for todo!
+  -m         check for transmute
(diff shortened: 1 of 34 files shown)


darach (approved):
LGTM 👍

PR #1020, opened 2021-05-18
Fix another case of ws offramp not reconnecting.

mfelsche:
# Pull request

## Description

When using websocat as server and terminating with `Ctrl-C` we trigger a connection termination without the required close frame being exchanged. This was triggering and error in the connection loop which left the connection unusable and did lead to stalls on new requests. With this PR the task handling connection will only terminate when there is a serious error in async channel comminucation, hinting at the runtime having a bigger problem, so terminating is ok.

No CHANGELOG entry as there is already one for another case, which got fixed.

Tested by using the workshop example where the error occurred (not ideal, i know).

## Checklist

* [x] The code is tested

### DIFF
--- a/src/sink/ws.rs
+++ b/src/sink/ws.rs
@@ -17,10 +17,11 @@
 use crate::sink::prelude::*;
 use crate::source::prelude::*;
 use async_channel::{bounded, unbounded, Receiver, Sender};
-use async_tungstenite::async_std::connect_async;
+use async_std::net::TcpStream;
 use async_tungstenite::tungstenite::error::Error as WsError;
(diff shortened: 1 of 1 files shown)


src/sink/ws.rs
```diff
@@ -320,8 +324,23 @@ async fn ws_connection_loop(
                                 correlation.as_ref(),
                             )
                             .await?;
-                            // close connection explicitly
-                            ws_stream.close(None).await?;
+                            if !matches!(
```
Licenser, line 327: could up extract the matches into a function so we only need to update it in one place?

mfelsche: done

Licenser (approved):
Wooohhh! Thank you! 🚀

Issue #1031, opened 2021-05-19
generic consume to remove the need for a rental


Licenser:
**Describe the problem you are trying to solve**

For combining two borrowed values we have a function that takes two rentals, deconstructs one, combines the two `owned` parts, and executes a merge function on the borrowed parts.

This is not elegant but sensible, sadly `rental` does not give us an out-of-the-box facility to achieve this.

**Describe the solution you'd like**

Ideally, a case like this could be generally handled by a `rental` replacement.

**Notes**
Followup to #1015

Licenser:
@Voultapher would it make sense / be possible to have a safe combine function in `self_cell`.

I'm thinking a trait for the `owed` part that indicates that it's combinable, could be pre-implemented for some types like linked lists where the values when concatenating won't move in memory and then have a merge function along the lines that `combine` takes.

Licenser:
I don't think that would work, let me elaborate a bit on what happens in this case.

The `owners` a `Vec<u8>` that represents a memory region that contains the content of some `&str` (or rather `Cow<'_, str>` that are part somewhere inside the `dependant`. For context, this is the result of in-situ parsing (more often than not of JSON :sob:) were to avoid allocations strings remain in the source region of memory. This `owner` is never changed, or mutated.

To combine two of those self-referential structs we add a layer of indirection to the `onwer` so it becomes a `Vec<Vec<u8>>` 

We use a vector here to protect the inner `Vec<u8>` or more precisely the points inside of them from relocation. Adding to the outer vector might relocate the outer `Vec` but the pointers of the inner vector remain in place.

For consuming what we do is add the `owner` of the consumed struct (`Vec<Vec<u8>>`) to the `owner` data of the consuming struct. Then in the merge function, both `dependants` can be merged as their `owners` are combined in the consumed referential struct.

The suggestion of using LineValue as owned data won't really solve the problem as the result could not be emitted as a new event in its new form.

Performance would be a second issue, batches can span thousands or tens of thousands of events so the cost here would get out of hand quickly :(, we could probably defer doing the iteration step towards the very end which would negate the cost but change the latency characteristic (i.e. small cost on every event would become large cost on emitting) which isn't great for a streaming workload either.

Voultapher:
Thanks for clarifying, so the result of this merge must be some Event again. During my refactoring I considered 2 solutions should this be the case:

A) Have Event be an Enum of a simple Event and MergedEvent
[...]

PR #1042, opened 2021-05-20
use Pin in rental to be more explicit

Licenser:

# Pull request

## Description

No code changes beside using `Pin` in our event rentals. This is to explain the rust compiler that "those things can't move in memory" to make the constraints we reasoned about logically more explicit now that rust can represent "This won't move!" in it's type system

## Related

This is in relation to #1031 

## Checklist

* [x] The code is tested

## Performance

No code Changes aside of a newtype wrapper.


mfelsche (approved):

Voultapher:
I'm not sure this change is sound, in `consume` you transmute it into `struct ScrewRental { [...] Vec<Vec<u8>>}` the `Pin` is gone. And in deed `self_unrent.raw.append(&mut other_unrent.raw);` could move the inner `Vec` during relocation.

Licenser:
Ja that was fixed in #1044 :)

And the `Vec` itself being moved is fine, it's just a container. The pointer in the Vec, that isn't moved and that's what Pin protects and that's what needs protection. The description in the docs isn't that great, but if you look at the [4th paragraph](https://doc.rust-lang.org/std/pin/index.html) under that light it makes sense.

PR #1049, opened 2021-05-21
Remove two transmutes, fixes #1025

### LINKED ISSUE #1025: Restructure Arc Based transmute in graph greation

During graph creation, we use Arcs along with transmuting to pass down shared data and create rentals on the content.

While this is safe and correct it opens the potential of subtle bugs if arcs and source data ever change.

It would be good to re-write those instances, a possible approach would be to move the extraction into the rental new.

**Notes**

Followup to #1015

Licenser:
# Pull request

## Description

This restructures the way we create the instantiated, we use an ARC for the source to reference from the borrowed entity. We transmuted this instead of re-looking the at node up when creating the self-referential struct. This is safe but depends on reasoning rather than automated checks which carry the risk that refactoring that the constraints get violated or overlooked.

This PR moves the borrowing inside of the rental to avoid transmutation.

## Related

* Related Issues: fixes #1025

## Checklist

* [x] The code is tested

## Performance

This does not touch any hot path code.

### DIFF
--- a/tremor-pipeline/src/op/trickle/select.rs
+++ b/tremor-pipeline/src/op/trickle/select.rs
@@ -35,7 +35,7 @@ use tremor_script::{
     query::StmtRental,
     query::StmtRentalWrapper,
     utils::sorted_serialize,
-    Value,
+    QueryRental, Value,
 };
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


mfelsche (approved):
LGTM, hopefully the gods of clippy, coverage and rustc have mercy on you!

darach (approved):
LGTM 🚀 Nice...

mfelsche (approved):

PR #1055, opened 2021-05-22
Add support for tls

### LINKED ISSUE #12: (r)syslog onramp / preprocessor
Enable Tremor to receive and send Syslog Protocol Messages (https://tools.ietf.org/html/rfc5424) , supporting as much syslog implementations as possible that might deviate from the standard.

In the wild we have different syslog protocols being used, the standard IETF format and the old BSD format. So ideally we should support both.

### via UDP

Syslog messages are usually sent via UDP where 1 UDP packet contains 1 syslog message. We already support receiving data via UDP with our [UDP onramp](https://docs.tremor.rs/artefacts/onramps/#udp). We need a way to turn the packet data we receive into a structured `Value`. For this, we have [codecs](https://docs.tremor.rs/artefacts/codecs/). A syslog codec should be able to handle both syslog message formats, or we write two different codecs, one for each format.

### via TCP/TLS

The story for supporting syslog over TLS/TCP is a bit more involved. We currently do not support TLS over our [TCP onramp](https://docs.tremor.rs/artefacts/onramps/#tcp), so this needs to be added. This is a major milestone towards full syslog support.

Given we have TLS, in order to support syslog messages over TCP, we need to support the RFC 5425 transport, that contains a textual length prefix before each message. This could be handled with a [Preprocessor](https://docs.tremor.rs/artefacts/preprocessors/) similar to the [length-prefixed preprocessor](https://docs.tremor.rs/artefacts/preprocessors/).

### via UDP
For sending syslog messages, we need to turn structured data in an Event (`Value`) into the the `syslog` protocol format.

### via TCP/TLS

For sending messages over TCP/TLS, we also need to add the textual length-prefix used in RFC 5425. Tremor already supports sending data via UDP and TCP via offramps. 

The [TCP offramp](https://docs.tremor.rs/artefacts/offramps/#tcp) needs to get TLS support for supporting sending syslog messages via TCP/TLS.

### Phase 1 - receive syslog via UDP

- [x] receive syslog data via UDP (onramp) and turn syslog messages into structured events
- [x] turn structured events into textual syslog messages and send them out via UDP (offramp)

### Phase 2 - syslog via TCP/TLS

- [x] add support for the RFC 5425 transport protocol (textual length prefix)

### Phase 3 - TLS support for TCP

- [x] add support for receiving TLS encrypted data via TCP onramp
- [ ] add support for sending TLS encrypted data via TCP offramp

## Reference

* IETF Syslog Message Format [RFC 5424](https://tools.ietf.org/html/rfc5424)
* BSD Syslog Message Format [RFC 3164](https://tools.ietf.org/html/rfc3164)
* Syslog Transport via UDP [RFC 5426](https://tools.ietf.org/html/rfc5426)
* Syslog Transport via TLS/TCP [RFC 5425](https://tools.ietf.org/html/rfc5425#section-4.3)

murex971:
Signed-off-by: murex971 <nupur202000@gmail.com>

# Pull request

## Description

add support for receiving TLS encrypted data via TCP onramp

## Related

* Related Issues: fixes #12 (Phase 3)

## Checklist

* [x] The code is tested

### DIFF
--- a/src/source/tcp.rs
+++ b/src/source/tcp.rs
@@ -13,9 +13,17 @@
 // limitations under the License.
 #![cfg(not(tarpaulin_include))]
 
+use crate::errors::{Error, ErrorKind, Result};
 use crate::source::prelude::*;
+use async_channel::Sender;
 use async_channel::TryRecvError;
 use async_std::net::TcpListener;
+use async_tls::TlsAcceptor;
+use rustls::internal::pemfile::{certs, pkcs8_private_keys, rsa_private_keys};
+use rustls::{Certificate, NoClientAuth, PrivateKey, ServerConfig};
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
 
 // TODO expose this as config (would have to change buffer to be vector?)
 const BUFFER_SIZE_BYTES: usize = 8192;
@@ -24,6 +32,13 @@ const BUFFER_SIZE_BYTES: usize = 8192;
 pub struct Config {
     pub port: u16,
     pub host: String,
+    pub tls: Option<TLSConfig>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct TLSConfig {
+    cert: PathBuf,
+    key: PathBuf,
 }
 
 impl ConfigImpl for Config {}
@@ -95,9 +110,16 @@ impl Source for Int {
         let (tx, rx) = bounded(crate::QSIZE);
         let uid = self.uid;
         let path = vec![self.config.port.to_string()];
+
+        let server_config: Option<ServerConfig> = if let Some(tls_config) = self.config.tls.as_ref()
+        {
+            Some(load_server_config(tls_config)?)
+        } else {
+            None
+        };
         task::spawn(async move {
             let mut stream_id = 0;
-            while let Ok((mut stream, peer)) = listener.accept().await {
+            while let Ok((stream, peer)) = listener.accept().await {
                 let tx = tx.clone();
                 stream_id += 1;
                 let origin_uri = EventOriginUri {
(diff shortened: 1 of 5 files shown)


src/source/tcp.rs
```diff
@@ -108,36 +124,30 @@ impl Source for Int {
                     // TODO also add token_num here?
                     path: path.clone(), // captures server port
                 };
+                let tls_acceptor: Option<TlsAcceptor> = conf
+                    .tls
+                    .as_ref()
+                    .and_then(|tls_config| load_server_config(tls_config).ok())
```
Licenser, line 130: we should fail when the certificates can't be loaded not silently switch to non SSL mode. That could be quite confusing for operators and be quite a security issue.

src/source/tcp.rs
```diff
@@ -108,36 +124,30 @@ impl Source for Int {
...
-                            break;
-                        };
-                    }
+                    if let Some(acceptor) = tls_acceptor {
+                        match acceptor.accept(stream).await {
+                            Ok(tls_stream) => {
+                                read_loop(tls_stream, tx, stream_id, origin_uri).await
+                            }
+                            Err(e) => {
+                                error!("TLS Error: {}", e);
```
Licenser, line 145: Since stream start was send above we would need to send a stream end here

Licenser (changes requested):
Other than the cert loading and a missing stream end this looks great :D! Some nice refactoring of the read_loop and I quite like how good the errors are for the certs!

mfelsche (approved):

src/source/tcp.rs
```diff
@@ -108,36 +130,31 @@ impl Source for Int {
...
-                            error!("TCP Error: {}", e);
-                            break;
-                        };
-                    }
+                    if let Some(acceptor) = tls_acceptor {
+                        match acceptor.accept(stream).await {
+                            Ok(tls_stream) => {
+                                read_loop(tls_stream, tx, stream_id, origin_uri).await
+                            }
+                            Err(e) => {
```
mfelsche, line 148: Dang, this `e` is not used.
```suggestion
                            Err(_e) => {
```

Licenser (approved):
woooh nice!

mfelsche (approved):

PR #1057, opened 2021-05-27
TCP/TLS offramp support

### LINKED ISSUE #12: (r)syslog onramp / preprocessor

murex971:
# Pull request

## Description

Add support for sending TLS encrypted data via TCP offramp.

## Related

* Related Issues: fixes #12 

## Checklist

* [x] The code is tested

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ beef = {version = "0.5", features = ["impl_serde"]}
 byteorder = "1"
 bytes = "1.0"
 chrono = "0.4"
+either = { version = "1.6", features = ["serde"]}
 elastic = "0.21.0-pre.5"
 error-chain = "0.12"
 futures = "0.3.15"
@@ -78,7 +79,9 @@ tremor-value = {path = "tremor-value"}
 url = "2.2"
 value-trait = "0.2"
 rustls = "0.19"
-async-tls = "0.11.0"
+# needs to be compatible with what rusttls is using
+webpki-roots = "0.21"
+async-tls = { version = "0.11", features = ["client", "server"]}
 
 mapr = "0.8"
 tempfile = {version = "3.2"}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,9 +79,7 @@ tremor-value = {path = "tremor-value"}
 url = "2.2"
 value-trait = "0.2"
 rustls = "0.19"
-# needs to be compatible with what rusttls is using
-webpki-roots = "0.21"
-async-tls = { version = "0.11", features = ["client", "server"]}
+async-tls = "0.11"
 
 mapr = "0.8"
 tempfile = {version = "3.2"}
--- a/src/sink/tcp.rs
+++ b/src/sink/tcp.rs
@@ -31,14 +31,13 @@ use halfbrown::HashMap;
 use async_tls::TlsConnector;
 use rustls::ClientConfig;
 
-use std::io::Cursor;
-use std::path::{PathBuf};
-use std::sync::Arc;
-use either::Either;
 use async_std::io::Write;
+use either::Either;
+use std::io::Cursor;
+use std::path::PathBuf;
+use std::sync::Arc;
 
 type Stream = Box<dyn Write + std::marker::Unpin + Send>;
-
 
 /// An offramp streams over TCP/IP
 pub struct Tcp {
@@ -56,14 +55,14 @@ pub struct Config {
     pub ttl: u32,
     #[serde(default = "default_no_delay")]
     pub is_no_delay: bool,
-    #[serde(with = "either::serde_untagged_optional")]
+    #[serde(with = "either::serde_untagged_optional", default = "Default::default")]
     pub tls: Option<Either<TLSConfig, bool>>,
 }
 
 #[derive(Deserialize, Debug, Default)]
 pub struct TLSConfig {
     cafile: Option<PathBuf>,
-    domain: Option<String>
+    domain: Option<String>,
 }
 
 fn default_no_delay() -> bool {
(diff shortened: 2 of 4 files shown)


src/sink/tcp.rs
```diff
@@ -28,9 +28,22 @@ use crate::sink::prelude::*;
...
+use either::Either;
+use async_std::io::Write;
+
+type Stream = Box<dyn Write + std::marker::Unpin + Send>;
+
+
 /// An offramp streams over TCP/IP
 pub struct Tcp {
-    stream: Option<TcpStream>,
+    // TODO: check if we can/should use an enum here where we implement async_std::io::Write manually?
```
mfelsche, line 45: @Licenser we had a really hard time coming up with a solution here. We need to handle both variants of streams here (`TcpStream` or `TlsStream<TcpStream>`) as if it was one, and need to store it inside this struct, without being able to use generics, as this is a dynamic thing. We resulted to a `Box<dyn SomeTrait>`. But i thought maybe you have an idea how to do it better.

src/sink/tcp.rs
```diff
@@ -39,17 +52,25 @@ pub struct Tcp {
 pub struct Config {
     pub host: String,
     pub port: u16,
-    #[serde(default = "ttl")]
+    #[serde(default = "default_ttl")]
     pub ttl: u32,
-    #[serde(default = "t")]
+    #[serde(default = "default_no_delay")]
     pub is_no_delay: bool,
+    #[serde(with = "either::serde_untagged_optional")]
```
mfelsche, line 59: In order to allow this to be optional, i think you need to do:

```suggestion
    #[serde(with = "either::serde_untagged_optional", default = "Default::default")]
```

src/sink/tcp.rs
```diff
@@ -85,8 +106,31 @@ impl Tcp {
         }
         Ok(())
     }
+
+    async fn connect(config: &Config) -> Result<Stream> {
+        let stream = TcpStream::connect((config.host.as_str(), config.port)).await?;
+        stream.set_ttl(config.ttl)?;
+        stream.set_nodelay(config.is_no_delay)?;
+        let s: Stream = match config.tls.as_ref() {
```
mfelsche, line 114: could you add a debug log to each of those branches? That would be helpful when debugging some issues

src/sink/tcp.rs
Licenser: The dyn looks fine, the perf impact by dyn probably isn't noticeable given it's network IO

Cargo.toml
```diff
@@ -78,7 +79,7 @@ tremor-value = {path = "tremor-value"}
 url = "2.2"
 value-trait = "0.2"
 rustls = "0.19"
-async-tls = "0.11.0"
+async-tls = { version = "0.11", features = ["client", "server"]}
```
darach, line 82: Do we need to specify the features as these should be defaulted? Its worth checking

Licenser (dismissed):
:rocket:

Cargo.toml
mfelsche:
```suggestion
async-tls = "0.11"
```

Licenser (approved):
Nice work! :)

mfelsche:
Yes!!!

Issue #348, opened 2020-07-09
Support cross building and packaging for ARM targets


anupdhml:
**Describe the problem you are trying to solve**

Add cross build and packaging support for targets `armv7-unknown-linux-gnueabihf` and `aarch64-unknown-linux-gnu` so that ARM device users can try tremor without having to compile the whole project.

https://github.com/wayfair-tremor/tremor-runtime/tree/master/packaging#supported-targets
https://github.com/wayfair-tremor/tremor-runtime/tree/master/packaging/builder-images

**Describe the solution you'd like**

Extend the implementation done for x86_64 linux targets as part of https://github.com/wayfair-tremor/tremor-runtime/pull/274.

Packaging formats to support: `archive (tar.gz), deb, rpm`

**Notes**

We know that (some) arm builds work for tremor, via https://github.com/wayfair-tremor/tremor-runtime/pull/347.

Licenser:
being able to build on arm is slowly (or quickly) creeping back into a possible state. @hkratz has made some great contributions to simd-json (our main blocker) to make it arm workable again we're mostly waiting on a rust-nightly update for the latest stdlib update and then we should be able to give this a shot.

It will only build on nightly for now but that's probably good enough :tm: for experimentation and testing :D

Licenser:
blocked on https://github.com/cuviper/autocfg/issues/36

Licenser:
The issue seems to be a code generation issue in rust https://github.com/rust-lang/rust/issues/85962 

good news:

it might be CPU related so it might only affect some users

bad news:

a codegen issue is kind of serious :(

Licenser:
Update after @Hoverbear tested the ARM issue, it seems based on the CPU involved so it might not be a general problem :rocket:

PR #1064, opened 2021-06-03
Add drone ci file for ARM testing

Licenser:
# Pull request

## Description

Add drone.io CI for testing arm builds

## Related

* Related Issues:  #348

## Checklist

* [x] The code is tested

## Performance

No code changes

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,16 @@
+kind: pipeline
+name: test-on-arm64
+
+platform:
+  arch: arm64
+
+steps:
+  - name: test
+    image: rust:1
+    environment:
+      RUSTFLAGS: "-C target-cpu=native"
+    commands:
+      - sudo apt-get -qy update && sudo apt-get install -y libssl-dev libssl1.1
+      - rustup default nightly
+      - rustup update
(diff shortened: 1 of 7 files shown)


mfelsche:
Do you think it is worth only executing those, if the repo is `tremor-rs/tremor-runtime` to exclude execution (and possibly fails) on forks? https://docs.drone.io/pipeline/triggers/#by-repository

Other than that and the deps LGTM

Licenser:
PR / Push usually means "this code" vs "this code merged to main" which both makes sense I think )plus they're running in parallel :D

Runs won't trigger on forks that don't have a PR, but I think we will want them to run on forks that open a PR (that works I do it for simd-json already). 

But that all said I'd not mark the ARM test as a mandatory pass until we decide to make ARM a 1st level target and more importantly until we don't depend on nightly. My experience with nightly is that it often just stops working intermittently and I don't think we would want that to block the CI, until then we can judge failures "manually" and keep the code working on ARM on a case by case basis.

Licenser:
We really don't want this on the blocking path for PR's :joy: so far we're 50min in with running it.

mfelsche (approved):

Issue #721, opened 2021-01-28
Extend EQC testing for tremor-script


Licenser:
**Describe the problem you are trying to solve**

Tremor Script is growing in capabilities and complexity. We started with using an EQC based specification on the language to cover the fundamentals. Tremor Script since then has gained a number of new features that are not cover, we want to increase the coverage of the specification and tests.

**Describe the solution you'd like**

We want to cover a large number of language features as part of the EQC testing framework. The ideal outcome would be full coverage of all features tremor script offers but it can be approached incrementally and covering a feature at a time.]

This also includes improving the interfacing between Erlang and Rust and testing integration.

An optimal outcome would be to also include the results of these tests in the code coverage.

Milestones:

* [x] Base expression language
* [x] string interpolation
* [ ] Match syntax
* [ ] Extractors
* [ ] comprehensions
* [ ] Function library
* [ ] Functions
* [ ] Fault testing

**Notes**

The current set of tests and the model can be found here:

https://github.com/tremor-rs/tremor-runtime/tree/main/tremor-script

Additional Material with ideas, helpful context or useful information:

* [a rust erlang interface](https://github.com/rusterlium/rustler)
* [EQC documentation](http://quviq.com/documentation/eqc/)
* [Some EQC getting started guide](https://gist.github.com/efcasado/3df8f3f1e33eaa488019)
* [Slindes from an EQC workshop](https://erlang.org/workshop/2008/Sess31.pdf)
* [Another quickcheck workshop](https://www.erlang-factory.com/upload/presentations/193/EUC2009-QuickCheck.pdf)
* [A chapter on EQC with elixir](https://livebook.manning.com/book/the-little-elixir-and-otp-guidebook/chapter-11/)
* [Software Testing with QuickCheck](https://www.researchgate.net/publication/225219256_Software_Testing_with_QuickCheck)
* [Testing C Code with EQC](https://blog.project-fifo.net/testing-c-code-erlang-quickcheck/)
* [Testing firmware with EQC](https://medium.com/@potsdamnhacker/quickcheck-and-the-deep-firmware-278f77cd2342)
* [Model-based testing of data types with side effects](https://dl.acm.org/doi/abs/10.1145/2034654.2034662?download=true)

diru1100:
Hey @Licenser 
Thank you for looping me in
Excited to work on this ^_^

PR #1065, opened 2021-06-03
Add SSE source.

### LINKED ISSUE #885: SSE Source

SSE (server sent events) are a still common way of streaming data from a web server to a client. Currently, we do not support this as a source. It would be nice to.

Either as a standalone source or an addition to the rest source, we should be able to subscribe to long-lived SSE HTTP connections.

dak-x:
# Pull request

## Description
Adds an onramp for Server-Sent-Events.  

## Related

* Related Issues: closes #885

## Checklist

* [x] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### New features
 
 - Add tests for patch feature in tremor-script [#721](https://github.com/tremor-rs/tremor-runtime/issues/721)
+- Add onramp for SSE (Server Sent Events)
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 ### New features
 
 - Add tests for patch feature in tremor-script [#721](https://github.com/tremor-rs/tremor-runtime/issues/721)
-- Add onramp for SSE (Server Sent Events)
+- Add onramp for SSE (Server Sent Events) [#885](https://github.com/tremor-rs/tremor-runtime/issues/885)
 
 ### Fixes
 
--- a/src/source/sse.rs
+++ b/src/source/sse.rs
@@ -167,7 +167,7 @@ impl Source for Int {
             uid: self.uid,
             scheme: "tremor-sse".to_string(),
             // What even is the host here?
-            host: "localhost".to_string(),
+            host: hostname(),
             port: None,
             path: vec![String::default()],
         };
--- a/src/source/sse.rs
+++ b/src/source/sse.rs
@@ -137,24 +137,24 @@ impl Source for Int {
     }
 
     async fn init(&mut self) -> Result<SourceState> {
-        // Neccessary check as Isahc(Curl) client panics if the url isn't complete. Seems like an edge case.
-        // Maybe use the H1-client as the backend. Though Heinz might have a better solution which is impl in rest sink.
-        // How shall I output this error.
-        let url = self.config.url.as_str().parse();
-        let _url: http::Uri = match url {
-            Ok(url) => url,
-            Err(err) => return Err(err.to_string().into()),
-        };
-
         info!(
             "[Source::{}] subscribing to {}",
             self.onramp_id.to_string(),
             &self.config.url
         );
 
+        // Neccessary check as Isahc(Curl) client panics if the url isn't complete. Seems like an edge case.
+        // Though Heinz might have a better solution which is Endpoint struct in rest sink.
+        let url_check = self.config.url.parse::<http::Uri>();
+        if let Err(err) = url_check {
+            return Err(err.to_string().into());
+        }
+
         let (tx, rx) = bounded(crate::QSIZE);
+
         let url: surf_sse::Url = self.config.url.parse()?;
         let headers = self.config.headers.clone();
+
         // The client runs with default configuration from crate
         task::spawn(handle_init(url, headers, tx));
         self.event_source = Some(rx);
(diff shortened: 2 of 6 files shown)


src/source/sse.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+async fn handle_init(
+    url: surf_sse::Url,
+    tx: Sender<surf_sse::Event>,
+) -> Result<()>{
+    // Run the event source here.
+    let mut event_source = surf_sse::EventSource::new(url);
+
+    // Potential Problem on calling next. There is try_next available we need future_util crate for that.
```
Licenser, line 89: This is fine to block here since we have an indirection from the channel where you handle the case of "we're waiting a long time for a reply"

Licenser (commented):
Nice :D this looks really good so far! Thank you :)

dak-x:
A couple questions:
1. What client config settings should be exposed in the config.
2. What should be the Origin_URI exactly.

I will try to add codecs (first study them)
Also currently the dep: surf_sse is linked to my git repo. Once goto-bus-stop/surf-sse#13 merges I will fix that.

src/source/sse.rs
darach: Yep - we can update the comment ( based on Heinz's text for example ) and its all good!

Licenser:
>     What client config settings should be exposed in the config.

Something that mirrors the [rest sink](https://github.com/tremor-rs/tremor-runtime/blob/main/src/sink/rest.rs#L229) would be nice, the option to set headers and client mostly, other than that I can't think of much needed. I mention the sink here not the source since it mirrors an SSE source better as that too is a client - so mirroring the configuration would be good, that way users that know one can quickly learn the other.

>     What should be the Origin_URI exactly.

The `origin_uri` usually encodes all the information we have on the source of the event. So host/path where the SSE events are read from would be nice and helpful in there along with the fine :)

> I will try to add codecs (first study them)

The good news is, for the most part, you don't need to worry about that, the way tremor is built those are decoupled concerns, a source is only responsible for handling the incoming data. Where it becomes a concern is when the SSE server adds content-encoding, that's where the codec_overwrite comes into play, you can take a look at either the REST  source or sink on how it's handled there.

dak-x:
@Licenser you may take the fork from my profile and publish it on your credentials so I may update the dependencies. My PR seems to be going no where. `https://github.com/dak-x/surf-sse`. Also I do not see of a way to get the internal port in which my client works to include in OriginUrl. I also request you to review the hack-ish way to validate a URL in my code. (http::Url is the type used by the isahc ( curl ) client ). The URL validation by Surf `Panics!`  on a few edge cases, because of an `unwrap` on https://docs.rs/http-client/6.4.1/src/http_client/isahc.rs.html#60 which seems really odd. This crate is used as a dep by `surf` for various clients.

Licenser:
Since this is in the top level not a crate we publish you could just link the git repository as a source that prevents having a fork and a competing package as that is a bit of a larger task (we will need to maintain that in the long run too if we publish it, we don't want to produce abondanware). So git dependency?

src/source/sse.rs
```diff
@@ -0,0 +1,196 @@
...
+impl Source for Int {
+    fn id(&self) -> &TremorUrl {
+        &self.onramp_id
+    }
+
+    async fn init(&mut self) -> Result<SourceState> {
+        // Neccessary check as Isahc(Curl) client panics if the url isn't complete. Seems like an edge case.
+        // Maybe use the H1-client as the backend. Though Heinz might have a better solution which is impl in rest sink.
+        // How shall I output this error.
+        let url = self.config.url.as_str().parse();
```
Licenser, line 143: Shouldn;t this check already be handled by Line 156?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 ### New features
 
 - Add tests for patch feature in tremor-script [#721](https://github.com/tremor-rs/tremor-runtime/issues/721)
+- Add onramp for SSE (Server Sent Events)
```
Licenser, line 18: Worth linking `[#885](https://github.com/tremor-rs/tremor-runtime/issues/721)` here

Licenser (changes requested):
This looks great :D just some tiny polishing then I think we're good to go :). Great work!

src/source/sse.rs
dak-x: It should but there seems to an edge case from my random inputs. In the crate `surf` `Surf::Url` is passed to the underlying http client. But in the implementation of a traits which abstract over the http clients (https://docs.rs/http-client/6.4.1/src/http_client/isahc.rs.html#60 ) an unwrap() panics. So here I verify it beforehand to avoid a thread panic.****

Licenser: Ohhh, now it makes sense, sorry I didn't gasp the comment fully :+1: makes sense!

src/source/sse.rs
```diff
@@ -0,0 +1,196 @@
...
+        let _url: http::Uri = match url {
+            Ok(url) => url,
+            Err(err) => return Err(err.to_string().into()),
+        };
```
Licenser, line 147: Our safety checks forbid `let _...` as it circumvents the "must use" rule for functions and can lead to unhandled errors. So this could be rewritten as:
```rust
        let url: http::Uri = self.config.url.as_str().parse();
        if let Some(err) = url.err() {
           return Err(err.to_string().into()),
        };
```

the upside being is that this also makes the intent a bit clearer (aka we don't care about the result just if it's an error)

dak-x: I have made the changes. This is a nice pattern. Will keep this in mind.

Licenser (approved):
Woooh :rocket: thank you :D this is great! Now tremor can SEE when a HTTP server sends events :joy:

mfelsche:
Great stuff @dak-x ! Love it. One tiny follow up request: Could you create a tag for your repository in the current state and point our `Cargo.toml` file to that tag? Otherwise our builds depend on the state of that `default` branch which might change frequently.

PR #1066, opened 2021-06-08
Lifetimes2.0

Licenser:
# Pull request

## Description

This PR reduces the use of the `'script` lifetime in the interpreter, this simplifies the logic a bit and allows some cleanup along the path.

There should not be any functional changes here.

## Checklist

* [x] The code is tested

## Performance

no difference outside of run to run varriance


Licenser:
I'm marking this as ready for review, it does include a significant number of cleanups and improvements and is worth stabilizing as it is. We can continue chomping down on issues in a separate PR

mfelsche (approved):
Yay, `to_static`!

PR #1067, opened 2021-06-08
Pub/Sub modification for message ordering

jigyasak05:
Signed-off-by: Jigyasa <jigyasakhaneja05@gmail.com>

# Pull request

## Description

Adding support for creating subscription with the option to enable/disable message ordering

## Related

https://github.com/tremor-rs/tremor-runtime/issues/724

## Checklist

* [x] The code is tested

### DIFF
--- a/src/connectors/gcp/pubsub.rs
+++ b/src/connectors/gcp/pubsub.rs
@@ -19,7 +19,7 @@ use googapis::google::pubsub::v1::{
     publisher_client::PublisherClient, subscriber_client::SubscriberClient,
 };
 use googapis::google::pubsub::v1::{
-    AcknowledgeRequest, PublishRequest, PubsubMessage, PullRequest, PullResponse,
+    AcknowledgeRequest, PublishRequest, PubsubMessage, PullRequest, PullResponse, Subscription,
 };
 use std::collections::HashMap;
 use tonic::transport::Channel;
@@ -35,7 +35,7 @@ pub(crate) async fn send_message(
         attributes: HashMap::new(),
         message_id: "".into(),
         publish_time: None,
-        ordering_key: "".to_string(),
+        ordering_key: "order".to_string(),
     };
 
     let response = client
@@ -89,4 +89,34 @@ pub(crate) async fn acknowledge(
         })
         .await?;
     Ok(())
+}
+
+pub(crate) async fn create_subscription(
+    client: &mut SubscriberClient<Channel>,
+    project_id: &str,
+    topic_name: &str,
+    subscription_name: &str,
+    enable_message_ordering: bool,
+) -> Result<Subscription> {
+    let response = client
+        .create_subscription(Subscription {
+            name: format!(
+                "projects/{}/subscriptions/{}",
+                project_id, subscription_name
+            ),
+            topic: format!("projects/{}/topics/{}", project_id, topic_name),
+            push_config: None,
+            ack_deadline_seconds: 0,
+            retain_acked_messages: true,
+            message_retention_duration: None,
+            labels: HashMap::new(),
+            enable_message_ordering,
+            expiration_policy: None,
+            filter: "".into(),
+            dead_letter_policy: None,
+            retry_policy: None,
+            detached: false,
+        })
+        .await?;
+    Ok(response.into_inner())
 }
--- a/src/connectors/gcp/pubsub.rs
+++ b/src/connectors/gcp/pubsub.rs
@@ -29,13 +29,14 @@ pub(crate) async fn send_message(
     project_id: &str,
     topic_name: &str,
     data_val: &[u8],
+    ordering_key: &str,
 ) -> Result<String> {
     let message = PubsubMessage {
         data: data_val.to_vec(),
         attributes: HashMap::new(),
-        message_id: "".into(),
+        message_id: "".into(), // ID of this message, assigned by the server when the message is published.
         publish_time: None,
-        ordering_key: "order".to_string(),
+        ordering_key: ordering_key.to_string(),
     };
 
     let response = client
--- a/src/sink/gpub.rs
+++ b/src/sink/gpub.rs
@@ -27,6 +27,7 @@ use crate::connectors::qos::{self, QoSFacilities, SinkQoS};
 use crate::sink::prelude::*;
 use futures::executor::block_on;
 use googapis::google::pubsub::v1::publisher_client::PublisherClient;
+use googapis::google::pubsub::v1::subscriber_client::SubscriberClient;
 use halfbrown::HashMap;
 use tonic::transport::Channel;
 use tremor_pipeline::{EventIdGenerator, OpMeta};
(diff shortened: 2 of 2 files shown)


src/connectors/gcp/pubsub.rs
```diff
@@ -35,7 +35,7 @@ pub(crate) async fn send_message(
         attributes: HashMap::new(),
         message_id: "".into(),
         publish_time: None,
-        ordering_key: "".to_string(),
+        ordering_key: "order".to_string(),
```
Licenser, line 38: Should we really hardcode this? Won't that enforce the user into a specific schema?

src/connectors/gcp/pubsub.rs
```diff
@@ -35,7 +35,7 @@ pub(crate) async fn send_message(
         attributes: HashMap::new(),
         message_id: "".into(),
```
Licenser, line 36: should this be read from the metadata so users could specify their message id's?

src/sink/gpub.rs
```diff
@@ -74,11 +78,13 @@ impl offramp::Impl for GoogleCloudPubSub {
                 .as_ref()
                 .ok_or("Offramp Google Cloud Pubsub (pub) requires a config")?,
         )?;
-        let remote = Some(block_on(pubsub_auth::setup_publisher_client())?);
+        let remote_publisher = Some(block_on(pubsub_auth::setup_publisher_client())?);
```
Licenser, line 81: block_on will block the entire thread, we really will want to avid that.

src/sink/gpub.rs
```diff
@@ -129,74 +141,117 @@ impl Sink for GoogleCloudPubSub {
...
+                        &pubsub::send_message(remote_publisher, &project_id, &topic_name, &msg)
+                            .await?,
+                    ));
+                }
+
+                PubSubCommand::CreateSubscription(
+                    project_id,
+                    topic_name,
+                    subscription_name,
+                    enable_message_ordering,
```
Licenser, line 192: this would probably be better as an boolean instead of a string, or an enum if we there are more then two values

src/sink/gpub.rs
```diff
@@ -49,6 +51,7 @@ pub struct Config {}
 
 enum PubSubCommand {
     SendMessage(String, String, Value<'static>),
```
Licenser, line 53: named structs would be better here, it'd allow conveying information about what the different parts mean. something like:

```rust
    SendMessage{this_gets_a_descripteive_name: String, that_gets_a_descriptive_anem: String, data: Value<'static>},
``

some for the create command

File: src/connectors/gcp/pubsub.rs (hunk 1 above)
Comment (jigyasak05) [reply src/connectors/gcp/pubsub.rs:38]: According to the docs, `ordering_key` -  ```If non-empty, identifies related messages for which publish order should be respected. If a Subscription has enable_message_ordering set to true, messages published with the same non-empty ordering_key value will be delivered to subscribers in the order in which they are received by the Pub/Sub system. All PubsubMessages published in a given PublishRequest must specify the same ordering_key value.```

So I decided to keep it a non-empty string for the case when we want to enable message ordering such that all the messages we send will have the same ordering_key that's been hardcoded.

File: src/connectors/gcp/pubsub.rs (hunk 2 above)
Comment (jigyasak05) [reply src/connectors/gcp/pubsub.rs:36]: Since here we are sending the message, so the message-id will be assigned by the server. When it's published, we display the message-id as a part of the metadata.
According to the docs for `message_id`: 
ID of this message, assigned by the server when the message is published. Guaranteed to be unique within the topic. This value may be read by a subscriber that receives a PubsubMessage via a Pull call or a push delivery. **It must not be populated by the publisher in a Publish call**.

File: src/sink/gpub.rs (hunk 3 above)
Comment (jigyasak05) [reply src/sink/gpub.rs:81]: I'm not sure what to replace it with :(

File: src/sink/gpub.rs (hunk 4 above)
Comment (jigyasak05) [reply src/sink/gpub.rs:192]: Yes I think these fields are a part of an enum 🤔

File: src/sink/gpub.rs (hunk 5 above)
Comment (jigyasak05) [reply src/sink/gpub.rs:53]: Oh nice ^_^ I didn't know we could do that too

File: src/connectors/gcp/pubsub.rs [hunk 6]
```diff
@@ -90,3 +90,33 @@ pub(crate) async fn acknowledge(
...
+    let response = client
+        .create_subscription(Subscription {
+            name: format!(
+                "projects/{}/subscriptions/{}",
+                project_id, subscription_name
+            ),
+            topic: format!("projects/{}/topics/{}", project_id, topic_name),
+            push_config: None,
+            ack_deadline_seconds: 0,
+            retain_acked_messages: true,
```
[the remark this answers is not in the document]
jigyasak05: Yes! We can add support for that if it's required

jigyasak05:
I'm not sure why this CI test is failing ;-; How do I fix it?

src/connectors/gcp/pubsub.rs
Licenser: Ohh nice :D that's an interesting API choice of google. Perhaps we add a comment so that in half a year we have not forgotten why it is empty?

src/sink/gpub.rs
Licenser: We probably would want to move the logic to creat a connection into `init` it's the general pattern we use for most offramps. `from_config` creates a skeleton that isn't functional yet, `Sink::init` initializes that skeleton. That way we have control over when we initiate connections. `Sink::init` for that reason is also `async` so you can just use `.await` there :) - you can take a look at TCP, REST for an example of how it is done in those

src/sink/gpub.rs
Licenser: Ja you define it here, you could change that along with named fields :)

https://github.com/tremor-rs/tremor-runtime/pull/1067/files/86b3ef2c2bef2fa7cefbb8726b8c7dea2c945051#diff-222dacbb009cbcb0efa06bc565c40b364ec2cd1be034a913d41fdb1672388c75R54

src/sink/gpub.rs
Licenser: Ja named fields are super powerful, a bit more verbose to write but that pays off 3 weeks in the future when no one remembers which field meant what :D

src/connectors/gcp/pubsub.rs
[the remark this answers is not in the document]
Licenser: Not urgent but perhaps good to keep it in a ticket or comment so we remember :)

src/connectors/gcp/pubsub.rs
Licenser: Ja it would probably be nice to let the user decide if they need ordering and for which messages, perhaps a `$gpubsub.ordering_key` ?

jigyasak05: Okay yes sounds good

src/sink/gpub.rs
jigyasak05: Yes! Fixed them all in the updated PR

Licenser (approved):
This looks great! Thank you so much, you rock!! :rocket:

src/sink/gpub.rs
Licenser: wooooh!

darach:
Woop woop!

Issue #587, opened 2020-11-13
Tremor Aggregation Function "stats::sum(int|float)" produces runtime error


abhie19:
**Problem**
It has been observed that when we are using the stats::sum(<event.value>), restarting tremor produces a runtime error as follows:
```
2020-11-13T20:20:52.823615192+00:00 ERROR tremor_runtime::pipeline - error: Error(Script(BadAccessInEvent(Range(Location { unit_id: 0, line: 77, column: 22, absolute: 1605 }, Location { unit_id: 0, line: 77, column: 37, absolute: 1620 }), Range(Location { unit_id: 0, line: 77, column: 34, absolute: 1617 }, Location { unit_id: 0, line: 77, column: 37, absolute: 1620 }), "sum", ["min", "mean", "percentiles", "count", "max"])), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })
```

Below is the full trickle script to reproduce the bug- 
```
#!config metrics_interval_s = 10

use std::type;

use std::record;

define script add_kafka_key

script
    use std::json;
    let $kafka_key = event.measurement+","+json::encode(event.tags);
    emit;
end;

create script add_kafka_key;

define tumbling window ten_secs
with
    interval = core::datetime::with_seconds(10),
script
    event.timestamp
end;

create stream normalize;

create stream aggregate;

define generic::batch operator batch

with

count = 3000,

timeout = 5

end;

create operator batch;

select {
"measurement": event.measurement,
"tags": event.tags,
"field": group[2],
"value": event.fields[group[2]],
"timestamp": event.timestamp,
}

from in
group by set(event.measurement, event.tags, each(record::keys(event.fields)))
into aggregate
having type::is_number(event.value);

select
{
"measurement": event.measurement,
"tags": event.tags,
"stats": aggr::stats::dds(event.value, [ "0.9", "0.95", "0.99" ]),
"field": event.field,
"timestamp": aggr::win::last(event.timestamp), # we can't use min since it's a float
}
from aggregate[`ten_secs`]
group by set(event.measurement, event.tags, event.field)
into normalize;

select {
"measurement": event.measurement,
"tags": event.tags,
"fields": {
"count_{event.field}": event.stats.count,
"min_{event.field}": event.stats.min,
"max_{event.field}": event.stats.max,
"mean_{event.field}": event.stats.mean,
"sum_{event.field}": event.stats.sum,
"p90_{event.field}": event.stats.percentiles["0.9"],
[...]

PR #1077, opened 2021-06-15
Connectors Implementation 

mfelsche:
# Pull request

## Description

Connectors are successors to onramps/offramps, sinks/sources. A connector can be a source of events and events can be sent to a connector. Not all connectors necessarily support all both the sink and source nature/trait.

This PR introduces a new config artefact
and tries to remodel the current onramp/offramp handling to move as much boilerplate and common logic from the connector implementation to the runtime (or where not possible to reasonable utilities), so that writing new connectors is getting easier, while maintaining all invariants enforced by the runtime.

Additional connector features:

* Pause / Resume
* Connect / Reconnect / connectivity probes
* Configurable Codecs/Processors

### TODOs

- [x] Create predefined reconnect strategies (e.g. randomized exponential backoff, ...)
- [x] correlation metadata in RPC style connectors (moved to connectors)
- [x] broadcast system metrics to all metrics connector instances
- [x] look into stdin connector if we want broadcasting to all instances as well
- [x] integration tests for verifying proper startup
- [x] integration tests for verifying pause/resume behaviour
- [x] rust in-process integration test suite for connectors (in `tests` folder)
  - getting a stream for events from `out`, `err`
  - getting a sender for sending events to `in`
  - verifying events send to and received from a connector
  - e.g. sending TCP data to it, asserting on received events
- [x] only register debug connector types (exit, debug, cb, benchmark) when a certain server flag is given (`--register-debug-connectors` or similar)
- [x] Binding Quiescence - assuming no overlaps or shared resources
- [x] Connectors API - status, change status (pause, resume, ...)
- [x] Flow API - add GET status, PUT status (pause, resume, ...) endpoints  
- [x] Ensure that metrics are emitted correctly to the metrics connector, same for pipelines
- [ ] Connector ports (tier 1; known to be in active use, will be prioritised for porting)
  - [x] discord (h)
  - [x] DNS (h)
    - [x] correlation
  - [x] elastic (confirmed) (m)
    - [x] correlation
  - [x] exit (m) 
  - [x] file (please not line-based this time, both sink and src) (m)
  - [x] kafka-consumer (m)
  - [x] kafka-producer (m)
  - [x] kv (h)
    - [x] correlation
  - [x] metronome (h)
  - [x] otel (src + sink, confirmed)
    - [ ] add echo client/server integration test 
  - [x] http (rest server + client, should this include SSE?,  confirmed) (d)
    - [x] correlation
    - [x] basic auth
    - [x] concurrency=1 in client issue with blocking
    - [ ] batch punctuation for better/correct batch event response handling
  - [x] stdio (stdin + stdout + stderr)
  - [x] tcp-client (m)
  - [x] tcp-server (m)
  - [x] udp-client (h)
  - [x] udp-server (h)
  - [x] wal (h)
  - [x] unix-socket (confirmed)
- [ ] Connector ports (tier 2; utility, testing & debugging, this will not be prioritized on porting form sink/src to connector )
  - [x] benchmark connector (blaster/blackhole previously) (h)
  - [x] cb (m)
  - [ ] debug
  - [ ] sse (could this be part of http?)
  - [ ] gpubsub (gsub + gpub, use cases known and expanding )
  - [ ] gcs (confirmed not used at the moment but future usecase exists)
- [ ] Connector ports (tier 3; not known to be in use - low priority, but comment this issue if you are using any of these, dear readers )
  - [x] websocket-client (near term use cases exist) (d)
  - [x] websocket-server (near term use cases exist) (d)
  - [ ] postgres (src + sink, do we still need this?, not known to be in production use at this time )
  - [ ] nats (src + sink, not known to be in production use at this time )
  - [ ] newrelic ( community contributed by grover, not known if it is in production use at this time )
  - [x] crononome (ported, not in production use that we can confirm)
  - [ ] amqp (confirmed not used)

## Test expectations and conditions

Tier 1 and Tier 2 connectors should have container tests in `/tests` contributed with the new connector test
harness ( this helps greatly with code coverage ) and integration tests in `/tremor-cli/tests/integration` that
demonstrate canonical and simple use and validation of the connector in all its configuration modes ( client or server, secure and insecure etc.. ). For Tier 3, container and integration tests are preferable.  

## Related

* [RFC](https://github.com/tremor-rs/tremor-rfcs/pull/32)

## Checklist

* [ ] The code is tested


src/config.rs
```diff
@@ -114,6 +118,101 @@ pub struct OffRamp {
...
+    ///
+    /// e.g.:
+    ///       codec_map:
+    ///         "application/json": "json"
+    ///         "text/plain": "string"
+    ///
+    /// A default builtin codec mapping is defined
+    /// for msgpack, json, yaml and plaintext codecs with the common mime-types
+    #[serde(default = "Default::default", skip_serializing_if = "Option::is_none")]
+    pub(crate) codec_map: Option<halfbrown::HashMap<String, String>>,
```
Licenser, line 204:
```suggestion
    pub(crate) codec_map: Option<halfbrown::HashMap<String, Either<String, CodecConfig>>>,
```

src/connectors/reconnect.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// and spawn a retry task
+    fn update_and_retry(&mut self, sender: Sender<Msg>, url: TremorUrl) -> Result<()> {
+        // update internal state
+        if let Some(max_retries) = self.config.max_retry {
+            if self.attempt >= max_retries {
+                return Err(ErrorKind::MaxRetriesExceeded(max_retries).into());
+            }
+        }
+        self.attempt += 1;
+        self.interval_ms = (self.interval_ms as f64 * self.config.growth_rate) as u64;
```
Licenser, line 87: Not sure if it's overkill or not, but would it make sense to trait out the next interval computation so we can have different strategies?

src/connectors/sink.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// Result for a sink function that may provide insights or response.
+///
+/// It can return None or Some(vec![]) if no insights/response were generated.
+///
+/// An insight is a contraflowevent containing control information for the runtime like
+/// circuit breaker events, guaranteed delivery events, etc.
+///
+/// A response is an event generated from the sink delivery.
+pub(crate) type ResultVec = Result<Option<Vec<SinkReply>>>;
```
Licenser, line 37: I know this came to originated somewhere else, but I wonder would it make sense to simplify this to `Result<Vec<SinkReply>>` or is there a difference between an empty Vec and a None?

src/connectors/reconnect.rs
mfelsche: indeed this makes sense

src/connectors/sink.rs
mfelsche: good point

src/connectors/source.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    pub(crate) fn sender(&self) -> Sender<crate::source::SourceReply> {
+        self.tx.clone()
+    }
+}
+
+#[async_trait::async_trait()]
+impl Source for ChannelSource {
+    async fn pull_data(&mut self, _id: u64) -> Result<crate::source::SourceReply> {
+        if self.rx.is_empty() {
```
Licenser, line 73: https://docs.rs/async-channel/1.6.1/async_channel/struct.Receiver.html#method.try_recv might be more applicable here?

mfelsche: i had it here, i though, maybe checking a number on the queue might be faster 😎

src/errors.rs
```diff
@@ -206,15 +212,6 @@ error_chain! {
         }
 
         // TODO: Old errors, verify if needed
```
Licenser, line 214:
```suggestion

```

tremor-common/src/ids.rs
```diff
@@ -15,131 +15,63 @@
 // common id handling
 
 /// we namespace onramp, offramp and operator ids differently in order to avoid clashes
-const ONRAMP_ID_BASE: u64 = 0b0;
-const OPERATOR_ID_BASE: u64 = 0b1000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_u64;
-const OFFRAMP_ID_BASE: u64 = 0b1000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_u64;
+const ONRAMP_ID_BASE: u64 =    0b0;
+const OPERATOR_ID_BASE: u64 =  0b1000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_u64;
+const OFFRAMP_ID_BASE: u64 =   0b1000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_u64;
+const CONNECTOR_ID_BASE: u64 = 0b1000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_u64;
```
Licenser, line 21: looking at those, in hindsight, would it make sense to use the first 4 bit as type identifier and the rest as counter that way the values would become masks instead of base ID's (not that it'd change much bit it kind of feels cleaner looking at it now)
```suggestion
const ONRAMP_ID_BASE: u64 =    0x10_00_00_00_00_00_00_00_u64;
const OPERATOR_ID_BASE: u64 =  0x20_00_00_00_00_00_00_00_u64;
const OFFRAMP_ID_BASE: u64 =   0x30_00_00_00_00_00_00_00_u64;
const CONNECTOR_ID_BASE: u64 = 0x40_00_00_00_00_00_00_00_u64

File: src/version.rs [hunk 7]
```diff
@@ -45,22 +43,11 @@ pub fn long_ver() -> String {
...
     eprintln!("tremor version: {}", long_ver().as_str());
     eprintln!("tremor instance: {}", instance!());
-    let (version_n, version_s) = get_rdkafka_version();
-    eprintln!("rd_kafka version: 0x{:08x}, {}", version_n, version_s);
 }
 
 /// Logs tremor and librdkafka version.
 pub fn log() {
     info!("tremor version: {}", long_ver().as_str());
-    let (version_n, version_s) = get_rdkafka_version();
```
Licenser, line 55: do we really want to get rid of the kafka version, that has been  super helpful printout

mfelsche: Good catch, that is an artifact of my attempts to move connectors to their own module, will re-add

src/lib.rs
```diff
@@ -201,52 +161,53 @@ pub async fn load_query_file(world: &World, file_name: &str) -> Result<usize> {
 /// # Errors
 /// Fails if the file can not be loaded
 pub async fn load_cfg_file(world: &World, file_name: &str) -> Result<usize> {
-    info!("Loading configuration from {}", file_name);
+    info!(
+        "Loading configuration from {}",
+        std::path::Path::new(file_name).canonicalize()?.display()
```
Licenser, line 166: We should use `tremor_common::file::canonicalize` here instead

Licenser: and since we use `file_name` below perhaps save a version of `canonical_file` and re-use that in all places we display filename?

Licenser:
A question to the std_stream connector, I noticed that the user needs to specify what type of std_stream it is which in my mind breaks the connector concept a bit since instead of a bidirectional connection it is now a unidirectional connection that the user specifies the direction to.

Would it make sense to turn the std_stream into a connector that allows reading/writing to? behaving more like `stdio`?

This would require connectors to be aware when their input (sink) and output (source) ports are connected to dynamically shut down / start the corresponding subsystems but it would feel a bit more intuitive than having to write:

```yaml
connector:
  - id: stdin
    type: std_stream
    config:
      stream: stdin
  - id: stdout
    type: std_stream
    config:
      stream: stdout
  - id: stderr
    type: std_stream
    config:
      stream: stderr
```

darach:
Once this branch goes green it would be worth merging the `troy` branch once both have rebased against main. Progressing the deployment language work now depends on having connectors available in the runtime.

src/connectors/metrics.rs
```diff
@@ -51,7 +51,11 @@ pub(crate) struct MetricsChannel {
 
 impl MetricsChannel {
     pub(crate) fn new(qsize: usize) -> Self {
-        let (tx, rx) = bounded(qsize);
+        let (mut tx, rx) = broadcast(qsize);
+        // We user overflow so that non collected messages can be removed
+        // FIXME: is this what we want? for Metrics it should be good enough
```
mfelsche, line 56: What would be ideal here would be that older messages get discarded, not newer ones. If we have this measure in place, and if sending will thus never wait for the channel being empty we can remove the hack with the keep_alive flag for sources

Licenser: Investigating the async-broadcast crate, this is exactly what happens :D I misread it the first time:

https://docs.rs/async-broadcast/0.3.3/async_broadcast/struct.Receiver.html#method.set_overflow

mfelsche: I will take care of removing the superfuous `keep_alive` stuff that i introduced for the metrics connector then. :)

Licenser: :rocket:

src/connectors/source.rs
```diff
@@ -0,0 +1,1052 @@
...
+/// with the generic information available in the connector
+/// the builder then in a second step takes the source specific information to assemble and spawn the actual `SourceManager`.
+pub fn builder(
+    connector_uid: u64,
+    config: &ConnectorConfig,
+    connector_default_codec: &str,
+    qsize: usize,
+    source_metrics_reporter: SourceReporter,
+) -> Result<SourceManagerBuilder> {
+    let preprocessor_names = config.preprocessors.clone().unwrap_or_else(Vec::new);
```
marioortizmanero, line 360: `unwrap_or_default` might be neater here :P

marioortizmanero: Also why not turn this function into `SourceManagerBuilder::new`?

mfelsche: It is my past as a java developer which is haunting my code to this day.

Licenser (approved):
🚀 also fixes #587

Licenser:
The rebase seems to have broken something, the tests don't compile :(

Licenser (approved):

darach (approved):
LGTM - Goodbye YAML, Hello `troy`!

Licenser (approved):

Licenser (dismissed):

PR #1082, opened 2021-06-17
display response - gpub

jigyasak05:
Signed-off-by: Jigyasa <jigyasakhaneja05@gmail.com>

# Pull request

## Description

Display the response received from GCP when a subscription is created.

## Related

https://github.com/tremor-rs/tremor-runtime/issues/724

## Checklist

* [x] The code is tested


jigyasak05:
The commented fields were not included in the response because the data type was not compatible.

src/sink/gpub.rs
```diff
@@ -220,15 +220,38 @@ impl Sink for GoogleCloudPubSub {
...
+                    let retain_acked_messages = res.retain_acked_messages;
+                    let enable_message_ordering = res.enable_message_ordering;
+                    // let labels = res.labels;
+                    let message_retention_duration = res.message_retention_duration.ok_or("Error getting `message retention duration` for the created subscription")?.seconds;
+                    // let expiration_policy = res.expiration_policy.ok_or("Error getting `expiration policy` for the created subscription")?;
+                    let filter = res.filter;
+                    // let dead_letter_policy = res.dead_letter_policy.ok_or("Error getting `dead letter policy` for the created subscription")?;
+                    // let retry_policy = res.retry_policy.ok_or("Error getting `retry policy` for the created subscription")?;
+                    let detached = res.detached;
+                    response.push(literal!({
```
mfelsche, line 245: It seems that the response that is sent back below will always be an array of records, even if the event is not batched (has only one value). Is that intended? The docs PR https://github.com/tremor-rs/tremor-www-docs/pull/160 describes it as an object. Maybe below you check if the `response` vec has 1 element and the event is not batched, then you can extract the first element and use this as response Value, otherwise use the full vec.

mfelsche (commented):
Looking good, just one single question about the response format :)

src/sink/gpub.rs
jigyasak05: Yes true, In case of gpub (sink), since we can only send one message at a time currently - we'll always have one value  in the response🤔 So I just extract the first element instead of sending the `vec` as response, right?

darach: This could just be something we can clarify in the docs? Batching in tremor is very different from how the pubsub APIs work and the responses from G pubsub are a vector of messages - instead of `object` perhaps we should have stated `value` instead - and explained. a little how the underlying messages are mapped to values in tremor?

Licenser: I think the confusion stems from that this seems to be returning an array with a struct
```js
[{
  "subscription": "projects/<project-id>/subscriptions/<subscription-name>",
  "topic": "projects/<project-id>/topics/<topic-name>",
  "ack_deadline_seconds": `<ack_deadline_seconds>`,
  "retain_acked_messages": `<retain_acked_messages>`,
  "enable_message_ordering": `<enable_message_ordering>`,
  "message_retention_duration": `<message_retention_duration>`,
  "filter": "<filter>",
  "detached": `<detached>`
}]
```
while the documentation states it returns a struct:
```js
{
  "subscription": "projects/<project-id>/subscriptions/<subscription-name>",
  "topic": "projects/<project-id>/topics/<topic-name>",
  "ack_deadline_seconds": `<ack_deadline_seconds>`,
  "retain_acked_messages": `<retain_acked_messages>`,
  "enable_message_ordering": `<enable_message_ordering>`,
  "message_retention_duration": `<message_retention_duration>`,
  "filter": "<filter>",
  "detached": `<detached>`
}
```

I would opt for changing the reply to match the documentation not the documentation to match the reply as at least from the docs there seems to be no reason to return an array with a singular element it would only making consuming the data harder for users.

jigyasak05: I get this response but I only mentioned the struct part because the GCP response consists of only that:
```
{
  "data": [
    {
      "subscription_name": "projects/<project-id>/subscriptions/tremor-test",
      "topic": "projects/<project-id>/topics/tremor",
      "ack_deadline_seconds": 10,
      "retain_acked_messages": true,
      "enable_message_ordering": true,
      "message_retention_duration": 604800,
      "filter": "",
      "detached": false
    }
  ],
  "meta": {}
}

Licenser: you get that because you combine data and meta in your pipeline, the event is only the data part :) so it'd still make sense not to wrap the struct in an array I think, unless we expect to get more than one reply back which I don't think is the case?

jigyasak05: Yes true! So I'll update it to not return it in the form of an array but just the `value`

Licenser: That would be wonderful :)! 🚀

mfelsche: To add some context here:

In Tremor it is possible that a single event actually contains more than one. This can be the case when the [batch operator](https://docs.tremor.rs/tremor-query/operators/#genericbatch) was used. In that case the event `value` contains multiple nested events over which we can iterate with `Event::value_iter()` (as done above) or `Event::value_meta_iter()`. This is kind of a special case that makes our life as offramp developers harder :/

So, if the event arriving at the gpub offramp is batched you get multiple entries back from `Event::value_iter` and you will do multiple calls to the gpub api, one for each `value`. The `response` vec will contain multiple values. In that case mark the reply event also as batched and take care that the actual `response` value is an array of records (which is the case right now) and that it has the same format as as is used in the batch operator: https://github.com/tremor-rs/tremor-runtime/blob/main/tremor-pipeline/src/op/generic/batch.rs#L97-L104

If it is not batched (the incoming event has `event.is_batch = false`, you could extract the first element from the `response` vec and use that as the event data, not an array. I think this is the most consistent handling here.

mfelsche: Sorry for stirring up the discussion again, i didnt see the above discussion when i was writing my lament. :)

jigyasak05: Alright!! I'll fix it ^_^

jigyasak05: Updated it

jigyasak05:
Why does this build fail? This test was passing earlier :-(

Licenser:
DroneCi is very temperamental, nothing you did wrong, nothing we did wrong either, sadly it's just how drone works. But we have it as optional so it's safe to ignore it :)

jigyasak05:
Okay cool! I had added the comment ^_^

Licenser (approved):
🚀 nicely done!

Issue #1090, opened 2021-06-25
Defaulting in Expressions


Licenser:
**Describe the problem you are trying to solve**

Defaulting, or the option to ensure that certain fields are present and if not give them a default value.

Today this can be solved with a construct like:

```tremor
merge default with event;
```

where default is the struct holding the default values. However, while this may be elegant it is not something that's easy to understand (nor is it fast, but that's a secondary concern).

This also touches on a somewhat related concern: looking up into structs that do not exist.

```
a.b.c + d.e.f
```

where `b` for example does not exist as a key in `a`.

**Describe the solution you'd like**

# filling a struct with default values

There are a number of possible approaches however they all have their advantages and problems.

### default with struct
a new patch keyword `default` could be used to specify a default structure for a record.

```
patch event of
  default {"key": "value"}
end
```

this would effectively be the same as `merge default of event end;` just in a more verbose but easier to understand syntax.

### default with k/v

a default keyword for each value could be used

```
patch event of
  default "key1" => "value1",
  default "key2" => "value2",
  # ...
end
```

This would be very clear on what we set but quite verbose, it could be combined with  `default {...}`

## a default operator

This is fairly simple, provide a struct as default values and have them fill the event as it passes through.

## a default infix operator

```
# (could clash with default for path)
let event = event default {"key": "value"};
# or (could cause gramer conflicts)
let event = event default merge {"key": "value"};
# or (could cause gramer conflicts)
let event = event merge {"key": "value"};
# or (uses up new keyword)
let event = event ensure {"key": "value"};
# or (uses up new keyword)
let event = event put_if_absent {"key": "value"};
```
The second example here is since we'll discuss `default` for path lookups later and the second option might get around issues but create new ones.

# default for lookups

sometimes we want to lookup a value in a nested structure and the path doesn't exist.

## return `null`

This is an option but it comes with serious issues, it would equate `null` with `absent` that makes it very hard to distinguish if a value is set to `null` or `absent` moving the problem from "What do we do when the value isn't set?" to "is the value set?"

## default as a keyword

```
a.b.c default 7
# or (would burn up a new keyword, also somewhat not as clear ensure might mean we ensure it's always 7)
a.b.c ensure 7
``` 

would allow being explicit to when we lookup and replace a default value. 

## an option type

yes please, but this is a lot of work, would extend the type system, introduce error handling and struct functions so there isn't even a proposed syntax for this

## a function

```
struct::get(struct, key, default)
```

[...]

PR #1091, opened 2021-06-25
Add patch default and clean up patch

Licenser:
# Pull request

## Description

Sketch of event defaulting as part of #1090. `default` as part of patch is less invasive and problematic then defaulting for the whole expression language since:

1) it is contained to the `patch` operation
2) it has no averse side effects to other possible implementations
3) no new keywords are introduced
4) it mirrors the already existing `merge` operation so is familiar in concept

This PR:

* Add `default => {...}` to patch (mirroring `merge => {...}`)
* Add `default "key" => "value"` (mirroring `merge "key" => {...}` but non recursive)
* Update patch to enforce string keys as part of the gramer instead of creating runtime errors
* Fixes errors for "expected string" to not show a tick instead of a quote
* Moves type errors on patch keys to compile time
* Simplifes the record creation

## Related

* Related Issues: #1090
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/168)

## Checklist

* [x] The code is tested

## Performance

A slight performance improvement of ~1% using the `patch` benchmark


tremor-script/src/ast/raw.rs
```diff
@@ -1286,52 +1260,64 @@ pub enum PatchOperationRaw<'script> {
     /// we're forced to make this pub because of lalrpop
     Insert {
         /// we're forced to make this pub because of lalrpop
-        ident: ImutExprRaw<'script>,
+        ident: StringLitRaw<'script>,
```
mfelsche, line 1263: Why do we need to restrict this to literals?
I could imagine to use patch to manipulate a key dynamically by some application logic. For such use cases we'd need an `ImutExprInt` here.

Licenser: we can still do that using `"#{dynamic_key}"` restricting it to string literals ensurews that we don't have runtime errors for keys not being strings - we can catch this at compiletime. It also prevents confusion when someone writes:

```
patch event of
  insert timestamp => 0
end;
```

which can be easiely be read (and written) as `timestamp` being they key, not `timestamp` being the variable that holds the key.

mfelsche (dismissed):
Looking good.

src/codec/json.rs
```diff
@@ -12,35 +12,60 @@
...
+use std::{cmp::max, marker::PhantomData};
 
 use super::prelude::*;
+use tremor_script::utils::sorted_serialize;
 use tremor_value::AlignedBuf;
 
-pub struct Json {
+/// Sorting for JSON
+pub trait Sorting: Sync + Send + Copy + Clone + 'static {
+    const SORTED: bool;
```
darach, line 23: Disgusting, I love it, very nice indeed! 🚀

Licenser: :D ja and efficient (see https://rust.godbolt.org/z/3nTsd1s1K)

mfelsche (approved):

PR #1101, opened 2021-07-01
Add zstd pre-and-postprocessor.

### LINKED ISSUE #1100: Add `zstd` pre- and postprocessors

Lemire has been publishing a small measurement about zstd vs. gzip for json compression https://lemire.me/blog/2021/06/30/compressing-json-gzip-vs-zstd/ . And zstd is super fast and offers better compression. But we dont have zstd yet.

We want what lemire has: `zstd`.
We want `zstd` pre- and postprocessors similar to what we have for `zlib` etc.

dak-x:
Signed-off-by: Daksh Chauhan <dak-x@outlook.com>

# Pull request
Add `zstd` pre-and-postprocessors

## Related

* Related Issues: closes #1100 

## Checklist

* [x] The code is tested

### DIFF
--- a/src/preprocessor.rs
+++ b/src/preprocessor.rs
@@ -69,6 +69,7 @@ pub fn lookup(name: &str) -> Result<Box<dyn Preprocessor>> {
         "ingest-ns" => Ok(Box::new(ExtractIngresTs {})),
         "length-prefixed" => Ok(Box::new(LengthPrefix::default())),
         "textual-length-prefix" => Ok(Box::new(TextualLength::default())),
+        "zstd" => Ok(Box::new(Zstd::default())),
         _ => Err(format!("Preprocessor '{}' not found.", name).into()),
     }
 }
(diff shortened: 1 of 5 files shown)


dak-x:
The code required seems too straightforward. I feel I am missing some stuff. Also I should probably add some tests, but I do not know what tests.

src/preprocessor.rs
```diff
@@ -69,6 +69,7 @@ pub fn lookup(name: &str) -> Result<Box<dyn Preprocessor>> {
         "ingest-ns" => Ok(Box::new(ExtractIngresTs {})),
         "length-prefixed" => Ok(Box::new(LengthPrefix::default())),
         "textual-length-prefix" => Ok(Box::new(TextualLength::default())),
+        "zstd" => Ok(Box::new(Zstd::default())),
```
Licenser, line 72: We should include this in the `decompress` preprocessor for autodetecting zstd compressed data.

Licenser (changes requested):
Heya :) awesome! thank you! only a few small things, one I commented inline, then there should be some tests for the processors and the lookup. You can find examples on how to dot hat in the `preprocessors.rs` there the other compression algorithms are tested too.

The https://github.com/tremor-rs/tremor-www-docs/blob/main/docs/artefacts/postprocessors.md and https://github.com/tremor-rs/tremor-www-docs/blob/main/docs/artefacts/preprocessors.md will need updating to reflect the new processor.

Otherwise 👍 great job :D

src/preprocessor.rs
darach: LGTM - Adding fingerprinting to `decompress` is the last piece of the puzzle - it's very useful when the decompression algorithm can vary. Great work!

mfelsche:
That was quick! 😎 🚀

mfelsche:
We might also need to add libzstd as dependencies to our rpm, deb packages and to the docker image.

dak-x:
@mfelsche Where should I update this dependency for the build.

mfelsche:
I think zstd brings zstd-sys which is bringing its own `libzstd`, so it is actually fine, no need to add anything :) sorry for the confusion.

mfelsche (approved):

Licenser (approved):
Woooh thank you!! 🚀

Issue #1081, opened 2021-06-17
Enhance rest sink for optional GCP auth using service account tokens


darach:
**Describe the problem you are trying to solve**

As a user of the Google Cloud Platform, I would like to be able to use service account based
authentication with the rest/http connectivity in tremor for ad hoc integration against GCP
APIs that are `http/json` based.

**Describe the solution you'd like**

Provide a mechanism to inject an authentication header deriving from authentication with
a specified authentication endpoint and injecting a header when the rest sink is configured
for external authentication.

If a configuration flag ( eg: `uses_gcp_auth` ) is set the `GOOGLE_APPLICATION_CREDENTIALS`
env variable should be resolved to a service account json which can be used to generate a token
for use via http basic authentication.

**Notes**

The mechanism already exists in the google storage sink in a form relatively easy to apply to
the generic rest endpoint as above.

GaryPWhite:
I'm gonna give this a shot! Sounds fun and exciting :)

mfelsche:
@GaryPWhite Awesome! Feel free to reach out if you need any help or bad jokes from us! :)

PR #1103, opened 2021-07-01
Allow for gcp headers to be included on `uses_gcp_auth` option

GaryPWhite:
# Pull request

## Description

Addresses #1081 
Will enhance usage of `rest` sinks by allowing for GCS credentials to be specified when specifying `uses_gcp_auth` feature, and specifying a credentials file with `GOOGLE_APPLICATION_CREDENTIALS` env var

## Checklist

* [x] The code is tested

## Performance
should not have significant impact to performance.
This will only authenticate against GCP API by specific request by a user specifying the `uses_gcp_auth` option. No other performance impact should be noticeable.

### DIFF
--- a/src/connectors/gcp/auth.rs
+++ b/src/connectors/gcp/auth.rs
@@ -18,46 +18,43 @@ use crate::errors::Result;
 /// in the file. The provided PEM file should be a current non-revoked and complete copy of the
 /// required certificate chain for the Google Cloud Platform.
 use gouth::Token;
+use halfbrown::HashMap;
 use reqwest::header::HeaderMap;
 use reqwest::Client;
-use halfbrown::HashMap;
 
 pub(crate) async fn authenticate_bearer() -> Result<String> {
     Ok(Token::new()?.header_value()?.to_string())
 }
 
 pub(crate) async fn json_api_client(extra_headers: &HeaderMap) -> Result<Client> {
-  let bearer = authenticate_bearer().await?;
-  let mut headers = reqwest::header::HeaderMap::new();
-  headers.insert(
-      "authorization",
-      reqwest::header::HeaderValue::from_maybe_shared(bearer)?,
-  );
-  headers.insert(
-      "content-type",
-      reqwest::header::HeaderValue::from_static("application/json"),
-  );
-  for header in extra_headers {
-      headers.append(header.0, header.1.clone());
-  }  
-  Ok(reqwest::Client::builder()
+    let bearer = authenticate_bearer().await?;
+    let mut headers = reqwest::header::HeaderMap::new();
+    headers.insert(
+        "authorization",
+        reqwest::header::HeaderValue::from_maybe_shared(bearer)?,
+    );
+    headers.insert(
+        "content-type",
+        reqwest::header::HeaderValue::from_static("application/json"),
+    );
+    for header in extra_headers {
+        headers.append(header.0, header.1.clone());
+    }
+    Ok(reqwest::Client::builder()
         .default_headers(headers)
         .build()?)
 }
 
-pub(crate) async fn merge_gcp_headers(extra_headers: &HashMap<String,String>) -> Result<HashMap<String, String>> {
-  let bearer = authenticate_bearer().await?;
-  let mut headers: HashMap<String,String> = halfbrown::HashMap::new();
-  headers.insert(
-      "authorization".to_string(),
-      bearer,
-  );
-  headers.insert(
-      "content-type".to_string(),
-      "application/json".to_string(),
-  );
-  for header in extra_headers { // should be a small list so clone em all, I say.
-      headers.insert(header.0.clone().to_string(), header.1.clone().to_string());
-  }
-  Ok(headers)
+pub(crate) async fn merge_gcp_headers(
+    extra_headers: &HashMap<String, String>,
+) -> Result<HashMap<String, String>> {
+    let bearer = authenticate_bearer().await?;
+    let mut headers: HashMap<String, String> = halfbrown::HashMap::new();
+    headers.insert("authorization".to_string(), bearer);
+    headers.insert("content-type".to_string(), "application/json".to_string());
+    for header in extra_headers {
+        // should be a small list so clone em all, I say.
+        headers.insert(header.0.clone().to_string(), header.1.clone().to_string());
+    }
+    Ok(headers)
 }
--- a/src/connectors/gcp/auth.rs
+++ b/src/connectors/gcp/auth.rs
@@ -18,16 +18,18 @@ use crate::errors::Result;
 /// in the file. The provided PEM file should be a current non-revoked and complete copy of the
 /// required certificate chain for the Google Cloud Platform.
 use gouth::Token;
-use halfbrown::HashMap;
 use reqwest::header::HeaderMap;
 use reqwest::Client;
 
-pub(crate) async fn authenticate_bearer() -> Result<String> {
-    Ok(Token::new()?.header_value()?.to_string())
+/// returns `gouth::Token` object and bearer token header as a string
+pub(crate) async fn authenticate_bearer() -> Result<(Token, String)> {
+    let token = Token::new()?;
+    let header_value_string = token.header_value()?.to_string();
+    Ok((token, header_value_string))
 }
 
 pub(crate) async fn json_api_client(extra_headers: &HeaderMap) -> Result<Client> {
-    let bearer = authenticate_bearer().await?;
+    let (_, bearer) = authenticate_bearer().await?;
     let mut headers = reqwest::header::HeaderMap::new();
     headers.insert(
         "authorization",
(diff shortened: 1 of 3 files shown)


jigyasak05:
LGTM🚀

mfelsche:
Could we change the config format for this to be:

```yaml
offramp:
  - id: gcp_rest
    type: rest
    config:
      auth: gcp
```

So that we can easily add other auth methods for other apis and not have a `gcp` specific key here?

darach (dismissed):
Nice! LGTM - @jigyasak05 might be able to give this a whirl on the env we've setup

src/sink/rest.rs
```diff
@@ -449,6 +451,9 @@ impl Sink for Rest {
             .collect::<HashMap<String, Box<dyn Codec>>>();
         let default_method = self.config.method.0;
         let endpoint = self.config.endpoint.clone();
+        #[cfg(uses_gcp_auth)]
```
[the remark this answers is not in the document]
GaryPWhite: I realized the intent when I saw @mfelsche 's comment. This makes sense, I've implemented something like this in the new version.

now... testing it....

src/sink/rest.rs
```diff
@@ -227,6 +229,11 @@ impl<'de> Deserialize<'de> for SerdeMethod {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct Config {
+    /// authorization method, if any.
+    /// Options are empty string (default) and "gcp"
+    #[serde(default)]
+    pub auth: String,
```
Licenser, line 235: It would be really nice to have a enum here so that we prevent strongly typed data and can validate early :)

src/sink/rest.rs
```diff
@@ -449,7 +456,12 @@ impl Sink for Rest {
             .collect::<HashMap<String, Box<dyn Codec>>>();
         let default_method = self.config.method.0;
         let endpoint = self.config.endpoint.clone();
-        let config_headers = self.config.headers.clone();
+        let mut config_headers = self.config.headers.clone();
+        // this could be a match case if other sink auth types are introduced
+        // for now clippy doesn't like equality checks disguised as match cases.
+        if self.config.auth.as_str() == "gcp" {
+            config_headers = block_on(auth::merge_gcp_headers(&config_headers))?
```
Licenser, line 463: This gets a bit complicated, so if I fail on explaining it please ask :D

Google uses OAuth, that means the bearer token has a lifetime and will eventually become stalled. If we only fetch it on initliasiation the sink would work for a while (hours, days or weeks even depending on googles config) and then suddenly throw up "invalid auth" errors. A restart will fix that but probably confuse the heck out of the operator.

To work around this OAuth2 implements multiple token types, one of them being a refresh token. The way we will need to handle this is by storing the `Token` in the struct and fetching `header_values` on every request. This will (based on the documentation from `gouth`, re-use an existing token and fetch and only renew it when required.

src/connectors/gcp/auth.rs
```diff
@@ -39,8 +40,21 @@ pub(crate) async fn json_api_client(extra_headers: &HeaderMap) -> Result<Client>
...
         .build()?)
 }
+
+pub(crate) async fn merge_gcp_headers(
+    extra_headers: &HashMap<String, String>,
+) -> Result<HashMap<String, String>> {
+    let bearer = authenticate_bearer().await?;
+    let mut headers: HashMap<String, String> = halfbrown::HashMap::new();
+    headers.insert("authorization".to_string(), bearer);
+    headers.insert("content-type".to_string(), "application/json".to_string());
```
Licenser, line 54: We should probably not set the content type here and leave it up for the user/service to decide what content type they need or is on a "must-have" for all GCP services?

Licenser (changes requested):

src/connectors/gcp/auth.rs
GaryPWhite: this is what existed in the other implementation for headers, let me see if I can find a doc.

src/sink/rest.rs
GaryPWhite: noted!

src/sink/rest.rs
GaryPWhite: I think I understand.

Here, in `init`, we're only ever refreshing this token once. Since the [authenticate_bearer](https://github.com/tremor-rs/tremor-runtime/blob/main/src/connectors/gcp/auth.rs) function does request `header_values` and specifically refreshes the token, we need to do this _per request_  (like in `build_request`) instead of in `init`.

One thing I want to point out with that approach is that it appears to be inconsistent, at least I think, with the `gcp` approach we had in [`gcs.rs`](https://github.com/tremor-rs/tremor-runtime/blob/main/src/sink/gcs.rs). Namely that the sink will contain the headermap, but never re-authenticate it. I'd be happy to understand how I'm wrong there, still learning :)

I believe I understand the ask, I'll put something in and tag you specifically. I'd also like to open an issue if we agree that the `gcs.rs` could be improved in a similar way.

Licenser: Ja nearly :D we don't need to refresh once per request but every now and then when it times out ( there are [a lot of details in the RFC](https://datatracker.ietf.org/doc/html/rfc6749) but you might not need all of that :joy: ). We are somewhat lucky as in `Token` already implements all that.

`authenticate_bearer` has the problem that it creates a new token for every request, which we really don't want either :D so we kind of need to get rid of `authenticate_bearer` and make `Token` part of the struct so we can re-use the `Token` and just call `header_value` on every request  (that way we get the caching "for free") :D.

For GCS you are absolutely right! It has the same problem :) I didn't notice that on the originally PR for that :( so not your fault at all! I did already threw in #1108.  It should be fairly easy to fix too :D so if you want to feel free to claim the ticket ;) but no need to do it in this PR (it's not your problem :P)

src/connectors/gcp/auth.rs
Licenser: 👍 I'm not sure to be honest, but setting headers can cause trouble for people if they don't know :D so better be safe then sorry I think

GaryPWhite: yeah I can take em out!

src/sink/rest.rs
GaryPWhite: ah, that was the detail I missed. The token itself would need to integrate into the sink, then `header_value` would be generated every request, not a new token... well in that case, maybe this is doable? just return the token instead? lemme fiddle.

Licenser: Ja I think if you just put something like:

```
gcp_token: Option<Token>
```

into the sink, populate that on startup and then use `self.gcp_token.ok_or_else("no token)?.that_header_value_function()?` on each request it should do the trick (just with a lot nicer and more sensible code ;)

src/sink/rest.rs
GaryPWhite: alright, I'm trying to figure out the best way to handle the enum situation.

I have an enum "Auth" defined like this:
```
/// Authorization methods
#[derive(Serialize, Deserialize)]
enum Auth {
  NoAuth,gcp
}
```
for "gcp" and "NoAuth". I've put them in this order intentionally to avoid false positives from `serde`. more on that in a moment.

My main issue with using this enum is identifying to serde which enum should be used for which situation. I've provided a default for it to fall back to (and put `NoAuth` as the first option)
```

```
    #[serde(default=Auth::NoAuth)]
    pub auth: Auth,
```
HOWEVER -- If I were to deserialize from a `.yaml` file, how would Serde know which `enum Auth` type to deserialize into? If provided with
```
config:
  auth: "gcp"
```
How would it possibly know that I meant the ENUM gcp, and not something else? Looking through [serde's docs on enum representations](https://serde.rs/enum-representations.html) it feels like there might be a disconnect between what serde can discern and what we can expect on our end. At the bottom of the page linked, it states:
> There is no explicit tag identifying which variant the data contains. Serde will try to match the data against each variant in order and the first one that deserializes successfully is the one returned.
for `untagged` variants, which is what we would likely want in the scenario of multiple "constant enums". I guess I'm a little lost on how the deserializer would identify the enum. Happy to talk more or hop on a call sometime to talk it through.

Comment (GaryPWhite) [reply src/sink/rest.rs:235]: @mfelsche  was able to help me out -- got the enum thing worked out with `serde`. Thanks!

File: src/sink/rest.rs (hunk 3 above)
Comment (GaryPWhite) [reply src/sink/rest.rs:463]: y'all are total wizards with rust and I am a humble learner still. I really appreciate the advice, it's quite elegant. I'll put that in.

File: src/sink/rest.rs (hunk 2 above)
Comment (Licenser) [reply src/sink/rest.rs]: Weeeh :D nice!

File: src/sink/rest.rs [hunk 5]
```diff
@@ -518,18 +532,30 @@ async fn codec_task(
         host: hostname(),
         ..EventOriginUri::default()
     };
-
+    // create token in this lifetime (ambiguous to compiler), instead of bound to the sink.
+    let token = match auth {
+        Some(Auth::GCP) => {
+            let (t, _) = auth::authenticate_bearer().await?;
```
Licenser, line 538: I would completely avoid `authentication bearer here as there are other things wrong with it too (for example it's not async yet has an async definition) and you incure extra wait by making a blocking requestion, that you don't need to here.
```suggestion
            let t = Token::new()?;
```

src/sink/rest.rs
```diff
@@ -518,18 +532,30 @@ async fn codec_task(
...
+            Some(t)
+        }
+        None => None,
+    };
     let codec: &mut dyn Codec = codec.as_mut();
     while let Ok(msg) = in_rx.recv().await {
         match msg {
             CodecTaskInMsg::ToRequest(event, tx) => {
+                let mut request_headers = default_headers.clone();
+                if let Some(ref t) = token {
```
Licenser, line 548: <3 perfect!

Licenser (approved):
Awesome :rocket: looks great! Thank you so much Gary!!!

Licenser (approved):

Issue #1037, opened 2021-05-19
replace rental


Licenser:
**Describe the problem you are trying to solve**

`rental` the library we are using for self-referential structs is no longer maintained and good alternatives like self_cell are now popping up. This is a good moment to consider replacing the rental code and freeing us from an unmaintained dependency.

**Describe the solution you'd like**

no more rental

**Notes**

This could be done as a multi-step project replacing one usage at a time.

Issue #1024, opened 2021-05-19
Free operators from transmute


Licenser:
**Describe the problem you are trying to solve**

In the trickle operator, we use a transmute due to limitations on what we can express in the trait, perhaps there is a way around this.

```rust
            // This lifetimes will be `&'run mut Value<'event>` as that is the
            // requirement of the `self.runtime.run` we can not declare them
            // as the trait function for the operator doesn't allow that
```

This is based on the pipeline outliving the event.

**Notes**

Followup to #1015

PR #1104, opened 2021-07-02
Unrental event

### LINKED ISSUE #1031: generic consume to remove the need for a rental

Licenser:
# Pull request

## Description

This pull request removes `rental` from our `Event`. And introduces `EventPayload` as an alternative to the rented version.

`EventPayload` follows the same general logic with a few changes. It, as rental or other self-referential structs, obliterates the lifetime of the borrowed part during creation and only allows access to this part as borrows or inside of functions.

To ensure sanity we take a few measures:

We do not generalize over the types, this limits the problem space we reason over. In the implementation, the owned part is **always** a `Vec<u8>` of which we can know that as long as it is not realized the data is not moved in memory. To encode this guarantee, upon creation the `Vec` is wrapped in a `Pin` - this tells the rest compiler it will remain where it is.

Any access to the data is happening to `rent`, `rent_mut`, or `suffix` functions ( the same as with rental ).

Since we have now ownership of the data this allowed us to re-write the `combine` method in a safe manner, also avoiding the cost of `into_static` for events that get combined.

Last but not least, wrapping the `Pin` into an `Arc` which makes this structure safely clonable without having to change the lifetimes of the borrowed part.

All in all, this reduced the need for unsafe and transmutes in the event data-related code to a single (technically two but they are the same) use of `unsafe` and `transmute` for the creation of the self-referential struct. Until rust provides functionality to express self referental structs safely this is something we can not get around.

NOTE:
There is more to be done here but since this is a rather invasive change in itself I want to solicit feedback early on to make sure I don't miss something and this goes horribly wrong o.O.

The next step will be to adopt the same logic for `Script` and `Query` that way we can use the `Arc` for the Script to programmatically ensure scripts out life (or live as long) as the Event which would handle #1024 and perhaps some other transmutes.

CC @Voultapher  would love your thoughts on this.

## Related

* Related Issues:  fixes #1031, start of #1037 

## Checklist

* [x] The code is tested

## Performance

First tests show a ~2% performance increase on non cloned logic, it will be significantly more when event closing is involved :)


src/sink/elastic.rs
```diff
@@ -67,7 +67,7 @@ pub struct Elastic {
     queue: AsyncSink<u64>,
     insight_tx: Sender<sink::Reply>,
     is_linked: bool,
-    response_sender: Sender<(LineValue, Cow<'static, str>)>,
+    response_sender: Sender<(EventPayload, Cow<'static, str>)>,
```
darach, line 70: ❤️  this

src/source.rs
```diff
@@ -242,17 +242,13 @@ where
...
         match self.handle_pp(stream, ingest_ns, data) {
             Ok(data) => {
                 let meta_value = meta.map_or_else(Value::object, |m| m.0);
                 for d in data {
-                    let line_value = LineValue::try_new(vec![Pin::new(d)], |mutd| {
-                        // this is safe, because we get the vec we created in the previous argument and we now it has 1 element
-                        // so it will never panic.
-                        // take this, rustc!
-                        let mut_data = unsafe { mutd.get_unchecked_mut(0).as_mut().get_mut() };
+                    let line_value = EventPayload::try_new::<RentalSnot, _>(d, |mut_data| {
```
darach, line 251: Guessing this is part of another refactoring. Looking forward to this going away too 😄  in due course!

darach (dismissed):
LGTM - 🎸

tremor-script/src/event_payload.rs
```diff
@@ -0,0 +1,285 @@
...
+{
+    fn from((v, m): (T1, T2)) -> Self {
+        ValueAndMeta {
+            v: Value::from(v),
+            m: Value::from(m),
+        }
+    }
+}
+
+/// A event payload in form of two borrowed Value's with a vector of source binaries.
```
mfelsche, line 114: love this comment! :)

tremor-script/src/event_payload.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /// Named after the original rental struct for easy rewriting.
+    ///
+    /// Borrows the borrowed (liftimed) part of the self referential struct
+    /// and calls the provided function with a reference to it
+    pub fn rent<F, R>(&self, f: F) -> R
+    where
+        F: for<'iref, 'head> FnOnce(&'iref ValueAndMeta<'head>) -> R,
+        R: ,
```
mfelsche, line 198: what does this mean, can you leave this out?

tremor-script/src/event_payload.rs
```diff
@@ -0,0 +1,285 @@
...
+        Self {
+            raw: Vec::new(),
+            structured: vm.into(),
+        }
+    }
+}
+
+/// An error occurred while deserializing
+/// a value into an Event.
+pub enum DeserError {
```
mfelsche, line 255: do we need this?

mfelsche (approved):
LGTM, just some questions

tremor-script/src/event_payload.rs
Licenser: I genuenly am not sure, I copied the signature from the rental, and I'm unsure if it means nothing or means `R can't include 'head or 'iref` so I aired on the side of caution

src/source.rs
Licenser: Ja I wanted to keep the surface as small as possible so the only things I changed are the absolutely minimal changes. There is a lot of cleanup that is made possible by the rental removal but I went for a small reviewable change so we can separate the effects of the rental removal from the removal itself and argue about safety on a minimal changeset

tremor-script/src/event_payload.rs
Licenser: yes we can! good catch :)

mfelsche (approved):

src/source.rs
Licenser: was tiny and with you both already having :+1: 'ed this I'll threw it in :D

mfelsche (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Experimental ARM support
 - Add `default => {...}` and `default "key" => "value` to patch
+- Remove `rental` from `Event` [#1031](https://github.com/tremor-rs/tremor-runtime/issues/1031) [#1037](https://github.com/tremor-rs/tremor-runtime/issues/1037)
+- Put event raw payload into `Arc` to improve cloing perf
```
mfelsche, line 10:
```suggestion
- Put event raw payload into `Arc` to improve cloning perf
```

mfelsche (dismissed):

mfelsche (approved):

mfelsche (approved):

Voultapher:
@Licenser it's great to see this PR! I plan on taking a look in the next days, maybe there is something I notice or a suggestion I can contribute.

Licenser:
That'd be great, any feedback is welcome

tremor-script/src/event_payload.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    /// Consumes one payload into another
+    ///
+    /// # Errors
+    /// if `join_f` errors
+    pub fn consume<E, F>(&mut self, mut other: Self, join_f: F) -> Result<(), E>
+    where
+        E: std::error::Error,
+        F: Fn(&mut ValueAndMeta<'static>, ValueAndMeta<'static>) -> Result<(), E>,
```
Voultapher, line 234: Because `join_f` receives both `ValueAndMeta` with supposedly static lifetimes, there are a myriad ways this could be abused and lead to UB. Simplest example, a user calls clone and keeps the ValueAndMeta around -> user after free. I'm sure how to express this ownership transfer into another `ValueAndMeta` while assuming that the thing you originally borrow from is still around, short of actually storing them in Vec. So I suggest at least marking this function `unsafe` and explaining the invariants in great detail.

Voultapher (commented):
This is a great step in the right direction. And I'm delighted to see you invest this much effort as response to the original issue. Given your constrains I agree that a custom hand rolled solution seems the most appropriate, still I suggest additional care around the static lifetime issues I commented on.

tremor-script/src/event_payload.rs
```diff
@@ -0,0 +1,278 @@
...
+impl std::fmt::Debug for EventPayload {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.structured.fmt(f)
+    }
+}
+
+impl EventPayload {
+    /// Borrows the structured part of the struct.
+    #[must_use]
+    pub fn suffix(&self) -> &ValueAndMeta {
```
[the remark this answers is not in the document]
Licenser: The code moved in #1112 so it is a bit outdated and the assumption of unsafety isn't correct. We were concerned about the same thing so added a test to ensure that such code would not be allowed by the compiler: https://github.com/tremor-rs/tremor-runtime/blob/0bf2c1cb123754f3a15a37dc0315dd4e69c85732/tremor-script/src/srs.rs#L420-L435

tremor-script/src/event_payload.rs
```diff
@@ -0,0 +1,278 @@
...
+    #[must_use]
+    pub fn new<F>(raw: Vec<u8>, f: F) -> Self
+    where
+        F: for<'head> FnOnce(&'head mut [u8]) -> ValueAndMeta<'head>,
+    {
+        let mut raw = Pin::new(raw);
+        let structured = f(raw.as_mut().get_mut());
+        // This is where the magic happens
+        // ALLOW: this is sound since we implement a self referential struct
+        let structured: ValueAndMeta<'static> = unsafe { mem::transmute(structured) };
```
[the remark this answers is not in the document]
Licenser: good call!

tremor-script/src/event_payload.rs
```diff
@@ -0,0 +1,278 @@
...
+        Ok(Self { raw, structured })
+    }
+
+    /// Named after the original rental struct for easy rewriting.
+    ///
+    /// Borrows the borrowed (liftimed) part of the self referential struct
+    /// and calls the provided function with a reference to it
+    pub fn rent<F, R>(&self, f: F) -> R
+    where
+        F: for<'iref, 'head> FnOnce(&'iref ValueAndMeta<'head>) -> R,
```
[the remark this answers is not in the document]
Licenser: We wanted to change as little as the interface compared to rental as possible to reduce the number of changes to the codebase (that's why the functions are still called rent, rent_mut) and this is the API rental provided.

tremor-script/src/event_payload.rs
Licenser: That's a good observation, it requires a bit of a contrived example to cause it to violate the guarantees but this was possible:
```rust
    let vec1 = br#"{"key": "value"}"#.to_vec();
    let mut e1 = EventPayload::new(vec1, |d| tremor_value::parse_to_value(d).unwrap().into());
    let vec2 = br#"{"snot": "badger"}"#.to_vec();
    let e2 = EventPayload::new(vec2, |d| tremor_value::parse_to_value(d).unwrap().into());
    let mut v = Value::null();
    // We try to move the data ov v2 outside of this closure to trick the borrow checker
    // into letting us have it even if it's referenced data no longer exist
    e1.consume::<Error,_>(e2, |v1, v2| {
      let (v2,_) = v2.into_parts();
      v = v2;
      Ok(())
    }).unwrap();
    drop(e1);
    println!("v: {}", v);
```

I changed the signature to:

```rust
    pub fn consume<E, F>(
        &mut self,
        mut other: EventPayload,
        join_f: F,
    ) -> std::result::Result<(), E>
    where
        E: std::error::Error,
        F: for<'iref, 'head> FnOnce(
            &'iref mut ValueAndMeta<'head>,
            ValueAndMeta<'head>,
        ) -> std::result::Result<(), E>,
```

which solves the issue and fails on the example given

tremor-script/src/event_payload.rs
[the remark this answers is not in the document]
Licenser: oh and one lifetime doesn't work:
```
   --> tremor-script/src/srs.rs:593:16
    |
578 |         &mut self,
    |         --------- this data with an anonymous lifetime `'_`...
...
593 |         join_f(&mut self.data, other.data)
    |                ^^^^^^^^^^^^^^ ...is captured and required to live as long as `'static` here
```

PR #1106, opened 2021-07-05
Ensure Event metadata is never null

mfelsche:
# Pull request

## Description

Otherwise we render the metadata unusable and unmodifiable in tremor-script. As we can only access it via  we require  to be a record/object.

## Checklist

* [x] The code is tested


Licenser:
I'm a bit unsure that moving the default metadata from `null` to `{}` is good, allocating an object, even if empty, is significantly more expensive then `null`.

`null` seems to be a decent standard for `no metadata`.

If we move away from `null` to an object I could we do some performance comparison to have an idea of the cost of this? I'm especially concealed for the recursion part.

mfelsche:
The reason to use an empty `object` instead of `null` here is that if we'd establish `null` people would always have to do:

```
match $ of
  case null => 
    let $ = {"foo": "bar"}
  case %{} => 
    let $foo = "bar",
end;
```

as it is necessary with `state` for each `script` operator in trickle.
this would be especially bad, since the metadata path `$` is set up so that the only possible access is via some identifier `$identifier` which requires `$` to be an object. The above dance of assigning in different ways is really ugly.

What we could do is to handle metadata assignments differently in that they allocate an object only upon access.

Licenser (approved):
:rocket: wooooh nice job!

PR #1112, opened 2021-07-07
Unrental pt2

### LINKED ISSUE #1024: Free operators from transmute

### LINKED ISSUE #1037: replace rental

Licenser:
# Pull request

## Description

Remove rental from everything but the Kafka source.

## Related

* Related Issues: fixes #1037, fixes #1024

## Checklist

* [x] The code is tested

## Performance

Slight, but repeated performance increase of 1-2%


tremor-cli/Cargo.toml
```diff
@@ -35,31 +35,30 @@ log4rs = "1.0.0"
...
+tremor-pipeline = { path="../tremor-pipeline" }
+tremor-runtime = { path="../" }
+tremor-script = { path="../tremor-script" }
 url = "2"
 # mimalloc-rs = { version = "0.1", default-features = true, optional = true }
 # allocator_api = "0.6.0"
 error-chain = "0.12"
 globwalk = "0.8"
 port_scanner = "0.1"
-rental = "0.5"
```
mfelsche, line 59: ❤️

tremor-script/src/srs.rs
```diff
@@ -0,0 +1,726 @@
...
+            v: Value::from(v),
+            m: Value::from(m),
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+
+    // /// this is commented out since it should fail
```
mfelsche, line 713: you could move this to a doctest using `compile_fail` like so:

```rust
/// ```compile_fail
/// let x = 5;
/// x += 2; // shouldn't compile!
/// ```
```

tremor-value/src/value.rs
```diff
@@ -91,9 +91,15 @@ impl<'value> Eq for Value<'value> {}
 impl<'value> Value<'value> {
     /// Creates an empty array value
     #[must_use]
-    pub fn array() -> Self {
+    pub const fn array() -> Self {
```
mfelsche, line 94: ❤️

mfelsche (dismissed):
LGTM!

That was a lot of reworking, but it turns out so much more elegant! Totally worth it!

tremor-script/src/srs.rs
Licenser: I like that!

mfelsche (approved):
Looking even more GTM!

PR #1126, opened 2021-07-12
Improve handling of EQC tests

Licenser:
# Pull request

## Description

Don't run EQC on dependabot updates


.github/workflows/eqc.yaml
```diff
@@ -3,8 +3,8 @@ name: EQC Tests
 on: [push]
 
 jobs:
-  eqc-tests:
-    if: github.repository_owner == 'tremor-rs'
+  api:
+    if: github.repository_owner == 'tremor-rs' && github.actor!= 'dependabot[bot]'
```
mfelsche, line 7: 🧙

mfelsche (approved):
Cool trick!

.github/workflows/eqc.yaml
Licenser: :D lets hope this one works, and dependabot shows nice and green when we only have EQC failing (or skipping)

PR #1127, opened 2021-07-12
Resturcutre function args to avoid transmutation

### LINKED ISSUE #1028: Remove transmute for function arguments

For custom functions, we use `const.args` as a way to ingest function arguments. In the implementation `env` where const lives is not mutable (as usually it isn't needed, nor should be mutated).

What happens in the function is that we swap the arguments for the function's arguments before calling the function, then swap it back to the original.

Since the only mutation is a pointer swap this should be safe, but hopefully, still avoidable.

Restructure function calls to not depend on transmuting, perhaps arguments can be handled differently. 

**Notes**
Followup to #1015

Licenser:
# Pull request

## Description

Remove the need for transmuting from the handling of function arguments by re-structuring the 'costs' to be referenced rather than owned values.

## Related

* Related Issues: fixes #1028

## Checklist

* [x] The code is tested

## Performance

Within run to run variance of the current code.


mfelsche (approved):
Doesn't look so bad with all those lifetimes at all!

LGTM

PR #1135, opened 2021-07-15
Add support for concatenating arrays

### LINKED ISSUE #1113: Add array::concatinate function

We are missing a function to concatenate two arrays

```
array::concatinate([1], [2]) == [1, 2] 
```

aryan9600:

# Pull request

## Description
This PR adds a new function in the `std::array` module named `concatenate` which takes two arrays and returns a single concatenated array.

```
define script concat
script
    use std::array;
    let event.tags = array::concatenate(event.tags, event.marks);
    event;
end;
```

## Related

* Related Issues: fixes #1113 

## Checklist

* [x] The code is tested

### DIFF
--- a/tremor-script/lib/std/array.tremor
+++ b/tremor-script/lib/std/array.tremor
@@ -73,3 +73,13 @@ intrinsic fn coalesce(array) as array::coalesce;
 ##
 ## Returns a `string`.
 intrinsic fn join(array, string) as array::join;
+
+## Concatenates two arrays returning a new array. The new array is not a set,
+## i.e. it can contain duplicates depending on the input arrays.
+##
+## ```tremor
+## array::concatenate([1, 2, 3], [3, 4]) == [1, 2, 3, 3, 4]
+## ```
+##
+## Returns an `array`
+intrinsic fn concatenate(left, right) as array::concatenate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Restructure functions to avoid transmutation [#1028](https://github.com/tremor-rs/tremor-runtime/issues/1028)
 - Update blackhole to print events / s [#1129](https://github.com/tremor-rs/tremor-runtime/issues/1129)
 - Improve soundness and documentation of SRS code.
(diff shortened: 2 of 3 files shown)


tremor-script/lib/std/array.tremor
```diff
@@ -73,3 +73,13 @@ intrinsic fn coalesce(array) as array::coalesce;
 ##
 ## Returns a `string`.
 intrinsic fn join(array, string) as array::join;
+
+## Concatenates two arrays returning a new array. The new array is not a set,
+## i.e. it can contain duplicates depending on the input arrays.
+##
+## ```tremor
+## array::concatenate([1, 2, 3], [3, 4]) => [1, 2, 3, 3, 4]
```
Licenser, line 81: I see this is copied from above but perhaps we can change `=>` to `==` instead? (same for the `=` and `=>` above while we spotted it) at one point wed' love to have doctests and that'd be a great start to have it in valid syntax :) (not your fault this was us messing it up in the first place :P)

Licenser (approved):
🚀 very nice :D

tremor-script/lib/std/array.tremor
aryan9600: Should we change it to `==` or to `=`? The latter makes more sense imo, as it implies that the executing the `array::concatenate([1, 2, 3], [3, 4])` will result in `[1, 2, 3, 3, 4]`, whereas the former implies an assertion.

mfelsche: I'd say lets do it as all the other docs are done. E.g. for the records it is all `==`: https://docs.tremor.rs/tremor-script/stdlib/std/record/

If we want to change it, lets move that to another issue.

Licenser: Ja the assertion is what we'd be looking for, doc tests are example code that gets executed and it's validity tested. This doesn't work yet - it's on the long list of things we want to get done :D - but it's what we're preparing for so `==` is better then `=` which would be illegal since you can't assign to a function execution

mfelsche (changes requested):
Please add a test for this function below in `std_lib/array.rs`. especially for those cases where one of those arrays is empty or both are.

mfelsche (approved):
😻

PR #1132, opened 2021-07-15
Incoperate feedback on new rnetal free code

Licenser:
# Pull request

## Description

Update of the unrental'ed code based on feedback @Voultapher gave on #1104

## Related

* [RFC](https://rfcs.tremor.rs/0000-.../)
* Related Issues: #1104 #1112 

## Checklist

* [x] The RFC, if required, has been submitted and approved
* [x] Any user-facing impact of the changes is reflected in docs.tremor.rs
* [x] The code is tested
* [x] Use of unsafe code is reasoned about in a comment
* [x] Update CHANGELOG.md appropriately, recording any changes, bug fixes, or other observable changes in behavior
* [x] The performance impact of the change is measured (see below)

## Performance

no relevant code changes

PR #1137, opened 2021-07-18
Make EventPayload lifetime laundering more robust

Voultapher:
This tries to address 2 issues:

- No correlation between 'iref and 'head
- 'head is 'static

The first one is problematic because we know 'head and 'iref are very much
correlated lifetimes. If one could take a reference 'head that outlives 'iref,
that would be unsound.

The second issue is related and maybe even more problematic. While there might
not be any code patterns allowing the abuse of this wrongly marked 'static
lifetime today. Upcoming features like Fn Traits and specialization could easily break
the assumptions here. More pressing even seem potential compiler optimization,
around 'will be alive for whole program duration' lifetimes.

To make the lie about 'static lifetimes more robust, we .. lie even more. By
transmuting the dependent lifetime from 'static to a lifetime bound to the
function call, we ensure that nothing can outlive the scope of the function we
know must be smaller than the lifetime of the object itself. In essence it's the
reverse lie found in the constructor, where we take a function local reference
lifetime and turn it into a 'static lifetime. Technically this approach is still
UB, but the assumption is that it has less potential of breaking in the future,
should for example the constrains of HRTBs be relaxed.

# Pull request

Followup suggestion to https://github.com/tremor-rs/tremor-runtime/pull/1132. It's not complete, there are further consume like functions left. This is aimed to be an POC.

## Checklist

* [x] The code is tested

## Performance

Not tested. But can't see how it would change.

### DIFF
--- a/tremor-script/src/srs.rs
+++ b/tremor-script/src/srs.rs
@@ -609,17 +609,19 @@ impl EventPayload {
         // READ: ORDER MATTERS!
         self.raw.append(&mut other.raw);
 
-        // ALLOW: we are turning a longer lifetime into a shorter one for a
-        // covariant type ValueAndMeta. &mut is invariant over it's lifetime,
-        // but we are choosing a shorter one and passing it down not up. So a
-        // user should not be able to choose an inappropriate lifetime for it,
-        // plus they don't control the owner here.
+        //  we are turning a longer lifetime into a shorter one for a covariant
+        // type ValueAndMeta. &mut is invariant over it's lifetime, but we are
+        // choosing a shorter one and passing it down not up. So a user should
+        // not be able to choose an inappropriate lifetime for it, plus they
+        // don't control the owner here.
         join_f(
             unsafe {
+                // ALLOW: See above for explenation            
                 mem::transmute::<&'iref mut ValueAndMeta<'static>, &'iref mut ValueAndMeta<'iref>>(
                     &mut self.data,
                 )
             },
+            // ALLOW: See above for explenation
             unsafe { mem::transmute::<ValueAndMeta<'static>, ValueAndMeta<'iref>>(other.data) },
         )
     }
--- a/tremor-script/src/srs.rs
+++ b/tremor-script/src/srs.rs
@@ -507,12 +507,13 @@ impl EventPayload {
     where
         F: for<'head> FnOnce(&'head ValueAndMeta<'head>) -> R,
     {
-        // ALLOW: we are turning a longer lifetime into a shorter one for a
-        // covariant type ValueAndMeta. &mut is invariant over it's lifetime,
-        // but we are choosing a shorter one and passing it down not up. So a
-        // user should not be able to choose an inappropriate lifetime for it,
-        // plus they don't control the owner here.
+        // we are turning a longer lifetime into a shorter one for a covariant
+        // type ValueAndMeta. &mut is invariant over it's lifetime, but we are
+        // choosing a shorter one and passing it down not up. So a user
+        // should not be able to choose an inappropriate lifetime for it, plus
+        // they don't control the owner here.
         f(unsafe {
+            // ALLOW: See above explenation
             mem::transmute::<&'iref ValueAndMeta<'static>, &'iref ValueAndMeta<'iref>>(&self.data)
         })
     }
--- a/tremor-script/src/srs.rs
+++ b/tremor-script/src/srs.rs
@@ -434,7 +434,7 @@ impl EventPayload {
     /// ```
 
     #[must_use]
-    pub fn suffix(&self) -> &ValueAndMeta {
+    pub fn suffix<'iref>(&'iref self) -> &'iref ValueAndMeta<'iref> {
         &self.data
     }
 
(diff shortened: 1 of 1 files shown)


tremor-script/src/srs.rs
```diff
@@ -503,11 +503,18 @@ impl EventPayload {
...
+        F: for<'head> FnOnce(&'head ValueAndMeta<'head>) -> R,
     {
-        f(&self.data)
+        // ALLOW: we are turning a longer lifetime into a shorter one for a
+        // covariant type ValueAndMeta. &mut is invariant over it's lifetime,
+        // but we are choosing a shorter one and passing it down not up. So a
+        // user should not be able to choose an inappropriate lifetime for it,
+        // plus they don't control the owner here.
+        f(unsafe {
+            mem::transmute::<&'iref ValueAndMeta<'static>, &'iref ValueAndMeta<'iref>>(&self.data)
```
Licenser, line 516: The allow checks are pretty basic, using grep not parsing or clippy
```suggestion
        // we are turning a longer lifetime into a shorter one for a covariant
        // type ValueAndMeta. &mut is invariant over it's lifetime, but we are
        // choosing a shorter one and passing it down not up. So a user
        // should not be able to choose an inappropriate lifetime for it, plus
        // they don't control the owner here.
        f(unsafe {
            // ALLOW: See above explenation
            mem::transmute::<&'iref ValueAndMeta<'static>, &'iref ValueAndMeta<'iref>>(&self.data)
```

tremor-script/src/srs.rs
```diff
@@ -523,11 +530,20 @@ impl EventPayload {
...
+        F: for<'head> FnOnce(&'head mut ValueAndMeta<'head>) -> R,
     {
-        f(&mut self.data)
+        // ALLOW: we are turning a longer lifetime into a shorter one for a
+        // covariant type ValueAndMeta. &mut is invariant over it's lifetime,
+        // but we are choosing a shorter one and passing it down not up. So a
+        // user should not be able to choose an inappropriate lifetime for it,
+        // plus they don't control the owner here.
+        f(unsafe {
+            mem::transmute::<&'iref mut ValueAndMeta<'static>, &'iref mut ValueAndMeta<'iref>>(
```
Licenser, line 543:
```suggestion
        // we are turning a longer lifetime into a shorter one for a covariant
        // type ValueAndMeta. &mut is invariant over it's lifetime, but we are
        // choosing a shorter one and passing it down not up. So a user 
        // should not be able to choose an inappropriate lifetime for it, plus
        // they don't control the owner here.
        f(unsafe {
            // ALLOW: See above explenation
            mem::transmute::<&'iref mut ValueAndMeta<'static>, &'iref mut ValueAndMeta<'iref>>(
```

tremor-script/src/srs.rs
```diff
@@ -574,23 +590,36 @@ impl EventPayload {
...
+        // but we are choosing a shorter one and passing it down not up. So a
+        // user should not be able to choose an inappropriate lifetime for it,
+        // plus they don't control the owner here.
+        join_f(
+            unsafe {
+                mem::transmute::<&'iref mut ValueAndMeta<'static>, &'iref mut ValueAndMeta<'iref>>(
+                    &mut self.data,
+                )
+            },
+            unsafe { mem::transmute::<ValueAndMeta<'static>, ValueAndMeta<'iref>>(other.data) },
```
Licenser, line 621:
```suggestion
        //  we are turning a longer lifetime into a shorter one for a covariant
        // type ValueAndMeta. &mut is invariant over it's lifetime, but we are
        // choosing a shorter one and passing it down not up. So a user should
        // not be able to choose an inappropriate lifetime for it, plus they
        // don't control the owner here.
        join_f(
            unsafe {
                // ALLOW: See above for explenation            
                mem::transmute::<&'iref mut ValueAndMeta<'static>, &'iref mut ValueAndMeta<'iref>>(
                    &mut self.data,
                )
            },
            // ALLOW: See above for explenation
            unsafe { mem::transmute::<ValueAndMeta<'static>, ValueAndMeta<'iref>>(other.data) },
```

Licenser (approved):
This looks really good :+1: and makes a lot of sense :) thank you!

I added a few suggested changes to appease the CI gods, they can be particular :joy: - but they exist to keep us honest. You got to sign the commits for the DCO the bot has [details on how this can be done](https://github.com/tremor-rs/tremor-runtime/pull/1137/checks?check_run_id=3098652393).

Other then those mechanics :+1:

Voultapher:
I'm not sure if this is the right place to talk about this, but I'm not aware of a better one. Something I noticed while looking at the code, is a proliferation of this code pattern. Both in additional functions added for merging, and with additional? self-referential structures. This worries me, because innocent looking changes like `apply_decl` taking the to be merged thing by ref and not by value, that seem innocent but might violate the tricky constrains. The natural instinct would be to abstract the underlying self-referential struct problem, but that sits in conflict with the deliberate choice to not make this logic generic. Which imo is a good choice because generic constrains are even harder to fulfill and reason about. I wonder if maybe the project should strictly limit the addition of new hand-rolled self-referential structs and member functions to existing ones. In earlier discussions we concluded that the special needs of `EventPayload` are best served with a custom hand-rolled approach. However does this special need extend to all self-referential structs of the project? If no maybe some could be replaced with proven library abstractions, ouroboros or self-cell. Maybe this could help avoid setting this project on a path towards UB in production.

Licenser:
It's totally fine as a place :) for this discussion.

Basically yes, we looked into generalizing but decided against it as it would make code a lot more complex. Even so, it might be more elegant it seemed the simple code is better than elegant code in this case. Specialization also allows us to reason more generally over the possible issues - for example, we know that `ValueAndMeta` never contains references the only place it uses lifetimes is inside of a `Cow` (that's why we could make `into_static` completely safe).

Basically the reasoning needed moves from "this is safe for any structure" to "this is safe for this specific structure" which is orders of magnitude easier ;)

We did move all the places of self-referential structs into the `srs` module so they're contained to a single place instead of - like before - being spread out over different crates and modules. The one place that still has a `rental` is the Kafka source but we ignore this for now as the entire connectors are being rewritten and this will include removing that rental (or more exactly the need for self preferential stcuts at all.

Voultapher:
I agree that making these abstractions generic would be a mistake for this project. Also good idea to group them in the same module, I missed that. Going forward, I'd appreciate it if you could take over this PR and apply the concept in the missing places. I'm happy -albeit a bit slow- to contribute reviews.

Licenser (approved):
:+1:

Licenser:
Closing this as requested, cleanup in #1145

PR #1143, opened 2021-07-23
Windowing again

Licenser:
# Pull request

## Description

- clean up windowing code (better eviction)
- fix a bunch of bugs in the unit testing framework

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/176)

## Checklist

* [x] The code is tested

## Performance

Within run to run variance of main


tremor-script/src/ast.rs
```diff
@@ -994,6 +1002,22 @@ pub enum Expr<'script> {
     Imut(ImutExprInt<'script>),
 }
 
+impl<'script> Expr<'script> {
+    /// Tries to borrow the Expor as an `Invoke`
```
mfelsche, line 1006:
```suggestion
    /// Tries to borrow the Expr as an `Invoke`
```

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -603,2059 +214,261 @@ impl Operator for TrickleSelect {
...
 
-                                // get current window group
-                                let this_group = stry!(get_or_create_group(
-                                    window,
-                                    event_id_gen,
-                                    aggregates,
-                                    group_str,
-                                    group_value,
-                                ));
+            let ctx = EventContext::new(ingest_ns, None);
```
mfelsche, line 382: we should set the cartdinality here, right?

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -603,2059 +214,261 @@ impl Operator for TrickleSelect {
...
-                                    std::mem::swap(&mut scratch1.aggregates, &mut this_group.aggrs);
-                                    scratch1.transactional = this_group.transactional;
-                                    // aggrs.init() and reset transactional state
-                                    this_group.reset();
-                                } else {
-                                    // add this event to the aggr state **AFTER** emit and propagate to next windows
+                    if window_event.emit {
+                        // push
+                        let mut env = env(&ctx, run, node_meta, recursion_limit);
+                        env.aggrs = dbg!(&w.aggrs);
```
mfelsche, line 398: 😱 leftover `dbg!`

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -603,2059 +214,261 @@ impl Operator for TrickleSelect {
...
-                                    }
-                                    this_group.track_transactional(scratch1.transactional);
-                                    this_group.id.track(&event_id_scratch1);
-                                }
-                            }
-                        } else {
-                            // if this happens, we fucked up internally.
-                            return Err("Internal Error. Cannot execute select with multiple windows, no scratches available. This is bad!".into());
+                        if let Some(next) = &mut w.next {
+                            can_remove = next.on_event(
```
mfelsche, line 429: why calling `on_event` for `next` when it has not been called on `w`? Shouldnt we handle the signal case differently?

tremor-script/src/ast/query.rs
```diff
@@ -107,7 +109,7 @@ pub struct SelectStmt<'script> {
     pub aggregates: Aggregates<'static>,
     /// scratches needed when executing multiple windows
     /// only necessary if we have multiple windows, otherwise empty
-    pub aggregate_scratches: Option<(AggregateScratch, AggregateScratch)>,
+    pub aggregate_scratches: (AggregateScratch, AggregateScratch),
```
mfelsche, line 112: i think those can be removed, as they are not used anywhere! 🥳

tremor-script/src/ast/query/raw.rs
```diff
@@ -157,12 +153,12 @@ impl<'script> Upable<'script> for StmtRaw<'script> {
                     .collect();
                 // only allocate scratches if they are really needed - when we have multiple windows
                 let aggregate_scratches = if stmt.windows.len() > 1 {
```
mfelsche, line 155: Kill it with :fire:

mfelsche (commented):
Wow, that is a ton of stuff in here! The resulting select code looks so much better than what was there before. 💯 

Left some comments and questions :)

tremor-pipeline/src/op/trickle/select.rs
Licenser: `w.on_event` would try to accumulate the data, which we don't have, on a select, that would mark the window as tracking data, adding empty values, emitting it, and keeping it around. for signal handling we needed a different logic for `on_event` that circumvent the ingest an event part.

tremor-pipeline/src/op/trickle/select.rs
Licenser: Yes! Good catch :)

tremor-pipeline/src/op/trickle/select.rs
mfelsche: Just wanted to check that this is on purpose and not a blind-spot. Happens with huge workloads like this PR (at least to me).

Licenser: It easily does :D yes, and sorry for the size :)

mfelsche (approved):
BAM!

PR #1145, opened 2021-07-26
Make EventPayload lifetime laundering more robust (DCO fixed)

Licenser:
See #1137 (fixed DCOs)


Voultapher:
Looks good, apart from the DCO there are other places I saw that have the same issue, for example the other consume like functions in the same impl.

Voultapher:
Friendly ping, what's the status of this PR?

mfelsche (approved):
I am not sure I 100% understand what is going on here, but i am 100% sure this is more safe than before. It stops instances of `f` to store their argument somewhere outside their scope with a static lifetime and so create an unsound leak. Now everything is correctly tracked via the `iref` lifetime.

What I don't get is why we need to do the following dance with the `'head` lifetime:

```rust
F: for<'head> FnOnce(&'head ValueAndMeta<'head>) -> R
```

But i will approve nonetheless! :)

Voultapher:
Of course further changes are of your discretion. I'd like to point out that this PR is not complete and the issues it addresses are found in more places in the same file. The original PR was intended as proof-of-concept.

PR #1170, opened 2021-08-09
Add fn for path resolution with defaults on no match

### LINKED ISSUE #1090: Defaulting in Expressions

darach:

# Pull request

Fixes #1090 without changes to grammar or language via introducing `path::try_default` intrinsic fn.

## Description

The function `path::try_default` can traverse structured values based on a list of segments
returning a match, or a default when unsuccessful.

## Related

Grammar changes for default in expressions were explored but not preferred compred to a fn ( this PR ).
A try/catch/throw mechanism was explored but dropped in favour of this PR as it would need an RFC and
it would form the basis of a good mentorship project.

* Related Issues: fixes #1090

## Checklist

* [x] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New featues
 
+- Add `std::path::try_default` fn
 - Experimental ARM support
 - Add `default => {...}` and `default "key" => "value` to patch
 - Add `zstd` pre- and post-processors [#1100](https://github.com/tremor-rs/tremor-runtime/issues/1100)
--- a/tests/script.rs
+++ b/tests/script.rs
@@ -201,4 +201,5 @@ test_cases!(
     heredoc_interpolation,
     heredoc_usefn_interpolation,
     heredoc_regression,
+    path_defaulting,
 );
--- a/tests/scripts/path_defaulting/in
+++ b/tests/scripts/path_defaulting/in
@@ -1,0 +1,4 @@
+{ "base": {"snot": "badger", "arr": [{"beep": "boop"}, 1234]}, "segments": ["snot"], "default": "not-used"}
+{ "base": {"snot": "badger", "arr": [{"beep": "boop"}, 1234]}, "segments": [], "default": "not-used"}
+{ "base": {"snot": "badger", "arr": [{"beep": "boop"}, 1234]}, "segments": ["bad"], "default": "good"}
+{ "base": {"snot": "badger", "arr": [{"beep": "boop"}, 1234]}, "segments": ["arr", 0, "beep"], "default": "good"}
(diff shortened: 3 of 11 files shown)


Licenser:
Could we move or copy some of the tests to `test/script/*`. Those test cases are easier for developers to run (as they're clippy integrated), run quicker (as they can be run w/o building tremor), are included in the coverage, and clippies tooling right now is better than ours when reporting issues.

darach:
Yep - i'll add a copy in scripts and update the PR - it does make sense to have variants there as well.

Licenser (approved):
:rocket:

PR #1172, opened 2021-08-10
Add FLOAT4 and FLOAT8 support to postgres connectors

### LINKED ISSUE #1161: Enhance postgres connectors to support a complete type mapping to/from tremor

The postgres connectivity doesn't support common field types such as float ( functionality ). In both the source
and sink this will produce an error for the field kind - but we aren't explicit about which field kind
cannot be mapped from/to either ( usability ).

All legal postgres types should be convertible.

Patch the to and from record type convertors to support wider support of pg column types.

**Notes**

The addition of floating point would be useful for a project that @draco2003 is working on

darach:

# Pull request

## Description

Plugs limitation in postgres connectors for float compatibility from its basic supported types.

## Related

* Related Issues: fixes #1161

## Checklist

* [ x ] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New featues
 
+- Add support for `FLOAT4` and `FLOAT8` serialization/deserialization to postgres connectors
 - Add `std::path::try_default` fn
 - Experimental ARM support
 - Add `default => {...}` and `default "key" => "value` to patch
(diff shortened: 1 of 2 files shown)


ernadh (dismissed):
LGTM.

Licenser (approved):
LGTM!

Small note (no change needed)

the `match` code like:
```rust
                let val = match self.value.as_i32() {
                    Some(v) => v,
                    None => return Err("Could not serialize as `INT4`".into()),
                };
```
can be made a bit easier to read as:

```rust
                let val = self.value.as_i32().ok_or("Could not serialize as `INT4`")?;
````

darach:
Yep - @Licenser  the `as_` refactoring is nice - i kept to the prevailing style here.
Wanted to see if `docker-compose` would work well in CI - it doesn't ( yet )

Licenser (approved):
🚀 or 🪨 who knows?!? LGTM :D

PR #1173, opened 2021-08-10
Fix GCS Token Refresh

### LINKED ISSUE #1109: GCS does not refresh OAuth2 tokens
**Problem**

The GCS Sink uses `auth::json_api_client` which in turn called `authenticate_bearer` which in turn fetches a OAuth2 token. This happens during startup time meaning the token is not re-generated or refreshed. Part of the OAuth flow is that tokens have a limited lifetime and require periodic refresh.

I am not sure what the times for google oauth tokens are but they probably range from hours to days so this is nothing we would have seen in tests, that said in production the current GCS sink will eventually fail to authenticate and throw errors, this will be very confusing to users.

**Possible Solution(s)**

Fortunately `Token` from `gouth` handles this logic for us. If it is cached and `header_value` is called on reach request it will take the token lifetime into account and request a new bearer token whenever the current one is outdated.

dak-x:
Signed-off-by: Daksh Chauhan <dak-x@outlook.com>

# Pull request

## Description
Introduce a wrapper around the `reqwest::Client`.

## Related

* Related Issues: fixes #1109

## Checklist

* [X] Any user-facing impact of the changes is reflected in docs.tremor.rs
* [X] The code is tested
* [X] The performance impact of the change is measured (see below)

### DIFF
--- a/src/connectors/gcp/auth.rs
+++ b/src/connectors/gcp/auth.rs
@@ -21,20 +21,50 @@ use gouth::Token;
 use reqwest::header::HeaderMap;
 use reqwest::Client;
 
-/// returns `gouth::Token` object and bearer token header as a string
-pub(crate) fn authenticate_bearer() -> Result<(Token, String)> {
-    let token = Token::new()?;
-    let header_value_string = token.header_value()?.to_string();
-    Ok((token, header_value_string))
+#[cfg(not(tarpaulin_include))]
+pub(crate) struct GcsClient {
+    client: Client,
+    token: gouth::Token,
 }
 
-pub(crate) fn json_api_client(extra_headers: &HeaderMap) -> Result<Client> {
-    let (_, bearer) = authenticate_bearer()?;
+#[cfg(not(tarpaulin_include))]
+impl GcsClient {
+    // Create a wrapper around reqwest::client
+    // which keeps the token with itself and inserts on each request.
+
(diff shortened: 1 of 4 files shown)


Licenser (dismissed):
I really like this aproach! Nicely done :) thank you! 🚀

darach:
Nice work! 🎸

mfelsche (approved):

src/connectors/gcp/auth.rs
```diff
@@ -21,28 +21,59 @@ use gouth::Token;
 use reqwest::header::HeaderMap;
 use reqwest::Client;
 
-/// returns `gouth::Token` object and bearer token header as a string
-pub(crate) fn authenticate_bearer() -> Result<(Token, String)> {
-    let token = Token::new()?;
-    let header_value_string = token.header_value()?.to_string();
-    Ok((token, header_value_string))
+pub(crate) struct GcsClient {
```
mfelsche, line 24: Since this has no coverage and can only get coverage when tested against GCS in some way, i would be ok with adding `#[cfg(not(tarpaulin_include))]`, so the coverage doesnt drop overall.

mfelsche: I think we also need to put this on the `impl` part. :/

Licenser (approved):

Licenser:
Thank you @dak-x!

PR #1203, opened 2021-09-07
Troy

darach:
# Pull request

This draft PR contains the language definitions and a prototypical runtime sketch for the troy
deployment language. It is a reasonably complete implementation of the troy syntax and
semantics - but is based on the current mainline or legacy deployment model.

To be complete - this work needs to be integrated with the `connectors` branch and the
runtime and deployment facilities implemented with appropriate changes to the grammar
definitions - specifically `flow` definitions before it is `complete`

## Related

## Checklist

* [ ] The code is tested

## Changes to discuss/breaking changes

* [ ] Mandatory args/wtih
* [ ] Split of args / with between a changeable interface and a static interface (how does this affect windows that have no configuration on startup)
* [ ] in all places `;` now separates commands (i.e. `case => ...; ... ; ...` instead of `case => ..., ..., ...`)
* [ ] `create` without a type (i.e. `create badger` instead of `create operator badger`)
* [ ] forbidding creating two entities with the same name (i.e. `operator snot` and  `script snot`)


src/onramp.rs
```diff
@@ -275,13 +289,14 @@ mod test {
...
             let aggr_reg = tremor_script::aggr_registry();
-            let artefact = tremor_pipeline::query::Query::parse(
+            let artefact = tremor_pipeline::query::Query::parse_with_args(
                 &module_path,
                 "select event from in into out;",
                 "<test>",
                 Vec::new(),
                 &*tremor_pipeline::FN_REGISTRY.lock()?,
                 &aggr_reg,
+                &literal!({}),
```
Licenser, line 299: `parse_with_args` and args being `&literal!({})` seems to be common, would it make sense to call parse w/o args and then default args to a literal `{}`?

src/url.rs
```diff
@@ -12,334 +12,45 @@
...
-    pub const OUT: Cow<'static, str> = Cow::const_str("out");
+    pub const OUT: Cow<'static, str> = tremor_common::url::ports::OUT;
 
     /// standard err port
-    pub const ERR: Cow<'static, str> = Cow::const_str("err");
+    pub const ERR: Cow<'static, str> = tremor_common::url::ports::ERR;
 
     /// standard metrics port
-    pub const METRICS: Cow<'static, str> = Cow::const_str("metrics");
+    pub const METRICS: Cow<'static, str> = tremor_common::url::ports::METRICS;
```
Licenser, line 37: Could be easier with:
```
pub use tremor_common::url::ports::*;
```

tests/deploy_errors/pipeline_bad_arg/deploy.troy
```diff
@@ -0,0 +1,8 @@
+define pipeline passthrough
```
Licenser, line 1: should this live in the pipeline error test folder?

Licenser: never mind I missed the deploy :D

tremor-script/src/ast.rs
```diff
@@ -665,6 +689,10 @@ where
 /// A tremor script instance
 #[derive(Clone, Debug, PartialEq, Serialize)]
 pub struct Script<'script> {
+    /// Start
+    pub start: crate::pos::Location,
```
Licenser, line 693: for consistency in AST we should probably follow the same `mid` (metadata id) logic we use for all the other AST elements instead of inlining the start/end

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+use crate::interpreter::Env;
+use crate::interpreter::ExecOpts;
+use crate::interpreter::LocalStack;
+
+pub type DeployWithExprsRaw<'script> = Vec<(IdentRaw<'script>, ExprRaw<'script>)>;
+
+/// Convenience to evaluate an expression at compile time so that
+/// evaluated results can be used during compilation
+#[macro_export]
+macro_rules! stateless_run_script {
```
Licenser, line 62: Could we turn this into a function? It looks like it doesn't require nay macroisms

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+                // )
+            }
+        }
+    }};
+}
+
+/// Convenience to evaluate an expression at compile time so that
+/// evaluated results can be used during compilation
+#[macro_export]
+macro_rules! stateless_run_expr {
```
Licenser, line 103: same here

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+                )
+            }
+        }
+    }};
+}
+
+/// Convenience to evaluate an expression at compile time so that
+/// evaluated results can be used during compilation
+#[macro_export]
+macro_rules! stateless_run_str {
```
Licenser, line 146: same here

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+        .transpose()
+}
+
+#[derive(Debug, PartialEq, Serialize)]
+#[allow(clippy::module_name_repetitions)]
+pub struct DeployRaw<'script> {
+    pub(crate) config: WithExprsRaw<'script>,
+    pub(crate) stmts: DeployStmtsRaw<'script>,
+}
+impl<'script> DeployRaw<'script> {
```
Licenser, line 203: This looks like it could be handled by `Uppable` making it reusable in all the places we use that trait

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+            let up_value = value;
+            args.insert(moo, up_value.clone_static());
+        }
+        // We inject the args here so that our `script` clause can override based on
+        // our `with` specificational argument overrides and their defaults
+        helper.consts.args = Value::Object(Box::new(args.clone()));
+    }
+
+    // NOTE We don't check for `extra args` here - this is deferred to the resolution in `arg_resolver`
+    helper.consts.args = upper_args.clone();
```
Licenser, line 421: I'm a bit confused about what happens where, we set args, then set it again, and then set it a third time, apparently without reading it in between. I think I'm missing some context here?

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+fn arg_resolver<'script, 'registry>(
+    outer: crate::pos::Range,
+    spec: &[Cow<'script, str>],
+    args_as_params: &Value<'script>,
+    params_spec: &Option<DeployWithExprsRaw<'script>>,
+    helper: &mut Helper<'script, 'registry>,
+) -> Result<Value<'script>> {
+    let mut args = args_as_params.clone();
+    // Process defaulted arguments via `with` expressions
+    helper.consts.args = args_as_params.clone();
```
Licenser, line 434: same as above, I'm a bit confused as to what happens here with `helper.consts.args`

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+    pub(crate) id: String,
+    pub(crate) kind: (Vec<IdentRaw<'script>>, IdentRaw<'script>),
+    pub(crate) args: WithArgsRaw<'script>,
+    pub(crate) params: Option<DeployWithExprsRaw<'script>>,
+}
+
+impl<'script> Upable<'script> for ConnectorDeclRaw<'script> {
+    type Target = ConnectorDecl<'script>;
+    fn up<'registry>(self, helper: &mut Helper<'script, 'registry>) -> Result<Self::Target> {
+        // NOTE As we can have module aliases and/or nested modules within script definitions
```
Licenser, line 532: <3 nice solution I love it :D

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,785 @@
...
+        let target_module = self
+            .module
+            .iter()
+            .map(|x| format!("{}::", x.to_string()))
+            .collect::<Vec<String>>()
+            .join("");
+        let fqsn = format!("{}{}", target_module, self.target.to_string());
+
+        let mut h2 = helper.clone(); // Clone to save some accounting/tracking & cleanup
+        h2.module = vec![];
```
Licenser, line 713: Is it save to remove all modules? Wouldn't that break modularity when used with `use` and `mod`?

src/onramp.rs
darach: Yep - we can remove the `parse` and rename `parse_with_args` to `parse` and pass through the args here I think

tremor-script/src/ast.rs
darach: Well spotted - yes it really should have a mid 👍

tremor-script/src/ast/deploy/raw.rs
darach: This actually started as a function / variant of `run` - but it went quickly downhill due to lifetime inconsistencies.
Also - its worth investigation if we can use `reduce` instead as well. Its worth another run to get these as functions/methods though and to remove the macros. Let's look into that tomorrow ( after coffee! ).

tremor-script/src/ast/deploy/raw.rs
darach: Ditto - as above

tremor-script/src/ast/deploy/raw.rs
darach: Ditto - as above

tremor-script/src/ast/deploy/raw.rs
darach: Now why didn't i think of this!

tremor-script/src/ast/deploy/raw.rs
darach: This was harder to get behaving correctly than it looks - if the args aren't updated again we will see the wrong arguments at various points in subsequent parsing. This is one attempt to correct that - the other was making helper cloneable. I suspect there's a better way to achieve this but there are 1 or 2 other runtime side-effects that still need to be resolved ( helper again ) - once those are patched we can refactor.

tremor-script/src/ast/deploy/raw.rs
darach: Now this one looks like I was low on coffee or high on distraction!

tremor-script/src/ast/deploy/raw.rs
darach: Lol. I am not a fan of this solution at all - but its simple and it works! :D

tremor-script/src/ast/deploy/raw.rs
darach: This is sneaky - we're removing modules from a clone of helper - so this gives us a fake `root` - and the empty modules is effectively the same as loading these from a file. This is sneaky enough that, I fear, it needs a lot more test cases. This is sadly consistent with other module management sneakiness where we inject a fake module to avoid collisions.

As we now have local, core ( special case ) absolute and relative module namespacing and resolution where we only had local, core and absolute before - i suspect we need to revisit this and perhaps other helper sneakiness across the AST.

darach: I feel guiltier ( yet shockingly productive! ) with making Helper cloneable - it really shouldn't be...

tremor-cli/tests/integration/system_metrics/metrics.trickle
```diff
@@ -4,7 +4,7 @@ script
 
   let state = match state of
     # exit when we have all metrics events
-    case %{ present onramp, present pipeline, present offramp } when state.onramp >= 9 and state.pipeline >= 4 and state.offramp >= 6 => emit {"exit": 0, "delay": 100} => "done"
+    case %{ present onramp, present `pipeline`, present offramp } when state.onramp >= 9 and state.`pipeline` >= 4 and state.offramp >= 6 => emit {"exit": 0, "delay": 100} => "done"
```
Licenser, line 7: TODO: Not a comment but a mental note to add `pipeline` as a non backward compatible change for the changelog.

tremor-script/src/ast/deploy.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Calculate the fully qualified name
+    #[must_use]
+    pub fn fqsn(&self, module: &[String]) -> String {
+        if module.is_empty() {
+            self.id.clone()
+        } else {
+            format!("{}::{}", module.join("::"), self.id)
+        }
+    }
+}
```
Licenser, line 103: This might be worth traiting out, we have the `join path + name or name if path is empty` in a number of places.

tremor-script/src/ast/raw.rs
```diff
@@ -2392,6 +2404,28 @@ impl<'script> Upable<'script> for InvokeRaw<'script> {
                         args,
                     })
                 } else {
+                    // Relative locability
```
Licenser, line 2407: This logic seems problematic as it could creat conflicting references if I understand correctly. For example if the same function is defined locally and absolute:

```tremor
mod b with
    function badger() with
      "I'm evil but possible"
    end
end;

mod a with
  mod b with
    function badger() with
      "badger"
    end;
  end;
  define fn snot() with
    badger
  end; 
end;

a::snot() # this when I read correctly would return "I'm evil but possible"
```

tremor-script/src/grammar.lalrpop
```diff
@@ -23,7 +24,172 @@ use beef::Cow;
 
 grammar<'input>;
 
-////////////////////////////// Root statements //////////////////////////////
+////////////////////////////// Troy statements //////////////////////////////
+// These statements for the root of the language.
+
+/// Root
+pub(crate) Deploy : DeployRaw<'input> = {
+    <config:ConfigDirectives> <stmts:DeployStmts> "<end-of-stream>"? => DeployRaw{config, stmts},
```
Licenser, line 32: We probably want a doc comment variant too so deployments can be documented?

tremor-script/src/grammar.lalrpop
```diff
@@ -23,7 +24,172 @@ use beef::Cow;
...
+DeployStmts: DeployStmtsRaw<'input> = {
+    <stmt:DeployStmt> ";" <list:DeployStmts> => {
+      let mut list = list;
+      list.insert(0,stmt);
+      list
+  },
+  <item:DeployStmt> ";"? => vec![item],
+}
+
+DeployStmt: DeployStmtRaw<'input> = {
```
Licenser, line 45: same for the defines, Yatin spotted that missing in trickle and is fixing it as part of the subqueries, we should do it here as well

tremor-script/src/grammar.lalrpop
```diff
@@ -23,7 +24,172 @@ use beef::Cow;
...
+
+ConnectorKind: (Vec<IdentRaw<'input>>, IdentRaw<'input>) = {
+    <mt:ModularTarget> => <>,
+}
+
+DeployArgs: WithArgsRaw<'input> = {
+    "args" <with:WithArgs> => <>,
+}
+
+DeployModuleStmt: DeployModuleStmtRaw<'input> = {
```
Licenser, line 83: I think we need to split the deployment statements into defines and deploys, having defines in modules follows what what tremor script (does today) and trickle (will do with Yatins pr) but we have always excluded anything non definitional in modules and it seems sensible to do the same here.

```troy
mod a with
 define flow snot #...
end
```
should be legal, but 


```troy
mod a with
 define flow snot #...
 deploy flow snot;
end
```
should not

Licenser: ignore this, I got confused in .larlpop naming :D this is already the case

tremor-script/src/grammar.lalrpop
```diff
@@ -23,7 +24,172 @@ use beef::Cow;
...
+DeployLinks: DeployLinksRaw<'input> = {
+    <fields:DeployLinks_> => {
+        let mut fields = fields;
+        fields.reverse();
+        fields
+    },
+}
+
+DeployLinks_: DeployLinksRaw<'input> = {
+    <Sep<DeployLinks_, DeployLink, ",">> => <>
```
Licenser, line 100: this should be `";"` to be in line with the RFC and the other statements we have

tremor-script/src/grammar.lalrpop
```diff
@@ -23,7 +24,172 @@ use beef::Cow;
...
+    <start:@L> "define" "flow" <id:Ident> <links:WithDeployLinks> "end" <end:@L> => DeployStmtRaw::FlowDecl(FlowDeclRaw { start, end, id: id.to_string(), args: vec![], params: None, links }),
+
+    // Allow imports of tremor-script definitions
+    Const => DeployStmtRaw::Expr(Box::new(<>)),
+    FnDecl => DeployStmtRaw::Expr(Box::new(ExprRaw::FnDecl(<>))),
+    Intrinsic => DeployStmtRaw::Expr(Box::new(ExprRaw::FnDecl(<>))),
+}
+
+
+EmbeddedQuery: QueryRaw<'input> = {
```
Licenser, line 146: this should probably use the modular subquery logic from Yatin once his PR is done

src/onramp.rs
darach: I went half way and changed the above to `parse` - assuming we remove support for legacy onramps ( depending on the connectors merge + needs ) - we can cleanup. This leaves deprecate vs hard removal open for discussion closer to the release. Which is the same as defaulting args by effect.

tremor-script/src/grammar.lalrpop
```diff
@@ -1421,6 +1588,15 @@ extern {
...
         "as" => Token::As,
 
+        "pipeline" => Token::Pipeline,
+        "connector" => Token::Connector,
+        "flow" => Token::Flow,
+        "links" => Token::Links,
+        "connect" => Token::Connect,
+        "to" => Token::To,
+        "deploy" => Token::Deploy,
+        "config" => Token::Config,
```
Licenser, line 1598: Maker to remember, not a review comment: this is a great list of new keywords that introduce breaking changes.

tremor-script/src/srs.rs
```diff
@@ -111,6 +239,15 @@ impl Debug for Query {
 }
 
 impl Query {
+    /// Creates a new Query with a pre-existing query
+    #[must_use]
+    pub fn new_from_ast(structured: ast::Query) -> Self {
```
Licenser, line 244: this is not sound we need to capture the the Arc/pin that `ast::Query` was created with or we risk referencing freed memory

src/url.rs
darach: Used the first as this wrapper/redirect will disappear with the connectors rebase.

tremor-script/src/srs.rs
```diff
@@ -13,13 +13,141 @@
...
+    /// This analysis will check that flow definitions and instances
+    /// are correctly defined based on static compile time checks.
+    ///
+    /// Runtime checks are not performed.
+    ///
+    /// # Errors
+    /// If definitions are incomplete or invalid and instances
+    /// are not deployable based on static analysis
+    ///
+    pub fn as_deployment_unit(&self) -> Result<UnitOfDeployment> {
```
Licenser, line 129: I don't think this is sound. since we clone out of `.script` which was transmuted into `'static` the result here will (incorrectly) claim to be `UnitOfDeployment<'static>` leading to the chance of referencing memory that is no longer owned.

To resolve this `UnitOfDeployment` will need to be lifetime less and have a `Vec<Arc<Pin<...>>>` that gets cloned from `self.raw`

tremor-cli/tests/integration/system_metrics/metrics.trickle
Licenser: better captured elsewhere

tremor-script/src/grammar.lalrpop
darach: Now `how` did I miss this? Yes, we absolutely do!

tremor-script/src/grammar.lalrpop
darach: `deploy` is illegal in module statements ( module statement inner production doesn't allow it )
`deploy` is currently constrained to the top level in troy - we'd need extensive changes to the rfc
and further design work to consider more modular deployments. But modular definitions are fine and
consistent with what we already allow in the query language, and Yatin is extending for sub queries.

Also - i think we want the sub-query work merged before considering modular deployments - for example
we likely want to make constraints on pipeline definitions in troy that are fairly strict ( no definitions ) in the
troy production. Yatin likely has good insights here based on his work on modular subquery here as well.

tremor-script/src/grammar.lalrpop
Licenser: I think because we entirely missed it with trickle as well :D until Yatin pointed it out I was absolutely oblivious to that too

tremor-script/src/ast.rs
darach: Done - added a todo for making the name name captured as metadata capture better context in future - hardwired it for expedience now.

tremor-script/src/grammar.lalrpop
darach: Added

tremor-script/src/grammar.lalrpop
darach: We use `,` for `args`, for `config` and for `with` and `;` to separate statements so `,` felt more consistent.
I've checked `;` and it will work here without introducing grammar ambiguity - but to me it looks odd / out of
place.

tremor-script/src/grammar.lalrpop
darach: That's the plan - i'll add a TODO so we have this captured

tremor-script/src/grammar.lalrpop
darach: I added a note with the alternative and we can pick one when we connectors are merged and we have the full syntax of the `connect` expressions in place with some working examples ( benchmarks, integration tests, tutorials ) ported over to the deployment language - perhaps then the better alternative will reveal itself? Defaulted to ';' per rfc but IMO the rfc is wrong and ',' is slightly better for the reasoning above

tremor-script/src/grammar.lalrpop
darach: I added an entry to `CHANGELOG.md` so we don't need to remember

tremor-script/src/srs.rs
darach: Sound-ness is one of the things we need to go through - and document carefully. I suspected
as much but wasn't certain that we'd need to capture the underlying pinned reference.

tremor-script/src/srs.rs
darach: As above - lets document these carefully!

darach:
Todo - Once we review modular scoping, ref-safety for self-referential types ( pin dep refs ) and the known issue that isn't easy to ( yet ) to capture in automated tests - the RFC needs to be updated to reflect local changes - such as module support in the implementation which didn't exist when the RFC was written.

tremor-script/src/grammar.lalrpop
Licenser: Ja unless we need to do it otherwise we should what what was agreed on in the RFC after all that's what it's for. Why `;` seems more correct to me is that we use `;` to separate commands (in script, in troy etc) so things that "do" something, and we use d`,` to separate data (arrays, records, config, with...) `connect a to b` seems closer to `select from a to b` then it does to `a: b`

There are exceptions to this, most notably in `case` which I think we all hate :joy:  but haven't found a way to make larlpop solve this (we might want to look at that) same goes for `patch` where we sepeerate commands with `,` for some reason. (but I think changing those two is out of the scope of that)

tremor-script/src/srs.rs
Licenser: I write a long explanation in the issue https://github.com/tremor-rs/tremor-runtime/pull/1104 which should probably have ended up in the file. but in short the guarantees of the SRS **only** hold true and offer a sound use of transmuting when for every clone, move other extraction of data the corresponding vac+arc+pin is also cloned. That is what guarantees that the referenced data isn't freed until the referencing data is freed. That's where this two functions fall short, they'll need to include the vecarcpin (we need a word for this :joy: )

darach: Great - this is the context I was missing and my brain still harks back to rental/transmute here as we kept that interface which I think I tripped up on!

tremor-script/src/ast/deploy/raw.rs
darach: Resolved as above

tremor-script/src/ast.rs
```diff
@@ -489,26 +492,41 @@ pub type Warnings = std::collections::BTreeSet<Warning>;
 
 /// don't use
 #[allow(clippy::struct_excessive_bools)]
+#[derive(Clone)]
 pub struct Helper<'script, 'registry>
 where
     'script: 'registry,
 {
+    // NOTE We should refactor this into multiple structs as its becoming a bit of a
```
Licenser, line 500: :100: Very good call, perhaps something nested? `Helper{definitions:Definitions, docs...}`

tremor-script/src/ast/raw.rs
```diff
@@ -137,7 +139,25 @@ impl<'script> ScriptRaw<'script> {
                 .map(|d| d.iter().map(|l| l.trim()).collect::<Vec<_>>().join("\n")),
         });
 
+        let start = Location {
```
Licenser, line 142: We could derive Default for Location that way we could use Location::default() here and below and also avoid the extra allow

tremor-script/src/ast/deploy.rs
darach: Traited out - and refactored instances that can use the trait

tremor-script/src/grammar.lalrpop
darach: All top level troy statements now support documentation comments and 🚀 to @yatinmaan for spotting our omission and mistake! Thanks Yatin!

tremor-script/src/grammar.lalrpop
darach: TODO added

tremor-script/src/ast.rs
darach: As i sketched out the basic implementation of troy its the helper where quite a lot of the tension and resistance to refactoring came from. The other locus of tension/resistance is module scoping and referencing. I suspect we can go a long way to cleaning this up through a mini refactoring

tremor-script/src/ast/raw.rs
darach: Yes - Q though - would we use the actual file/stream/buffer offsets by line/column/byte or something that's identifiable as a marker eg: -1/-1/-1 which would mean 'start of stream' or 'end of stream' respectively. If the latter, perhaps -1/-1/-1 is a sensible default here for Script/Query/Deploy - but it wouldn't be for any other AST node that must be within a file/stream/buffer ... ?

darach: For embedded queries - location would have a real line/col/byte offset - so that's nice i think

Licenser: An enum perhaps? That avoids the magic numbers and later confusion how we're at -`1 byte (I know I'll be confused by that :joy: eventually)

` ` ` 
enum Location {
  Start,
  End,
  Position {...}
}
` ` `

darach: Enum might be the one yes

tremor-script/src/ast/raw.rs
darach: A correct script:

```tremor
# FIle: script.tremor
mod b with
    fn badger() with
      "I'm evil but possible"
    end
end;

mod a with
  mod b with
    fn badger() with
      "badger"
    end;
  end;
  fn snot() with
    badger
  end; 
end;

a::snot() 
```

Correctly evaluates to the error:

```
Error in script.tremor:14:5 
   14 |     badger
      |     ^^^^^^ Trying to access a non existing local key `badger`

```

This should exhibit the same behaviour with a troy embedded query with the above as a script operator...

Licenser: ah yes, there was `()` missing in the original it would be

```
# FIle: script.tremor
mod b with
    fn badger() with
      "I'm evil but possible"
    end
end;

mod a with
  mod b with
    fn badger() with
      "badger"
    end;
  end;
  fn snot() with
    badger()
  end; 
end;
```

tremor-script/src/ast/deploy/raw.rs
```diff
@@ -0,0 +1,816 @@
...
+            &helper.meta,
+        ),
+    }
+}
+
+/// Evaluate an expression at compile time with an empty state context
+/// for use during compile time reduction
+/// # Errors
+/// If evaluation of the expression fails, or a legal value cannot be evaluated by result
+pub(crate) fn run_expr<'script, 'registry>(
```
Licenser, line 95: Note from the chat with @yatinmaan before I forget it. In the subquery PR we discovered that using `run` for this is problematic as it will allow evaluating non-const expressions leading to unexpected behaviour, the `reduce` functions are a better fit - modular subqueries has a template on how to do that we can copy here.

darach: Yes, that's fine. Once subquery is merged in we can normalize and prep Troy for merging into connectors

darach: We can look at this now that SQ is merged and rebased into troy

tests/deploy_errors/connector_bad_with/deploy.troy
```diff
@@ -0,0 +1,7 @@
+define snot::badger connector fake
```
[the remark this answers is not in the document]
darach: I like this - lets fix it on the troy branch 👍

tremor-script/src/ast/query.rs
```diff
@@ -217,18 +209,18 @@ pub struct ScriptDecl<'script> {
...
-    pub fn fqsn(&self, module: &[String]) -> String {
-        if module.is_empty() {
-            self.id.clone()
-        } else {
-            format!("{}::{}", module.join("::"), self.id)
-        }
-    }
-}
+impl_fqsn!(ScriptDecl);
+// impl<'script> ScriptDecl<'script> {
```
Licenser, line 213: lets kill this :)

tremor-script/src/srs.rs
```diff
@@ -17,9 +17,200 @@ use crate::{
     errors::{Error, Result},
     prelude::*,
 };
+use halfbrown::HashMap;
 use std::{fmt::Debug, mem, pin::Pin, sync::Arc};
 
-///! Thisn file includes our self referential structs
+/// A fully resolved deployable artefact
+#[derive(Clone, Debug, PartialEq)]
+pub enum AtomOfDeployment {
```
Licenser, line 25: `<'static>` where will violate the SRS guarantees as the enum doesn't include the corresponding vectors.

tremor-script/src/srs.rs
Licenser: unresolving this since it's not resolved

tremor-script/src/srs.rs
```diff
@@ -248,6 +551,79 @@ impl Stmt {
 =========================================================================
 */
 
+/// A query declaration
+#[derive(Clone, PartialEq)]
+pub struct PipelineDecl {
+    /// The identity of this pipeline
+    pub id: String,
+    raw: Vec<Arc<Pin<Vec<u8>>>>,
```
Licenser, line 559: do we really need this here, nothing in the rest of the struct seems to be referencing it

tests/deploy_errors/connector_bad_with/deploy.troy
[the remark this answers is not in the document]
darach: Done. ConnectorKind is now an `<ident>` instead of a `<modular_target>` 👍

tremor-script/src/ast/query.rs
darach: Killed :)

tremor-script/src/srs.rs
darach: Yes - i may have forgotten to check in the updates for connector/flow here. I'll redo. Well spotted

darach: Done

tremor-script/src/srs.rs
darach: `UnitOfDeployment` is now free of lifetimes so I think this is resolved as a consequence

tremor-script/src/ast/deploy/raw.rs
darach: Outdated and the args/with refactoring is more recent. Closing.

tremor-script/src/ast/deploy/raw.rs
darach: Outdated, and the args/with refactoring replaces this. Closing.

tremor-script/src/ast/deploy/raw.rs
darach: I think the refactoring for helper to be non-clone fixes this.

tremor-script/src/ast/raw.rs
darach: Not 100% certain without tests - but I think the args/with and subquery refactoring fixes this. Questions remain on the correct hygienic use of `use` though.

tremor-script/src/ast/deploy/raw.rs
darach: This was resolved with the args/with and subquery/pipeline refactoring. Closing.

tests/deploy_errors/connector_not_found/error.txt
```diff
@@ -1,3 +1,3 @@
 Error: 
     1 | deploy broken from snot::badger
-      |                    ^^^^^^^^^^^^ Artefact used in `from` is not defined or not found: snot::badger
+      |        ^^^^^^ Artefact used in `from` is not defined or not found: snot::badger
```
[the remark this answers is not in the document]
darach: Its a not found error but we probably want to add the kind ( connector, flow, pipeline ) here to be consistent with `create` in flows - and we also want to lock down the top level to deploying `flow` only IIRC from the friday design discussion a week ago...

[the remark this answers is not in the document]
Licenser: Ja even so we'll only have deploy for flows `deploy flow ...` reads better then `deploy ...` 👍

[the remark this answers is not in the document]
Licenser: Oh also what I was referring to is that it should point to `snot::badger` not `broken`

Licenser:
let's close this :D it's merged into #1077

PR #1207, opened 2021-09-10
WIP: Subqueries

### LINKED ISSUE #940: Support for Subqueries in Tremor-Query

Currently tremor supports composition through composing pipelines together, through function composition and through allowing references to query operator definitions and constants in externalizable modules that can be loaded via a module path.

It would be excellent if the modularity in tremor extended fully to the query language so that distinct subgraphs could be modularized and consumed by multiple queries to optimise for reuse of flow oriented logic in tremor.

This would require extending module support in the tremor query language to support sub-graph definitions with parameters that can be declared and used as part of a higher level query. 

Modules in tremor-query in their current state: https://docs.tremor.rs/tremor-query/modules/ 

A Tremor Query constitutes a DAG of nodes and their connections through which events flow on some path, decided at runtime. Subqueries also consitute a DAG and will be part of the query DAG they are used within. The challenge here is how to express such a subquery and how to integrate it into the tremor-query syntax (e.g. how to use it from a `select` query). Also look at how [scripts](https://docs.tremor.rs/tremor-query/operators/#script) and [operators](https://docs.tremor.rs/tremor-query/operators/) are defined in tremor-query.

yatinmaan:
Signed-off-by: yatinmaan <yatinmaan1@gmail.com>

# Pull request

Work in progress.

## Related

* Related Issues: closes #940

## Checklist

* [ ] The code is tested


tremor-script/src/lexer.rs
```diff
@@ -137,6 +137,7 @@ pub(crate) fn ident_to_token(ident: &str) -> Token {
         "use" => Token::Use,
         "as" => Token::As,
         "recur" => Token::Recur,
+        "subquery" => Token::Subquery,
```
Licenser, line 140: `subquery` is a good description for what we implement but reading over this I want to toss in the thought that it might not be a good name for users.

From a developer's point of view, yes this is implemented as a subquery to another query, but is that relevant to a user? Would it be easier to digest to call it `query` instead (probably not?) or use `flow` as troy does?

tremor-script/src/ast/query/raw.rs
```diff
@@ -17,30 +17,73 @@
...
-    params.map(|params| up_params(params, helper)).transpose()
+    match params {
+        MaybeParams::Raw(params) => params
+            .map(|params| up_params(params.into(), helper))
+            .transpose(),
+        MaybeParams::Processed(params) => Ok(params),
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize)]
```
Licenser, line 72: this would be better solved by `Option<Params<'script>>` that way there is no need to duplicate the code

tremor-script/src/ast/query/raw.rs
```diff
@@ -17,30 +17,73 @@
...
-    StreamStmt, Upable, Value, WindowDecl, WindowKind,
+    HashMap, Helper, ImutExpr, ImutExprInt, Location, NodeMetas, OperatorDecl, OperatorKind,
+    OperatorStmt, Path, Query, Registry, Result, ScriptDecl, ScriptStmt, Select, SelectStmt,
+    Serialize, Stmt, StreamStmt, SubqueryDecl, Upable, Value, WindowDecl, WindowKind,
 };
 use crate::ast::visitors::{GroupByExprExtractor, TargetEventRefVisitor};
 use crate::{ast::InvokeAggrFn, impl_expr};
 use beef::Cow;
 
 fn up_params<'script, 'registry>(
```
Licenser, line 30: This could probably be implemented as `impl Upable for Params`

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+}
+impl_expr!(SubqueryDeclRaw);
+
+impl<'script> Upable<'script> for SubqueryDeclRaw<'script> {
+    type Target = SubqueryDecl<'script>;
+    fn up<'registry>(self, helper: &mut Helper<'script, 'registry>) -> Result<Self::Target> {
+        let from = match self.from {
+            Some(ports) => ports.into_iter().map(|ident| ident.up(helper)).collect(),
+            None => Ok(vec!["in".into()]),
+        }?;
```
Licenser, line 262: I think this should work here:

```rust
let from.up(helper)?.unwrap_or_else(||vec!["in".into()]);
```

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+        format! {"A__SUBQ__{}_{}",self.id, id}
+    }
+
+    fn get_args_map<'registry>(
+        &self,
+        decl_params: &Option<HashMap<String, Value<'script>>>,
+        stmt_params: Option<HashMap<String, Value<'script>>>,
+        helper: &Helper<'script, 'registry>,
+    ) -> Result<HashMap<String, Value<'script>>> {
+        match &decl_params {
```
Licenser, line 308: I think `unwrap_or_default` could simplify un un-nest this code

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+        }
+    }
+
+    fn inline_params<'registry>(
+        params: Params<'script>,
+        sq_args: &HashMap<String, Value<'script>>,
+        helper: &mut Helper<'script, 'registry>,
+    ) -> Result<HashMap<String, Value<'script>>> {
+        match params {
+            // ALLOW: Only Raw params can be inlined.
```
Licenser, line 340: This needs a bit of explanation, how do we guarantee that we never try to inline processed params? And even if we do, an error would probably be better (just to be safe) then a `unreachable` this is compile time the bit of overhead isn't that bad compared to the risk of a odd script crashing the runtime

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+            Params::Processed(_) => unreachable!("Can't inline processed params."),
+            Params::Raw(params) => params
+                .into_iter()
+                .map(|(name, value)| {
+                    let value = value.up(helper)?;
+                    if let ImutExprInt::Path(Path::Reserved(ReservedPath::Args {
+                        segments, ..
+                    })) = &value
+                    {
+                        if let Segment::Id { key, .. } = &segments[0] {
```
Licenser, line 350: I'm not sure about this, it seems to assume that the form is always `args.<key>` how do we guarantee that? could someone use `args.snot.badger` here? if so `badger` would be ignored. if someone uses `args` this would also crash the runtime.

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+    }
+
+    fn inline_maybe_params<'registry>(
+        params: MaybeParams<'script>,
+        sq_args: &HashMap<String, Value<'script>>,
+        helper: &mut Helper<'script, 'registry>,
+    ) -> Result<Option<HashMap<String, Value<'script>>>> {
+        match params {
+            // ALLOW: Only Raw params can be inlined.
+            MaybeParams::Processed(_) => unreachable!("Can't inline processed params."),
```
Licenser, line 370: see above, let's expand why this isn't reachable and rather use an error then a crashing construct

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+        // Calculate the fully qualified name for the subquery declaration.
+        let fq_subquery_defn = if self.module.is_empty() {
+            self.target.clone()
+        } else {
+            let module: Vec<String> = self.module.iter().map(ToString::to_string).collect();
+            format!("{}::{}", module.join("::"), self.target.clone())
+        };
+
+        match helper.subqueries.get(&fq_subquery_defn) {
+            None => error_generic(&self, &self, &"subquery not found", &helper.meta),
```
Licenser, line 391: I think this error should spell out which subquery isn't found

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+                                s.params,
+                                &subq_args,
+                                &mut helper,
+                            )?
+                            .into();
+                            let mut s = s.up(helper)?;
+                            s.module.push(subq_module.clone());
+                            query_stmts.push(Stmt::Script(s));
+                        }
+                        StmtRaw::Stream(mut s) => {
```
Licenser, line 453: I think we should check if streams defined here do not overlap with ports otherwise there is the potential for great confusion on the user side

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+                            s.id = self.mangle_id(&s.id);
+                            query_stmts.push(Stmt::Stream(s.up(helper)?));
+                        }
+                        StmtRaw::Select(mut s) => {
+                            // TTODO Inline args in where, having, groupby clauses
+                            s.from.0.id = Cow::owned(self.mangle_id(&s.from.0.id.to_string()));
```
Licenser, line 459: Is this enough to prevent overlap when there are multiple sub queries? especially when they're in modules?

tremor-script/src/ast/query/raw.rs
```diff
@@ -181,6 +239,274 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+                            o.params = SubqueryStmtRaw::inline_maybe_params(
+                                o.params,
+                                &subq_args,
+                                &mut helper,
+                            )?
+                            .into();
+                            query_stmts.push(StmtRaw::OperatorDecl(o).up(&mut helper)?);
+                        }
+                        StmtRaw::SubqueryDecl(mut s) => {
+                            s.params = SubqueryStmtRaw::inline_maybe_params(
```
Licenser, line 484: A short documentation would be nice on how args in nested subqueries behave, I believe they're not recursively resolved right?

Licenser (commented):
This is, all in all, a very elegant solution! I left a few inline comments on tiny bits that stood out, otherwise :rocket: :+1:

Licenser:
Also some tests would be nice `tests/queries` and `tests/query_errors` to show it is working

tremor-script/src/lexer.rs
darach: Yes - the keywords `pipeline` and `flow` are introduced in troy but `query` would be best here I think

tremor-script/src/ast/query/raw.rs
Licenser: this isn't an issue since `custom_subquery` here is not the definition name of the subquery but rather the name given during creating which is enforced to be unique.

tremor-pipeline/src/query.rs
```diff
@@ -376,6 +376,9 @@ impl Query {
                 | Stmt::ScriptDecl(_)
                 | Stmt::OperatorDecl(_)
                 | Stmt::SubqueryDecl(_) => {}
+                Stmt::SubqueryStmt(_s) => {
+                    // TTODO: passthrough `s.id`/`port` to `s.port_stream_map.get(`port)`
```
Licenser, line 380: Example:
```
define query snot
  from in
  into out
with
end;

create subquery badger from snot;

select event from in into snot/in
select event from snot/out into out;
```

should compile to something like

```
# fake input streom for the pipeline
create stream in;

# fake output for the pipeline
create stream out;

# START create subquery badger from snot;
create stream __SUBQ__badger_in;
create stream __SUBQ__badger_out'
select event from __SUBQ__badger_in/out into __SUBQ__badger_out/in;
# END create subquery badger from snot;

select event from in/in into __SUBQ__badger_in;
select event from __SUBQ__badger_out into out/out;
```

tests/query_errors/subquery_unknown_param/error.txt
```diff
@@ -0,0 +1,6 @@
+Error: 
+    8 | create query passthrough_42
+    9 | with
+   10 |  new_param = 43
+   11 | end;
+      | ^^^ Unknown parameter
```
Licenser, line 6: this should point to the parameter that is unknown (I suspect `new_param`)

tremor-script/src/ast/query/raw.rs
```diff
@@ -188,10 +228,401 @@ impl<'script> Upable<'script> for OperatorDeclRaw<'script> {
...
+            .unwrap_or_else(|| vec!["out".into(), "err".into()]);
+
+        let ports_set: HashSet<_> = from
+            .iter()
+            .chain(into.iter())
+            .map(std::string::ToString::to_string)
+            .collect();
+        for stmt in &self.subquery {
+            if let StmtRaw::Stream(stream_raw) = stmt {
+                if ports_set.get(&stream_raw.id).is_some() {
```
Licenser, line 266:
```suggestion
                if ports_set.contains(&stream_raw.id) {
```

darach:
All systems go for launch. We have liftoff 👍 

Let's merge this beauty 💯

PR #1238, opened 2021-10-01
Add changelog entry for our bench runner

Licenser:


darach (approved):
LGTM but ... this will never work 👍

PR #1271, opened 2021-10-19
Factor out ID + Module into a `NodeId` for different Node types

### LINKED ISSUE #1211: refactor out module + string for identifying enteties into a own struct

Many of our AST nodes carry the following elements:

```rust
    /// Module of the node
    pub module: Vec<String>,
    /// ID of the node
    pub id: String,
```

they are used to identify the node and its module, it goes along with the repeated task of concatenating or matching against those. This leads to a lot of duplicated (and potentially diverging) code

A new struct that encapsulates module and id into a struct that generalizes access, rendering and any other common functions over the node name

fisherdarling:
# Pull request

## Description

This PR factors out many instances of `id: String` and `module: Vec<String>` into a new type, `NodeId`.

```rust
/// Identifies a node in the AST.
#[derive(Clone, Debug, PartialEq, Serialize)]
pub struct NodeId {
    /// The ID of the Node
    id: String,
    /// The module of the Node
    module: Vec<String>,
}
```

Thoughts / Main changes.
1. Factoring it out will break anything that relies on the serialization, `#[serde(flatten)]` will keep the same functionality. I did not add it in since I passed all of the tests.
2. I created a trait `FullyQualifiedName` with a `fqn` function that calculates the fully qualified name. I then replaced the `fqXn` methods with `fqn`. It feels cleaner. 
3. `NodeId` feels like an ok name, could use suggestions.
4. Might get some performance boosts by switching to an inline string crate. Something like [smartstring](https://github.com/bodil/smartstring) might help, I think it's worth making an issue for. You could also even make `module` an `ArrayVec`.

## Related

* closes #1211

## Checklist

I'll update the changelog once we commit to the trait / naming.

### DIFF
--- a/tremor-pipeline/src/op/trickle/operator.rs
+++ b/tremor-pipeline/src/op/trickle/operator.rs
@@ -55,7 +55,7 @@ impl TrickleOperator {
             ast::Stmt::OperatorDecl(ref op) => {
                 let op = op.clone().into_static();
                 let config = mk_node_config(
-                    op.id,
+                    op.node_id.id().to_string(),
                     format!("{}::{}", op.kind.module, op.kind.operation),
                     op.params,
                 );
--- a/tremor-pipeline/src/query.rs
+++ b/tremor-pipeline/src/query.rs
@@ -426,24 +426,21 @@ impl Query {
                     subqueries.insert(s.id.clone(), s.clone());
                 }
                 Stmt::Operator(o) => {
-                    if nodes.contains_key(&common_cow(&o.id)) {
-                        let error_func = if has_builtin_node_name(&common_cow(&o.id)) {
+                    if nodes.contains_key(&common_cow(o.node_id.id())) {
+                        let error_func = if has_builtin_node_name(&common_cow(o.node_id.id())) {
                             query_node_reserved_name_err
                         } else {
                             query_node_duplicate_name_err
                         };
-                        return Err(error_func(o, o.id.clone(), &query.node_meta).into());
+                        return Err(
+                            error_func(o, o.node_id.id().to_string(), &query.node_meta).into()
+                        );
                     }
 
-                    let target = o.target.clone().to_string();
-                    let fqon = if o.module.is_empty() {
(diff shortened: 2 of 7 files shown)


Licenser:
Sorry for the late reply on this, it seems like some of the test are failing at the current state :(

fisherdarling:
> Sorry for the late reply on this, it seems like some of the test are failing at the current state :(

No worries. I'll go through those tests and get them fixed this weekend.

fisherdarling:
Alright, so I got clippy to run and `cargo test --all --locked` to pass. The only issue now, is that 3 cli integration tests do not pass.

I'm getting this error for test fg.12
```
Error: 
    1 | use std::string;
      |                 ^ Module `std/string` not found or not readable error in module path: 
    2 | 
    3 | let trimmed = string::trim(event.foo);
```
It seems to be an issue with path resolving. Thoughts on where the issue could lie? I think it has to do with pulling in upstream changes.

mfelsche:
@fisherdarling you need to set the environmnt variable `TREMOR_PATH` to `$PWD/tremor-script/lib` (it needs to be an absolute path) so it points to the standard library location.

fisherdarling:
It looks like it's just missing a couple of tests for `node_id` then?

Licenser:
Ja given the number probably only a single one :D

fisherdarling:
Alrighty! I added a test just pushed a squash commit. I think it's ready for a final PR run

Licenser (approved):
:rocket: lovely improvement :D

Licenser:
ah sod, didn't quite make it :(

fisherdarling:
> ah sod, didn't quite make it :(

That should do it!

Licenser (approved):
Done :D you're awesome @fisherdarling !

fisherdarling:
Thank you @Licenser! Sorry that took a while!

Licenser:
No reason to be sorry at all :) we appreciate the contribution a lot!

PR #1287, opened 2021-10-25
Plugin Development Kit: Connectors

marioortizmanero:
# Pull request

## Description

This Pull Request is my first approach to integrating the new Plugin Development Kit (#791) for the connectors artifact. It requires a *ton* of changes because it moves every single connector into a separate crate, and rewrites part of the internals so that they can be used safely through the FFI boundary.

## Related

* [RFC](https://www.tremor.rs/rfc/accepted/plugin-development-kit)
* Related Issues: #791 #1077
* Related docs: https://nullderef.com/series/rust-plugins/

## Checklist

* [x] The RFC, if required, has been submitted and approved
* [ ] Any user-facing impact of the changes is reflected in docs.tremor.rs
* [ ] The code is tested
* [ ] Use of unsafe code is reasoned about in a comment
* [ ] Update CHANGELOG.md appropriately, recording any changes, bug fixes, or other observable changes in behavior
* [ ] The performance impact of the change is measured (see below)

Custom checklist:
* [ ] Panicking or aborting?
* [ ] Cleaning up `simd_json::Value` in `OpMeta`
* [ ] Supporting `StableAbi` derive in `value_trait`

## Performance

I do have a few benchmarks so far in https://github.com/marioortizmanero/pdk-experiments, which I will analyze in a new post on nullderef.com. These do not represent a real usage of Tremor, however, so I will make a few more once everything is working properly.

PR #1303, opened 2021-10-27
Adding `abi_stable` support for `tremor-value`

marioortizmanero:
# Pull request

## Description

This is an intermediate PR for #1287 so that it's more manageable. In this PR, all I do is make `tremor-value` compatible with `abi_stable`. The main crate won't compile still because it's unmodified.

Everything compiles just fine and it's ready for a review. I still have to fix the tests, which are giving me a bit of a hard time, but maybe the review will help for that.

Some notes:

* `beef::Cow` is removed in place of `RCow`, so we no longer have the memory optimization of `beef::Cow`
* `halfbrown::HashMap` is removed in place of `RHashMap`, which doesn't have `raw_entry` nor `known_key`, so the optimization using the hash directly doesn't work.
* Most of the new dependencies use git because they required changes that aren't pushed upstream. This is blocked by:
  * https://github.com/rodrimati1992/abi_stable_crates/pull/61
  * Which blocks the 0.2.9 release of https://github.com/simd-lite/value-trait, so I had to fork it into https://github.com/marioortizmanero/value-trait so that it's not an alpha version
* Ideally, in the future `simd_json` itself should also have support for `stable_abi`, which would let us remove a few type conversions. But that was a lot of work for this PR.
* Some parts of the code are suboptimal in terms of performance due to issues with lifetimes and conversions between std types and abi_stable. They have a `FIXME` attached above to keep track of them.


Licenser (approved):
:rocket:

marioortizmanero:
After talking with @darach, we both agreed that this might not be the best way to do it. Instead of having a single `Value` that's now `abi_stable` compatible, I suggest having two of them: one for the PDK, and another for regular usage. We can then convert to and from the regular type into the PDK type in order to communicate with the plugins.

* It's much simpler: `abi_stable` was propagating throughout the *entire* Tremor codebase and it was quickly getting out of hands. I was having trouble trying to make `tremor-script` compile, because it also uses `Value` for reasons outside of the PDK. I basically had to rewrite lots of stuff just for the PDK.
* Less importantly: performance. At first it seems like it might be more costly, because having two types means there's a conversion from the regular value to the PDK value in order to pass it to the plugin, and then the inverse to use it in the plugin again. But it turns out that by using a single `Value` we were missing out on lots of optimizations, such as `known_key`, `raw_entry`, the memory efficiency of `beef::Cow`, or the performance of `halfbrown::HashMap`. And we were performing lots of conversions as well anyway, so it's not like this was better.

So yeah, I'm not fully sure having two value types is going to be better, but I'm certain it's worth trying. @Licenser was absolutely right that I shouldn't worry much about this PR because I was probably going to end up rewriting most of it anyway haha. At least having `value-trait` be `abi_stable`-compatible is still useful because I'll use it for the PDK value.

I'll try to rewrite this from almost scratch and let you know when I'm more or less done :)

Issue #791, opened 2021-03-01
Plugin Development Kit ( PDK )


darach:
**Describe the problem you are trying to solve**

The plugin development kit ( PDK ) is an interface that allows developers to develop plugins to tremor with a minimal set of dependencies rather than having to build the full tremor runtime with its full set of dependencies.

**Describe the solution you'd like**

Separating plugins from the runtime has many advantages:
* Faster build and test for the tremor runtime project
* Easier extensibility as plugins ( dynamic shared object libraries ) can be built and deployed without a release cycle
* Fewer dependencies in the tremor runtime project
* Each plugin can be built, tested, and packaged separately
* Standardises lifecycle of the pluggable artefacts
* Encourages standard internal interfaces for tremor micro-architecture where there is no standard internal API or interface today ( design by contract )
* Would allow runtime upgrades of pluggable artefacts without changes to the runtime ( where compatible )

There is an existing RFC for tremor that covers the requirements in depth:

 https://rfcs.tremor.rs/0006-plugin-development-kit/

Suggested initial target:
* Pipeline operators
* Codecs
* Preprocessors
* Postprocessors
* Modules/Functions

Enhancements to initial target:
* Implement connectors RFC ( pre-requirement for connector plugins )
* Contribute to and finalize https://github.com/tremor-rs/tremor-rfcs/pull/32 
* Add source, sink, and peering connectors to pluggable artefacts
* Add a PDK TCK ( test compatibility kit ) that asserts plugin invariants and provides testing mechanisms for plugin developers
* Consider plugin documentation generation and another tooling for better developer convenience and usability
* Make trickle sub-graphs a first-class modular and pluggable artefact

**Notes**

RFC https://rfcs.tremor.rs/0006-plugin-development-kit/

Raman-Kumar:
hi, I have just learned the kubernates and studying the source code.
Can I contribute to this PDK or I need more experience.

marioortizmanero:
As the maintainers of Tremor already know, I was preparing myself to work on this project during the summer via the Google Summer of Code. My proposal wasn't accepted in the end so I won't be able to continue, but the research I made is probably worth taking a look at for whoever attempts this next. I'm posting it here so that it's not lost in the Discord server:

* My proposal, with a brief introduction and some research for FFI: https://nullderef.com/blog/gsoc-proposal/
* An update with research on different technologies that could be used for the PDK: https://nullderef.com/blog/plugin-tech/

Cheers!

[...]

PR #1317, opened 2021-11-03
Adding `abi_stable` support for `tremor-value`, v2

marioortizmanero:
# Pull request

## Description

This is an intermediate PR for #1287 so that it's more manageable. This is the second attempt for `tremor-value`, see #1303 for the first one. My last comment points out why I'm doing this:

> After talking with @darach, we both agreed that this might not be the best way to do it. Instead of having a single `Value` that's now `abi_stable` compatible, I suggest having two of them: one for the PDK, and another for regular usage. We can then convert to and from the regular type into the PDK type in order to communicate with the plugins.
> 
> * It's much simpler: `abi_stable` was propagating throughout the *entire* Tremor codebase and it was quickly getting out of hands. I was having trouble trying to make `tremor-script` compile, because it also uses `Value` for reasons outside of the PDK. I basically had to rewrite lots of stuff just for the PDK.
> * Less importantly: performance. At first it seems like it might be more costly, because having two types means there's a conversion from the regular value to the PDK value in order to pass it to the plugin, and then the inverse to use it in the plugin again. But it turns out that by using a single `Value` we were missing out on lots of optimizations, such as `known_key`, `raw_entry`, the memory efficiency of `beef::Cow`, or the performance of `halfbrown::HashMap`. And we were performing lots of conversions as well anyway, so it's not like this was better.
> 
> So yeah, I'm not fully sure having two value types is going to be better, but I'm certain it's worth trying. @Licenser was absolutely right that I shouldn't worry much about this PR because I was probably going to end up rewriting most of it anyway haha. At least having `value-trait` be `abi_stable`-compatible is still useful because I'll use it for the PDK value.
> 
> I'll try to rewrite this from almost scratch and let you know when I'm more or less done :)

In this PR, all I do is make `tremor-value` compatible with `abi_stable`. The main crate will still compile because it's a backward-compatible change. I'll soon make another PR with the changes to `tremor-runtime`, which are much more complicated than this simple PR.

Once this is approved and more or less ready, I'll merge https://github.com/marioortizmanero/tremor-runtime/pull/3 over it with the full PDK changes.

## Related

* [RFC](https://www.tremor.rs/rfc/accepted/plugin-development-kit)
* Related Issues: #791 #1077
* Related docs: https://nullderef.com/series/rust-plugins/

## Checklist

* [ ] The code is tested


src/pdk/value.rs
```diff
@@ -0,0 +1,68 @@
...
+pub type Object<'value> = RHashMap<RCow<'value, str>, Value<'value>>;
+/// Bytes
+pub type Bytes<'value> = RCow<'value, [u8]>;
+
+/// FFI-safe `Value` type to communicate with the plugins. It's meant to be
+/// converted to/from the original `tremor_value::Value` type and back so that
+/// it can be passed through the plugin interface.
+#[repr(C)]
+#[derive(StableAbi)]
+pub enum Value<'value> {
```
mfelsche, line 17: If this is the `Value` PDK developers get into their hands i think we need to implement more traits from `value_trait` on this one, to make it usable. But this is not a concern for this PR. Please add a Note about it at least.

marioortizmanero: Supposedly it's only used as an intermediate. The regular value is converted to the FFI one, and the in the plugin it's converted back to the regular one.

I still have to figure out how to do these conversions in the least annoying way for the plugin developer but it's definitely doable.

mfelsche: Got it, i wasn't aware of how this was handled on the plugin side. 👍

marioortizmanero:
Closing in favor of #1336

PR #1321, opened 2021-11-05
Introduce the size module

### LINKED ISSUE #1320: module to convert sizes

In many configurations we have to specify sizes, they are (usually) defined as a number representing bytes, but it would be nice to have some more expressive ways.

A module specifying functions to generate different sizes for example `size::mb(10) + size::mb(0.5)`  for 10mb and 5kb and enable this for the common sizes. This can fully be implemented in tremorscript

Agares:
# Pull request

## Description
This PR adds the `std::size` module, as requested in https://github.com/tremor-rs/tremor-runtime/issues/1320

## Related

* Related Issues: fixes #1320

## Checklist

* [x] The code is tested


tremor-script/lib/std/size.tremor
```diff
@@ -0,0 +1,7 @@
+fn kibi(size) with
```
Licenser, line 1: I'm not sure about the names, while they are technically correct I suspect a lot more people will be used to abbreviations as `mb`, `kb`, or `gb` (which are also technically correct :P) but more helpful to the average person.

Agares: yes, but the point here is that while maybe for "mebi" and "kibi" it's almost exclusively used for sizes, we can also add "kilo" and "mega", which are used way more widely, and that will lead to an asymmetry

Licenser: They are but they're infrequently used, `mb`, `kb`, `gb`, are frequently used and do not clash with something like `kilo` or `mega` wither. (using uppercase B's also os fine).

but `kibi` and `mebi` arent units that people are working with on a daily basis. `mb`, `kb` are most unix tools report in those units for example.

I don't see a reason why we would want to make this hard for users?

Agares: Can we at least call them kiB and MiB? those aren't as commonly used, but are out there, and are correct.
I think the issue with units is that there people tend to use incorrect abbreviations, and tough they're technically standardised, nobody in practice cares, and sometimes 1kB means 1024B, sometimes 1000B (and don't even get me started on `KB` and `kb`...).
So the trick here is to find the right balance between correctness and user friendliness

Licenser: I honestly don't know, but I think people will get bitten by it not being the abbreviated that are used on a daily basis (this is a bit of the imperial system problem isn't it?) I don't like it either so. Perhaps it's worth to get some wider feedback from people what they would be expect?

Licenser: With having gotten 5 answers from 4 people on this it seemed like an exercise that didn't move us further :joy: at least we know now that no matter what we choose it's going to be wrong. What you suggested `kiB`/`MiB`/`GB` at least came up twice so let's go with that.

Licenser (approved):
<3 I love that this is our first pure tremor-script PR :D nice job! Thank you!

Licenser (approved):

Licenser (approved):

PR #1325, opened 2021-11-12
Add aggregate functions

ramonacat:
# Pull request

## Related

* [RFC](https://github.com/tremor-rs/tremor-www/pull/118/files)

## Checklist

* [ ] The code is tested


tremor-script/src/ast/aggregate_fn.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// public because lalrpop
+    pub end: Location,
+    /// public because lalrpop
+    pub name: IdentRaw<'input>,
+    /// public because lalrpop
+    pub body: RawAggregateFnBody<'input>,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct RawAggregateFnBody<'input> {
```
Licenser, line 21: just for the sake of consistent naming 
```suggestion
pub struct AggregateFnBodyRaw<'input> {
```

tests/queries/custom_aggregate/query.trickle
```diff
@@ -0,0 +1,22 @@
...
+
+mod m with
+    aggregate collect_joined of
+        init ( ) =>
+            "not an empty string"
+
+        aggregate(st, x) =>
+            st + x
+
+        mergein(st, x) =>
```
Licenser, line 14: We probably want to look at a better name then `mergin`

tests/queries/custom_aggregate/query.trickle
```diff
@@ -0,0 +1,22 @@
+define tumbling window by_two
+with
+  size = 2
+end;
+
+mod m with
+    aggregate collect_joined of
```
Licenser, line 7: reading this I am wondering if `define aggregate` would be better here,  since it's a troy level entity we use `define` for any other troy entity. OTOH, `fn` doesn't use define and while `fn` is in tremorsciprt level there is also a argument to keep them the same.

tests/queries/custom_aggregate/query.trickle
```diff
@@ -0,0 +1,22 @@
+define tumbling window by_two
+with
+  size = 2
+end;
+
+mod m with
```
Licenser, line 6: being in a model makes it look like the aggregate is defined in tremor-script scope, which somewhat doesn't make a lot of sense if it can't be used in a tremor-script context in that case I think it would be better at trickle level

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -69,11 +69,27 @@ impl Select {
             .collect();
         let select = srs::Select::try_new_from_stmt(stmt)?;
         let event_id_gen = EventIdGenerator::new(operator_uid);
-        if let ast::Stmt::Select(SelectStmt { aggregates, .. }) = stmt.suffix() {
+        if let ast::Stmt::Select(SelectStmt {
+            stmt: _,
+            aggregates,
+            consts,
+            locals: _,
```
Licenser, line 76: using `, ..` is usually a better choice as it makes refactoring easier (if we add a field we don't need to change every occurrence where it's not used)

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -311,18 +322,18 @@ impl Operator for Select {
                             dflt_group.value = group_value;
                             dflt_group.value.try_push(v.key().to_string());
 
+                            let recursion_limit = sel_ctx.recursion_limit;
+                            let node_meta = sel_ctx.node_meta.clone();
```
Licenser, line 326: I don't think want to clone this

tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -389,7 +400,8 @@ impl Operator for Select {
 
                     let mut run = consts.run();
                     run.group = &g.value;
-                    run.window = &w.name;
+                    let window_name = w.name.clone();
```
Licenser, line 403: can we avoid this clone?

tremor-script/src/ast/aggregate_fn.rs
```diff
@@ -0,0 +1,174 @@
...
+    type Target = AggregateDecl<'script>;
+
+    fn up<'registry>(
+        self,
+        helper: &mut Helper<'script, 'registry>,
+    ) -> crate::ast::visitors::prelude::Result<Self::Target> {
+        let AggregateDeclRaw(args, body) = self;
+        let args: Vec<Ident> = args.up(helper)?;
+
+        let mut locals = HashMap::new();
```
Licenser, line 99: What happens to locals? it seems to be created filled and then never read

tremor-script/src/grammar.lalrpop
```diff
@@ -580,11 +582,20 @@ ModuleExprs: ExprsRaw<'input> = {
 
 ModuleExpr: ExprRaw<'input> = {
     Const => <>,
+    AggregateFnDecl => ExprRaw::AggregateFnDecl(<>),
```
Licenser, line 585: It's worth considering if this should perhaps live in trickle scope as tremor-script can't use aggregates it doesn't make too much sense to declare it in tremor-script

tremor-script/src/grammar.lalrpop
```diff
@@ -580,11 +582,20 @@ ModuleExprs: ExprsRaw<'input> = {
...
     Intrinsic => ExprRaw::FnDecl(<>),
     Module => ExprRaw::Module(<>),
 }
 
+AggregateFnDecl: AggrFnDeclRaw<'input> = {
+    <doc:DocComment> <start:@L> "aggregate" <name:Ident> "of" <body:AggregateFnBody> "end" <end:@L> => AggrFnDeclRaw{start, end, name, body}
+}
+
+AggregateFnBody: AggrFnBodyRaw<'input> = {
+    <start:@L> "init" "(" ")" <init_body:Effectors> "aggregate" "(" <aggregate_args:FnArgs> ")" <aggregate_body:Effectors> "mergein" "(" <merge_args:FnArgs> ")" <merge_body:Effectors> "emit" "(" <emit_args:FnArgs> ")" <emit_body:Effectors> <end:@L> => AggrFnBodyRaw{start, end, init: InitDeclRaw(init_body), aggregate: AggregateDeclRaw(aggregate_args, aggregate_body), merge: MergeInDeclRaw(merge_args, merge_body), emit: EmitDeclRaw(emit_args, emit_body)}
```
Licenser, line 596: I think a newline would be nice somewhere in this (or being split out into sub rules)

tremor-script/src/registry/custom_aggregate_fn.rs
```diff
@@ -0,0 +1,237 @@
...
+// limitations under the License.
+
+use crate::ast::Exprs;
+use crate::interpreter::{Cont, Env, LocalStack};
+use crate::prelude::{Builder, ExecOpts};
+use crate::registry::{FResult, FunctionError};
+use crate::{AggrType, Value};
+use beef::Cow;
+
+/// public because lalrpop
```
Licenser, line 22: I don't think all those things need to be public, lalrpop should only ever use the `Raw` versions lets make them private so they don't leak implementation and interface

tremor-script/src/registry/custom_aggregate_fn.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub fn aggregate<'event>(&mut self, args: &[&Value], env: &Env<'_, 'event>) -> FResult<()>
+    where
+        'script: 'event,
+    {
+        let mut body_iter = self.aggregate_body.iter().peekable();
+        let mut no_meta = Value::null();
+        let mut state = Value::null().into_static();
+        let mut no_event = Value::null();
+        let mut local_stack = LocalStack::with_size(2);
+        local_stack.values.insert(0, Some(self.state.clone()));
```
Licenser, line 105: can we get around this clone by making `self.state` an option of value and swapping it into the local stack that way we don't need to clone it on every call to aggregate as that would be extremely expensive - but I'm not sure this will work but it might be worth a shot.

tremor-script/src/registry/custom_aggregate_fn.rs
```diff
@@ -0,0 +1,237 @@
...
+                &env,
+                &mut no_event,
+                &mut state,
+                &mut no_meta,
+                &mut local_stack,
+            )?;
+
+            if body_iter.peek().is_none() {
+                if let Cont::Cont(value) = cont {
+                    self.state = value.into_owned().clone_static();
```
Licenser, line 133: I don't think we need `into_owned()` and `clone_static()` here as it would clone the value twice

tremor-script/src/registry/custom_aggregate_fn.rs
```diff
@@ -0,0 +1,237 @@
...
+        let mut no_meta = Value::null();
+        let mut state = Value::null().into_static();
+        let mut no_event = Value::null();
+        let mut local_stack = LocalStack::with_size(2);
+        local_stack.values.insert(0, Some(self.state.clone()));
+        local_stack
+            .values
+            .insert(1, args.get(0).map(|x| (*x).clone()));
+
+        let env = Env {
```
Licenser, line 110: since we have env passed in, do we need to create a new one here?

Licenser (changes requested):
There are a few things in there but it looks really great!

tests/queries/custom_aggregate/query.trickle
Agares: I agree :). I have a hard time coming up with anything though, any suggestions? `merge` is causing conflicts in the grammar, since it's a keyword

Licenser: Brainstorming (so those are not necessarily good ideas :joy:):

* Moving it first could help with the grammar, adding `end` to each section could also help with the grammar (but might not look nice and would break the existing pattern of case)



alternative names:
* `combine` - just another word for merge
* `tilt` - we call the concept of chained windows `tilt frames` so calling the function that ingests the previous window into this `tilt` but it will require more domain knowledge and understanding that `combine`
* `into` - that is already a keyword so might be a win of not adding a newone and the function takes state from another window and puts it `into` this one 
* `consume` or `ingest` - to represent that we're taking state in
* `append` since we're adding the state of another window to this
* `add` - it adds two aggregates together (possibly misleading with adding a singular value, but it's what hdr histogram uses)

Licenser:
closing this as it is too far behind and no longer maintained

PR #1343, opened 2021-11-26
S3

### LINKED ISSUE #1176: Add AWS S3 Connector with Streaming support

It is very common in event processing to stream data to some kind of persistent storage engine for later processing or archiving purposes. One very prominent storage engine is AWS S3.

A common practice is to stream data into files that aggregate across a time window (e.g. 1 hour) or that accumulate a certain number of events or grow to a certain size. An AWS S3 connector should support this style of streaming.

We would like to have an AWS S3 Connector that enables tremor to read S3 objects in a streaming fashion (source-part of the connector) and to write data to S3 objects also in a streaming fashion (sink-part of the connector).

It should support all the common ways of authentication to AWS and maintain authentication across the whole lifetime of the connector (e.g. through token refresh etc.).

It should use the official Rust SDK: https://github.com/awslabs/aws-sdk-rust

dak-x:
# Pull request

## Description
Support for S3 Sink

## Related

* Related Issues: fixes #1176 

## Checklist

* [ ] The code is tested

### DIFF
--- a/src/connectors/s3.rs
+++ b/src/connectors/s3.rs
@@ -1,0 +1,343 @@
+// Copyright 2020-2021, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::connectors::prelude::*;
+use crate::Event;
+use async_std::channel;
+use aws_sdk_s3 as s3;
+use s3::model::{CompletedMultipartUpload, CompletedPart};
+use std::{env, mem};
+use value_trait::ValueAccess;
+
+use s3::Client as S3Client;
+struct S3Connector {
+    client_sender: Option<channel::Sender<S3Client>>,
+    config: S3Config,
+}
+
+struct S3Sink {
+    client_receiver: channel::Receiver<S3Client>,
+    bucket: String,
+    client: Option<S3Client>,
+    buffer: Vec<u8>,
+    current_key: String,
+    upload_id: String,
+    part_number: i32,
+    // bookeeping for parts.
+    parts: Vec<CompletedPart>,
+}
+
+#[derive(Deserialize, Debug, Default)]
+pub struct S3Config {
+    aws_access_token: String,
+    aws_secret_access_key: String,
+    aws_region: String,
+    max_clients: Option<usize>,
+    bucket: String,
+    min_part_size: i32,
+}
+
+impl ConfigImpl for S3Config {}
+
+#[derive(Debug, Default)]
+pub(crate) struct Builder {}
+
+#[async_trait::async_trait]
+impl ConnectorBuilder for Builder {
+    async fn from_config(
+        &self,
+        _id: &TremorUrl,
+        raw_config: &Option<OpConfig>,
+    ) -> Result<Box<dyn Connector>> {
+        if let Some(config) = raw_config {
+            let mut config = S3Config::new(config)?;
+            // Fetch the secrets.
+            config.aws_secret_access_key = env::var(config.aws_secret_access_key)?;
+            config.aws_access_token = env::var(config.aws_access_token)?;
+            Ok(Box::new(S3Connector {
+                client_sender: None,
+                config,
+            }))
+        } else {
+            Err(ErrorKind::MissingConfiguration(String::from("aws-s3")).into())
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl Sink for S3Sink {
+    async fn on_event(
+        &mut self,
+        _input: &str,
+        event: Event,
+        _ctx: &SinkContext,
+        serializer: &mut EventSerializer,
+        _start: u64,
+    ) -> Result<SinkReply> {
+        // todo!("Working on it!")
+
+        if self.client.is_none() {
+            if let Ok(new_client) = self.client_receiver.recv().await {
+                self.client = Some(new_client);
+            }
+        }
+        // FIXME: Check if a new client is required or not.
+        else if let Ok(new_client) = self.client_receiver.try_recv() {
+            self.client = Some(new_client);
+        }
+
+        let ingest_id = event.ingest_ns;
+
+        for (event, meta) in event.value_meta_iter() {
+            // What to do when meta-data is not present.
+            // FIXME: ayyyyy
+            // let operation = meta.get("operation").map(|value| value.to_string());
+
+            let object_key = meta.get("key").map(|value| value.to_string());
+
+            // Also handle aggreagtion here.
+            // FIXME: what to do for multiple data??
+            let data: Vec<u8> = serializer
+                .serialize(event, ingest_id)?
+                .into_iter()
+                .flatten()
+                .collect();
+
+            // self.upload_object(object_key.unwrap(), data[0].clone())
+            //     .await?;
+
+            if object_key.as_ref().unwrap() != &self.current_key {
+                // Complete the uploads for previous key.
+                self.set_new_key(object_key.unwrap()).await?;
+            }
+
+            self.buffer.extend(data);
+
+            // FIXME: Make a better aggregation logic.
+            if self.buffer.len() > 5 * 1024 * 1024 {
+                self.upload_part().await?;
+            }
+
+            // FIXME: Identify errors which indicate whether the client failed.
+            // then notify.
+        }
+
+        Ok(SinkReply::default())
+    }
+
+    // Required later
+    async fn on_signal(
+        &mut self,
+        _signal: Event,
+        _ctx: &SinkContext,
+        _serializer: &mut EventSerializer,
+    ) -> Result<SinkReply> {
+        // Check if a new client is required or not.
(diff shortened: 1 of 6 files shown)


dak-x:
Please review the unwraps. @mfelsche

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+        let s3_config = s3::config::Config::builder()
+            .credentials_provider(Credentials::from_keys(
+                self.config.aws_access_token.clone(),
+                self.config.aws_secret_access_key.clone(),
+                None,
+            ))
+            .region(Region::new(self.config.aws_region.clone()));
+
+        let s3_config = match &self.endpoint {
+            Some(uri) => s3_config.endpoint_resolver(Endpoint::immutable(uri.clone())), // unwrap checked before in config generation
```
mfelsche, line 239: no need for this comment anymore

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+            .bucket(self.config.bucket.clone())
+            .send()
+            .await;
+
+        if exist_bucket.is_err() {
+            return Err(exist_bucket.unwrap_err().into());
+        }
+
+        // Send the client to the sink.
+        self.client_sender.as_ref().unwrap().send(client).await?;
```
mfelsche, line 257:
```suggestion
        if let Some(client_sender) = self.client_sender {
           client_sender.send(client).await?;
        }
```

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+            }
+        } else if let Ok(new_client) = self.client_receiver.try_recv() {
+            self.client = Some(new_client);
+        }
+
+        let ingest_id = event.ingest_ns;
+
+        for (event, meta) in event.value_meta_iter() {
+            // Handle no key in meta.
+            let object_key = match meta.get("key").map(|value| value.to_string()) {
```
mfelsche, line 132: Remember that the metadata should be nested at `$s3.key`. So you'd need to do something like:

```
let object_key = meta.get("s3").get_str("key").map(ToString::to_string) {
...
```

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+                Some(key) => key,
+                None => {
+                    self.current_key.clear();
+                    error!("s3: Missing Key Meta Data in Event");
```
mfelsche, line 136: This will add info about the connector instance to the log:
```suggestion
                    error!("{} Missing Key Meta Data in Event", &ctx);
```

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+use std::{env, mem};
+use value_trait::ValueAccess;
+
+use aws_sdk_s3 as s3;
+use aws_types::{credentials::Credentials, region::Region};
+use s3::model::{CompletedMultipartUpload, CompletedPart};
+use s3::Client as S3Client;
+use s3::Endpoint;
+
+const FIVEMBS: i32 = 5 * 1024 * 1024 + 100; // Some extra bytes to keep aws happy.
```
mfelsche, line 26:
```suggestion
const FIVEMBS: usize = 5 * 1024 * 1024 + 100; // Some extra bytes to keep aws happy.
```

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+    client_receiver: channel::Receiver<S3Client>,
+    bucket: String,
+    client: Option<S3Client>,
+    buffer: Vec<u8>,
+    current_key: String,
+
+    // bookeeping for multipart uploads.
+    upload_id: String,
+    part_number: i32,
+    min_part_size: i32,
```
mfelsche, line 44: Make it `usize` to avoid a cast later.

```suggestion
    min_part_size: usize,
```

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+            .bucket(self.bucket.clone())
+            .key(self.current_key.clone())
+            .send()
+            .await
+            .map(|resp| resp.upload_id.unwrap())?; // This unwrap is safe as we have a successful response.
+        Ok(())
+    }
+
+    async fn upload_part(&mut self) -> Result<()> {
+        // FIXME: remove this once done.
```
mfelsche, line 299: Make it a debug log and add the `&ctx` to give it context.

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+            .key(self.current_key.clone())
+            .send()
+            .await?;
+
+        // warn!("Completed part upload");
+
+        // Insert into the list of parts to upload.
+        self.parts.push(
+            CompletedPart::builder()
+                .e_tag(resp.e_tag.unwrap())
```
mfelsche, line 325: is the etag necessary? What if the response doesn't have it?

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+        let (client_sender, client_receiver) = channel::bounded(2);
+
+        // set the sender and receiver for a new client
+        self.client_sender = Some(client_sender);
+
+        let s3_sink = S3Sink {
+            client_receiver,
+            bucket: self.config.bucket.clone(),
+            client: None,
+            buffer: Vec::new(),
```
mfelsche, line 215: to avoid re-allocations and copying you could do a `Vec::with_capacity(self.config.min_part_size)`. It might not avoid all copying, but at least while you grow the buffer.

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+    async fn complete_multipart(&mut self) -> Result<()> {
+        // Upload the last part if any.
+        if self.buffer.len() > 0 {
+            self.upload_part().await?;
+        }
+
+        let completed_parts = mem::take(&mut self.parts);
+        self.client
+            .as_ref()
+            .unwrap()
```
mfelsche, line 341: instead do an:

```rust
let client = self.client.as_ref().ok_or_else(|| ErrorKind::S3Error("No s3 client available".to_string()).into())?;
```

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+            .upload_id(mem::take(&mut self.upload_id))
+            .key(mem::take(&mut self.current_key))
+            .multipart_upload(
+                CompletedMultipartUpload::builder()
+                    .set_parts(Some(completed_parts))
+                    .build(),
+            )
+            .send()
+            .await?;
+        // FIXME: Remove this.
```
mfelsche, line 353: make it a debug log with `&ctx`.

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    async fn upload_part(&mut self) -> Result<()> {
+        // FIXME: remove this once done.
+        warn!(
+            "Uploading part {} with buffer size {}",
+            self.part_number,
+            self.buffer.len()
+        );
+        let buf = mem::take(&mut self.buffer);
```
mfelsche, line 305: you could do a `mem::swap` with a `Vec::with_capacity(self.config.min_part_size)` to ensure you don't need to copy and reallocate too much when filling the buf.

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+        self.part_number += 1;
+        let resp = self
+            .client
+            .as_ref()
+            .unwrap()
```
mfelsche, line 310: `self.client.as_ref().ok_or_else(|| Error::from("No s3 client available"))?`

mfelsche (commented):
This is good stuff, simple and elegant.
Some minor comments about lingering unwraps.

Also pls add an integration test.

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+        Ok(())
+    }
+
+    async fn initiate_multipart(&mut self, key: String) -> Result<()> {
+        self.current_key = key;
+        self.part_number = 0; // Reset to new sequence.
+        self.upload_id = self
+            .client
+            .as_ref()
+            .unwrap()
```
mfelsche, line 288: See below

src/connectors/s3.rs
```diff
@@ -0,0 +1,373 @@
...
+            .create_multipart_upload()
+            .bucket(self.bucket.clone())
+            .key(self.current_key.clone())
+            .send()
+            .await
+            .map(|resp| resp.upload_id.unwrap())?; // This unwrap is safe as we have a successful response.
```
mfelsche, line 294: could you make this an `and_then` and turn the `upload_id` option into a result with `ok_or_else`?

src/connectors/s3.rs
dak-x: would the meta be `$connector.s3.key` referring from the elastic sink.

mfelsche: Sorry, we needed to strip the first level of nesting, because we ran into language problems (`connector` being a keyword now with troy). So it is `$s3.key`. If you rebase on top of the current `connectors` branch, `elastic` should also follow this pattern.

dak-x:
It seems that I have also included the s3_source file that is not complete in my commits. Sorry for this mistake.

src/connectors/s3.rs
```diff
@@ -0,0 +1,414 @@
...
+        // Check for the existence of the bucket.
+        let exist_bucket = client
+            .head_bucket()
+            .bucket(self.config.bucket.clone())
+            .send()
+            .await;
+
+        if exist_bucket.is_err() {
+            return Err(exist_bucket.unwrap_err().into());
+        }
```
Licenser, line 296: We don't want panicing code if we can avoid it lets rewrite this to:
```suggestion
        if let Err(e) = exist_bucket {
            return Err(exist_bucket.into());
        }
```

mfelsche:
Yes, congrats! It is in!

mfelsche:
Great achievement! Mentorship passed!

Issue #790, opened 2021-03-01
gRCP support 


Licenser:
**Describe the problem you are trying to solve**

Many cloud-native initiatives ranging from Kubernetes, open telemetry, istio, Thanos, and Prometheus and many services and APIs that form part of the GCP suite of APIs are normatively specified in gRPC and Protocol Buffers with backward compatibility via JSON and HTTP/1.1.

Tremor has first-grade support for JSON in its type system, in an environment where protocol buffers are ever-present and growing in popularity, as such extending its type system to natively support protocol buffers and other ordinal type systems and serialization formats is a natural evolution to tremor to reduce the burden on tremor application developers and operators.

**Describe the solution you'd like**

gRPC is a modern protocol that allows remote calls and interactions over the network. It is a de facto standard in the cloud-native world where it has in many cases displaced REST as the way to interact with a services API.

gRPC uses HTTP/2 as a protocol to enable multiplexing, asynchronous communications, and long-lasting connections between client and server.

gRPC also uses Protocol Buffers as an encoding format, this reduces both wire size, encoding, and decoding times by using a more optimized binary format.

From this there are a few requirements:

Tremor needs to be able to encode and decode Protocol Buffer based on a given .proto specification
To be used as a client, tremor needs to have a http/2 client connector
To be used as a server, tremor needs to have a http/2 server connector

**Notes**

Suggested initial target:
* [ ] HTTP/2 linked source (server)
* [ ] HTTP/2 linked sink (client)
* [ ] Protobuf codecs for at least 3 gRPC services
  * [ ] Prometheus
  * [ ] Istio
  * [ ] OpenTelemetry

Enhancements to initial target:
* [ ] An ability to add new Protobuf to value mappings without recompilation (i.e. code generation for the PDK, WASM support for codecs)
* [ ] Code generation for the Protocol Buffer service definitions

NamikoToriyama:
@Licenser 
Hi! I'm interested in working on this project as a part of GSoC. 

I have some questions about this issue.

> HTTP/2 linked source (server)
> HTTP/2 linked sink (client)

According to this [link](https://github.com/cncf/mentoring/blob/master/summerofcode/2021.md#add-grpc-client-and-server-support-to-tremor), we plan to use rust for coding. Is it correct to implement HTTP/2 server with rust?
And is it correct to think that the http2 server here is like a gRPC server?

Finally, I'd appreciate it if you can point me to where I can start to get a better understanding of how things work.

[...]

PR #1382, opened 2021-12-21
Add a gRPC client connector

aryan9600:
# Pull request

## Description

This PR adds a gRPC client connector. Tremor needs to be built with the protobuf information and with the grpc feature flag. For example: `PROTO_FILES="proto/helloworld/helloworld.proto,proto/routeguide/route_guide.proto" INCLUDE="proto/helloworld,proto/routeguide" cargo build --features grpc`

## Related

* Related Issues: affects #790 

## Checklist

* [x] The code is tested


.cargo/config
```diff
@@ -5,7 +5,7 @@ rustflags = ["-C", "target-feature=+avx,+avx2,+sse4.2"]
 rustflags = ["-C", "target-feature=+avx,+avx2,+sse4.2"]
 
 [target.aarch64-apple-darwin]
-rustflags = ["-C", "target-feature=+avx,+avx2,+sse4.2"]
```
mfelsche, line 8: Good catch!

Cargo.toml
```diff
@@ -206,8 +208,14 @@ testcontainers = "0.12"
 signal-hook = "0.3"
 signal-hook-async-std = "0.2"
 
+[build-dependencies]
+tremor-grpc-build = { git = "https://github.com/aryan9600/tremor-grpc-build", optional = true }
```
mfelsche, line 212: Can we pin this to a certain release tag?

src/connectors/impls/grpc/client.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    async fn connect(&mut self, _ctx: &ConnectorContext, _attempt: &Attempt) -> Result<bool> {
+        let grpc_client_handler = tremor_grpc::GrpcClientHandler::connect(
+            format!("{}:{}", self.config.host, self.config.port),
+            self.tx.clone(),
+        )
+        .await?;
+        if let Some(sink) = &mut self.sink {
+            let mut client_handler = sink.grpc_client_handler.lock().await;
```
mfelsche, line 119: sink and source now also have a `connect` method, so you can (and should) move this there, as the clone above will result in operating on two distinct entities

src/lib.rs
```diff
@@ -14,7 +14,7 @@
 
 //! Tremor runtime
 
-#![deny(warnings)]
+// #![deny(warnings)]
```
mfelsche, line 17: can you comment this in? If this shall be changed, i'd recommend doing this in another PR. (It is totally fine for the Draft, though)

Cargo.toml
```diff
@@ -175,6 +175,8 @@ tonic = { version = "0.5.2", default-features = false, features = [
   "transport",
   "tls",
 ] }
+prost = "0.8"
```
mfelsche, line 178: do we need those unconditionally, or can they be also only included when activating the `grpc` feature?

src/connectors.rs
```diff
@@ -1159,6 +1159,8 @@ pub fn builtin_connector_types() -> Vec<Box<dyn ConnectorBuilder + 'static>> {
         Box::new(impls::crononome::Builder::default()),
         Box::new(impls::s3::Builder::default()),
         Box::new(impls::kafka::consumer::Builder::default()),
+        #[cfg(feature = "grpc")]
+        Box::new(impls::grpc::client::Builder::default())
```
mfelsche, line 1163: Does this mean, we can only have 1 connector that runs any number of grpc services provided in the `PROTO_FILES` env var per compiled tremor binary? If we want to have, say, two different connectors from two different sets of proto files, would this be possible? 

Can you clarify what exactly can be run with tremor given the setup here? How is the grpc connector scoped?

What can we do to have different connector types per service? Or somehow group services on the command line? One connector type per proto file?

mfelsche (commented):
First set of tired evening eyes glances over this draft. Looks good. One big question on the scope of the connector - will it handle all the services that were given in `PROTO_FILES` etc?

src/connectors.rs
aryan9600: Right now, we have one connector that talks to all the services defined in all the provided files under `PROTO_FILES`. I don't know whether it'd be possible to have one connector per file, but it'd definitely require a lot of refactoring and restructuring.

PR #1401, opened 2022-01-04
Strip unused dependencies and get rid of dependency on openssl.

mfelsche:
# Pull request

## Description

This gets rid of the dependency on openssl, as this was always compiled in a new build, we might be a little faster for fresh builds, but can also get rid of `openssl` as a dep in our docker container and packages. `ldd ./target/debug/tremor` does not list `libssl` anymore :)

We now use `rustls` for amqp via `lapin`, for http offramp via `surf` and `reqwest`. We might need to reintroduce `openssl` again, when we need support e.g. for kafka.

## Checklist

* [x] The code is tested

## Performance

not applicable, we have no benches using TLS.


Licenser (approved):
:rocket:

PR #1412, opened 2022-01-12
Update to clap 3

Licenser:

# Pull request

Update to clap 3, this moves from the cli.yaml to cli.rs. In addition to that clap 3 introduces a few breaking changes on the CLI interface:
  - `tremor server run -f file1 file2` now is `tremor server run file1 file2`
  - `tremor test -i i1 i2 -e e1 e2` is now `tremor test -i i1 -i i2 -e e1 -e e2`

The API portion of the cli is disabled for now and not translated as it will change with 12 and it made no sense to do the work when this will never be released in the current form and just deleted with the connectors / troy work


mfelsche:
Would it make sense to only do this upgrade against the connectors branch and not main, if it is never released anyways?

Licenser:
We do all updates to dependencies vs main none of them will be released before connectors so I would rather stay consistent there than make an exception for one but not all other dependencies.

mfelsche:
I am just afraid of the unnecessary merge/rebase work this will incur if we do this on main.

Licenser:
Since the work is done already, we'll get this one way or the other, and we'll get that to a degree with everything we merge to main.

Licenser:
And rebasing of connectors would be the next thing I would want do after this is merged  :D unless someone has work in progress things going on on `connectors` there are quite a few things we're behind on on the branch

mfelsche (approved):

mfelsche (approved):

PR #1484, opened 2022-02-24
Integrates clap_mangen to tremor-cli

### LINKED ISSUE #1469: Generate manpages from tremor-cli clap definition

We lost our `cli.yaml` due to a `clap` upgrade. But we generated docs for our website from this yaml file. So we cannot generate any docs on our cli anymore. It would be nice to have this docs.

With https://crates.io/crates/clap_mangen we have a crate that we could use to generate manpages. Those could be used to be rendered to markdown or HTML and be displayed on the website and they could also be packaged alongside with our release packages.

AllMight2099:
Integrates clap_mangen into tremor-cli and exports man pages to 'target/man-page/' directory

# Pull request

## Description
As discussed earlier with @mfelsche, this PR adds man page generation to tremor-cli by using the clap-mangen crate and exports the man page to the ```man-page``` directory under ```target```. This PR adds a ```build.rs``` to the tremor-cli project and allows the pages to be generated when building the project

## Related

* Related Issues: fixes #1469 
* Related repository [clap-mangen example](https://github.com/sondr3/clap-man-example)

## Checklist

* [ ] The code is tested

## Performance

r? @mfelsche

### DIFF
--- a/tremor-cli/build.rs
+++ b/tremor-cli/build.rs
@@ -1,0 +1,43 @@
+use clap::IntoApp;
+use clap_mangen::Man;
+use std::{
+    env,
+    fs::File,
+    io::Error,
+    path::{Path, PathBuf},
+};
+
+include!("src/cli.rs");
+
+
+fn build_manpages(outdir: &Path) -> Result<(), Error> {
+    let app = Cli::command();
+
+    let file = Path::new(&outdir).join("tremor-cli-man-page.1");
+    let mut file = File::create(&file)?;
+
+    Man::new(app).render(&mut file)?;
+
+    Ok(())
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-changed=src/cli.rs");
+    println!("cargo:rerun-if-changed=man");
+
+    let outdir = match env::var_os("OUT_DIR") {
+        None => return Ok(()),
+        Some(outdir) => outdir,
+    };
+
+    // Create `target/man-pages/` folder.
+    let out_path = PathBuf::from(outdir);
+    let mut path = out_path.ancestors().nth(4).unwrap().to_owned();
+    path.push("man-pages");
+    std::fs::create_dir_all(&path).unwrap();
+
+    // build_shell_completion(&path)?;
+    build_manpages(&path)?;
+
+    Ok(())
+}
--- a/tremor-cli/build.rs
+++ b/tremor-cli/build.rs
@@ -1,3 +1,17 @@
+// Copyright 2020-2021, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


tremor-cli/build.rs
```diff
@@ -0,0 +1,43 @@
+use clap::IntoApp;
+use clap_mangen::Man;
+use std::{
+    env,
+    fs::File,
+    io::Error,
+    path::{Path, PathBuf},
+};
+
+include!("src/cli.rs");
```
Licenser, line 10: Do we really want to copy in the entire source here?

Licenser:
I did run this and take a  look, it seems that the manfile is  still missing quite a lot of content and not details the sub commands, perhaps they have to be generated additionally?

tremor-cli/build.rs
mfelsche: What else would you do? We can't `use` our code from the `src` dir, but we got to have the definition at hand somehow. I don't see another way and I don't dislike this one.

mfelsche:
Worst case, we need to create one manpage per subcommand or adapt clap_mangen upstream to handle subcommands better. But on the other hand, i do recall that e.g. `git` has 1 manpage for the global `git` command, and 1 for each subcommand (e.g. `man 1 git-branch` ). My first guess is that the structure of a man page is not optimized for commands with many subcommands. I would like to follow the model of git here. Thoughts?

tremor-cli/build.rs
AllMight2099: Yeah, I couldn't find another way of importing the Cli object because we need it to generate the man-page, so I did it this way

AllMight2099: But is there another way of doing it?

AllMight2099:
Just looked at the way git has man pages and it does make a lot more sense. I initially thought having one man page per subcommand was a bit too much. But if we could reference them in the tremor-cli manpage and then add more details to the subcommand, it would help users understand it better.

AllMight2099:
> I did run this and take a look, it seems that the manfile is still missing quite a lot of content and not details the sub commands, perhaps they have to be generated additionally?

Yeah, just realized that, but I also wanted to ask you guys if the format of the man page was alright. I think I'll have to tweak mangen a little bit, but I think it is possible to generate them for the other sub commands

AllMight2099:
The subcommands don't have that much information which would require their own man page, so it makes sense to just keep them in one man page and section them

AllMight2099:
@Licenser @mfelsche I won't be available for 2 days cause I have my exams. I'll push in my updates and probably the final code after 2 days.

Licenser:
Thanks for the heads up! We appriciate it, and no worries Univerity, Family, life should always come first :)

darach:
👋  @AllMight2099 Howdy, wondering if the final updates are ready to push/review? Looking forward to see what you've done with the final tweaks! 👍

Licenser:
Closing this because of inactivity.

Issue #1353, opened 2021-12-05
`deny` statemements in `lib.rs` should be enforced in the CI rather than in the code


marioortizmanero:
**Problem**

Many developers will agree with me that this is considered harmful (not the recursion limit):

![image](https://user-images.githubusercontent.com/25647296/144759812-6a6977d3-596f-41d0-a544-7b179e378246.png)

Personally, for my work on the PDK I had to remove these in order to be able to prototype. Getting a hard error from a warning about not needing `mut` or not using a function is frustrating. I'm prototyping, all I want is to run the code for now, and I'll clean it up later! My final version won't have any warnings anyway because they should be caught by the CI.

I'll link a couple sources with information outside of my personal opinion, that I would strongly suggest to read, as they aren't too long. They expose different issues that come with `deny(warnings)`, though they may be less relevant in binaries.

1. https://github.com/rust-unofficial/patterns/blob/main/anti_patterns/deny-warnings.md
2. https://www.reddit.com/r/rust/comments/f5xpib/psa_denywarnings_is_actively_harmful/

This occurs in basically all of the crates in the repo, in their `lib.rs` file.

**Possible Solution(s)**

Removing the `deny` statements, especially `deny(warnings)`. `deny(missing_docs)` should be turned into `warn(missing_docs)`. Same for clippy.

Licenser:
Moving this to a CI only task makes sense 👍 if you want throw in a PR otherwise we'll get to it eventually :)

marioortizmanero:
I know this is small but I honestly don't have that much time right now: I have to work on the PDK and I'm on finals. So I'd appreciate it if someone else took care of this if possible.

PR #1538, opened 2022-04-13
ClickHouse connector - sink

scrabsha:
# Pull request

## Description

This pull request adds a working version of the ClickHouse sink connector.

## Checklist

* [x] The code is tested


src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn connector_type(&self) -> ConnectorType {
+        "clickhouse".into()
+    }
+
+    async fn build(&self, alias: &str, raw_config: &ConnectorConfig) -> Result<Box<dyn Connector>> {
+        // So something bad happens here and nobody know what or why.
+        let config = raw_config
+            .config
+            .as_ref()
+            .ok_or_else(|| ErrorKind::MissingConfiguration(alias.to_string()))?;
```
mfelsche, line 33: such elegance

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn connection_url(&self) -> String {
+        let ClickhouseConfig {
+            host,
+            port,
+            database,
+            ..
+        } = &self.config;
+        let port = port.unwrap_or(ClickhouseConfig::DEFAULT_PORT);
+
+        format!("tcp://{host}:{port}/{database}?compression=lz4")
```
mfelsche, line 72: i am curious about the compression param. Is it required? Could we lift it to the connector configuration?

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,182 @@
...
+    async fn on_event(
+        &mut self,
+        _input: &str,
+        event: Event,
+        _ctx: &SinkContext,
+        _serializer: &mut EventSerializer,
+        _start: u64,
+    ) -> Result<SinkReply> {
+        debug!("on_event is getting called");
+        // TODO: is this the correct ErrorKind variant?
```
mfelsche, line 135: yes

src/connectors/tests/clickhouse.rs
```diff
@@ -0,0 +1,167 @@
...
+async fn simple_insertion() -> Result<()> {
+    let _ = env_logger::try_init();
+
+    let docker = clients::Cli::default();
+    let image = GenericImage::new(format!("{CONTAINER_NAME}:{CONTAINER_VERSION}"));
+
+    // We want to access the container from the host, so we need to make the
+    // corresponding port available.
+    let port_to_expose = Port {
+        local: SERVER_PORT,
```
mfelsche, line 49: just for the sake of your sanity, use our tooling in `free_port::find_free_tcp_port()` or use `#[serial_test]`, otherwise this will haunt you (or maybe us later on) forever. ;-)

mfelsche: You might even leave this out completely and the OS will pick a free port for you :)

src/connectors/tests/clickhouse.rs
```diff
@@ -0,0 +1,167 @@
...
+    // TODO: check for some kind of ack or something i guess?
+
+    info!("SUCCESSFULLY INSERTED DATA");
+
+    // Now that we have sent some events to the sink, let's check that
+    // everything was properly inserted in the database.
+
+    harness.stop().await?;
+
+    // Please don't judge me too much - this is a yolo test.
```
mfelsche, line 123: Haha! Don't worry :D

src/connectors/tests/clickhouse.rs
```diff
@@ -0,0 +1,167 @@
...
+        },
+    });
+
+    let harness = ConnectorHarness::new("clickhouse", &connector_config).await?;
+
+    let in_pipe = harness.get_pipe(IN).expect("No pipe connected to port IN");
+
+    harness.start().await?;
+    harness.wait_for_connected().await?;
+
```
mfelsche, line 85: I would advise you to add a call to ```rust

src/connectors/tests/clickhouse.rs
```diff
@@ -0,0 +1,167 @@
...
+    let block = client.query(request).fetch_all().await?;
+
+    let ages = block
+        .rows()
+        .map(|row| row.get::<u8, _>("age").map_err(Error::from))
+        .collect::<Result<Vec<_>>>()?;
+
+    info!("CHECKING EQUALITY");
+    assert_eq!(ages, [42, 101]);
+    info!("IT WORKED SOMEHOW!");
```
mfelsche, line 141: weird, isn't it?

src/lib.rs
```diff
@@ -14,7 +14,7 @@
 
 //! Tremor runtime
 
-#![deny(warnings)]
+// #![deny(warnings)]
```
mfelsche, line 17: :scream_cat:

src/connectors/tests/mod.rs
```diff
@@ -16,6 +16,10 @@
 //! Connector testing framework
 //!
 //! ....
+
+// TODO: what is this feature? Do we know what it does?
```
mfelsche, line 20: it tries to guard those tests being run with every call to `cargo test`, as they take long, spinning up docker and stuff.
They are all subsumed underneath the `integration` feature. See our `Cargo.toml`. You'd need to create a new flag for `clickhouse-integration` and add it to the activated features for `integration`, and finally guard this mod by `clickhouse-integration`.

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,182 @@
...
+        } = &self.config;
+        let port = port.unwrap_or(ClickhouseConfig::DEFAULT_PORT);
+
+        format!("tcp://{host}:{port}/{database}?compression=lz4")
+    }
+}
+
+#[derive(Deserialize)]
+struct ClickhouseConfig {
+    host: String,
```
Licenser, line 78: We've standardized to using a `url` for all connectors it would be good to do that here too - it's in `connectors::utils::url::Url`

src/connectors/tests/mod.rs
scrabsha: Alright, so `es` stands for *ElasticSearch*. That makes more sense now. Thanks :slightly_smiling_face:

src/connectors/tests/clickhouse.rs
scrabsha: Thanks for the tip. That explains why I had to stop my local, non-docker clickhouse server :sob:

scrabsha: This should be fixed by commit [2964597](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/296459726f49053020aaf3ba33178634be8a2a53). Feel free to reopen if some changes are still needed :)

src/lib.rs
Licenser: Not here directly but worth the contest, there have been discussions on removing `deny(warnings)` from code and instead of moving it to CI only to make refactoring / testing easier. That might be something we should consider

src/connectors/impls/clickhouse.rs
scrabsha: This means I have to add a new `TcpDefaults`, similar to [`HttpDefaults`](https://github.com/tremor-rs/tremor-runtime/blob/28d37fcf8adc71cefbc34c18bc204d7aaa27721c/src/connectors/utils.rs#L72), right?

Licenser: The `*Defaults` is used to allow omititng the protocol so that the system can auto-fill them. That's why we have a wrapper around `Url`. For `clickhouse` it seems `tcp` with a default port of `9000` would be a good default? that way people can just write `url = "localhost"` and it's equivalent to `url = "tcp://localhost:9000"` (I think I saw 9k as a default value in the code)

src/lib.rs
scrabsha: I removed this lint because I wanted to quietly start working. I do think it's a super good idea to guarantee no warning (from both `rustc` and `clippy`) on `main`, but it makes things hard when working with unfinished code.

Anyway, I promise i'll bring the lints back to errors once everything is implemented the way it should!

mfelsche: I'd say lets disable it here and enable it for CI.

src/connectors/impls/clickhouse.rs
scrabsha: @Licenser 

This *should* be fixed by [23b05f8](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/23b05f8980efa8abeda7c5db104f75fb822a4871), but i'm definitely not sure i did it the correct way; can you make sure I used the `Url` type the way you expected?

src/connectors/tests/clickhouse.rs
scrabsha: Fixed by [1187b8c](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/1187b8c29068f1325467137098fc89fbeeb14b7b)

src/lib.rs
Licenser: I'll make a ticket

Licenser: we already had a ticket :D https://github.com/tremor-rs/tremor-runtime/issues/1353 thanks @marioortizmanero  for bringing this up and ticketing it in the past!

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        let block = columns
+            .into_iter()
+            .fold(Block::new(), |block, (name, ty)| match ty {
+                ClickHouseType::UInt8 => add_uint8_column(block, name, &event),
+            });
+
+        client.insert("people", block.clone()).await?;
+
+        Ok(SinkReply::ACK)
```
mfelsche, line 168: As the sink is `auto_ack: true` you don't need to explicitly ACK. The runtime will handle this for you.

```suggestion
        Ok(SinkReply::NONE)
```

src/connectors/tests/mod.rs
scrabsha: This should be fixed with [66fe72f](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/66fe72f4b8be8f933d9e412f1776980c6b669681)

src/connectors/tests/clickhouse.rs
darach: https://github.com/tremor-rs/tremor-runtime/blob/main/src/connectors/tests/ws.rs#L547-L551

Perhaps adding a TestClickhouse client to drive `expect()` style tests as in the ws connector test would be helpful here?

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,308 @@
...
+
+                _ => todo!(),
+            };
+
+            Ok(expected_type.wrap_if_nullable(value_as_non_null))
+        }
+
+        (Value::Array(values), DummySqlType::Array(expected_inner_type)) => values
+            .iter()
+            .map(|value| clickhouse_value_of(value, expected_inner_type))
```
scrabsha, line 277: We don't check that all the elements of the array have the same type. Instead, we check that each element of the array is convertible to the expected array type.
I fail to see a situation where this could be a problem, but I am not sure it is the best way to proceed.

Licenser: That's probably not only fine but the better approach. For example we don't give users a easy way to differentiate between a u8 and a u64 so using "is the same type" instead of "converts to the same type" could end up causing surprising troubles :)

Cargo.toml
```diff
@@ -50,6 +50,9 @@ bimap = "0.6"
 byteorder = "1"
 bytes = "1.1"
 chrono = "0.4"
+clickhouse-rs = { git = "https://github.com/suharev7/clickhouse-rs", rev = "c61904a", default-features = false, features = [
```
scrabsha, line 53: Is it ok to use the most recent commit (as of today) from `clickhouse-rs` or should I use the most recent published version with a set of patches?

src/connectors/impls/clickhouse.rs
scrabsha: That makes complete sense, thanks!

src/errors.rs
```diff
@@ -313,5 +315,25 @@ error_chain! {
...
+            description("Expected object event")
+                display("Expected an object event, found a \"{found_type:?}\"")
+        }
+
+        MissingEventColumn(column: String) {
+            description("Missing event column")
+                display("Column \"{column}\" is missing")
+        }
+
+        UnexpectedEventFormat(column_name: String, expected_type: String, found_type: ValueType) {
```
scrabsha, line 334: I feel like it's a bad smell to type `expected_type` as a `String`. The reason why I did that is because using `DummySqlType` would require it to be public, which looks a bit overkill to me.
Any opinion on this?

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,290 @@
...
+pub(crate) struct ClickhouseSink {
+    db_url: String,
+    pool: Option<Pool>,
+    columns: Vec<(String, DummySqlType)>,
+}
+
+#[async_trait::async_trait]
+impl Sink for ClickhouseSink {
+    async fn connect(&mut self, _ctx: &SinkContext, _attempt: &Attempt) -> Result<bool> {
+        let pool = Pool::new(self.db_url.as_str());
```
mfelsche, line 137: Is there a way we can verify here in `connect` that the connection was successfully established?

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,290 @@
...
+        _start: u64,
+    ) -> Result<SinkReply> {
+        let mut client = self
+            .pool
+            .as_ref()
+            .ok_or_else(|| Error::from(ErrorKind::NoClickHouseClientAvailable))?
+            .get_handle()
+            .await?;
+
+        let (event_min_size, event_max_size) = event.value_iter().size_hint();
```
mfelsche, line 158: you can use `event.len()` to get the correct size :)

src/connectors/utils.rs
```diff
@@ -81,6 +81,14 @@ pub(crate) mod url {
         const HOST: &'static str = "localhost";
         const PORT: u16 = 443;
     }
+
+    pub(crate) struct ClickHouseDefaults;
```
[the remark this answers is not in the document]
mfelsche: For future refactoring's sake, it would be best to move it to the clickhouse connector file.

src/connectors/tests/clickhouse.rs
mfelsche: You could simply use a clickhouse client that checks in a loop until a certain timeout if the table already has the expected number of rows. If the table doesn't have the expected number within a certain timeout (say 10 seconds), then throw an error.

src/connectors/impls/clickhouse/conversion.rs
```diff
@@ -0,0 +1,442 @@
...
+
+        DummySqlType::Nullable(expected_inner_type) if value.is_null() => {
+            // Null value (when we expect it).
+            Ok(CValue::Nullable(Either::Left(
+                expected_inner_type.as_ref().into(),
+            )))
+        }
+
+        DummySqlType::Nullable(expected_inner_type) => {
+            // TODO: we don't want to `?` here because the expected type it
```
mfelsche, line 63: i think `?`-ing the error here is fine, isn't it?

src/errors.rs
mfelsche: This is fine. The fields inside those error types are only for conveying information to users, and you, as the error-type owner control how to use them. Also i think it is good practice, to keep the dependencies to other types low for those errors.

src/connectors/impls/clickhouse.rs
scrabsha: One solution would be to store a [`ClientHandle`] instead of storing a [`Pool`] in the `ClickhouseSink` structure. I think that's the best way to proceed.

[`ClientHandle`]: https://docs.rs/clickhouse-rs/1.0.0-alpha.1/clickhouse_rs/struct.ClientHandle.html
[`Pool`]: https://docs.rs/clickhouse-rs/1.0.0-alpha.1/clickhouse_rs/pool/struct.Pool.html

src/connectors/impls/clickhouse.rs
mfelsche: Just seen your implementation on size hint, this also makes a lot of sense. Lets keep it this way! :)

Cargo.toml
mfelsche: Lets track the github repo here until we have a release, maybe add a comment that a release is pending.

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,290 @@
...
+        format!("{url}?compression={compression}")
+    }
+}
+
+#[derive(Deserialize)]
+struct ClickhouseConfig {
+    url: Url<ClickHouseDefaults>,
+    compression: Option<Compression>,
+    columns: Vec<Column>,
+}
```
mfelsche, line 94: the table name is actually still missing here :)

src/connectors/impls/clickhouse.rs
scrabsha: This should be fixed by [d948725](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/d9487251771ab07d16b4caac11e9789ef8616ce0). Feel free to reopen if needed :)

src/connectors/impls/clickhouse.rs
scrabsha: This should be fixed as part of [d948725](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/d9487251771ab07d16b4caac11e9789ef8616ce0)

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,290 @@
...
+        let mut rslt = Vec::new();
+
+        let object = input
+            .as_object()
+            .ok_or_else(|| Error::from(ErrorKind::ExpectedObjectEvent(input.value_type())))?;
+
+        for (column_name, expected_type) in self.columns.iter() {
+            let cell = object.get(column_name.as_str()).ok_or_else(|| {
+                Error::from(ErrorKind::MissingEventColumn(column_name.to_string()))
+            })?;
```
scrabsha, line 189: We could just use null instead of returning an error

src/errors.rs
scrabsha: Perfect. Thanks for your feedback!

src/connectors/impls/clickhouse/conversion.rs
scrabsha: The "we must return an error if the inner conversion failed" is correct. The only problem is that the error will show something like: `Expected type T, found U`, while it should show `Expected type Nullable(T), found U`.

mfelsche: aaaaaah, got it. Makes sense

src/connectors/impls/clickhouse.rs
mfelsche: That would only work if the type is nullable, right? If not we would get a conversion error?

scrabsha: Correct; this is handled by the Tremor value -> clickhouse value function

src/connectors/impls/clickhouse.rs
scrabsha: Nice catch! This should be fixed by [1353320](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/135332017d87623f6d3b41296522d9518ad0ac70)

src/connectors/impls/clickhouse.rs
scrabsha: After sketching a few solutions, I see no simple way of doing this. Given how small the UX impact is, I think the best thing to do is to just leave it *as is*.

Cargo.toml
```diff
@@ -51,6 +51,22 @@ bimap = { version = "0.6", features = ["serde"] }
...
+# following:
+#
+# clickhouse-rs = { version = "...", default-features = false, features = [
+#   "async_std",
+# ] }
+#
+# Releases can be tracked at [1].
+#
+# [1]: https://crates.io/crates/clickhouse-rs/versions
+clickhouse-rs = { git = "https://github.com/suharev7/clickhouse-rs", branch = "async-await", default-features = false, features = [
```
Licenser, line 67: should we pin this to a ref? We don't control the fork so a branch could break the build at a random time if it gets updated in a unpleasant way.

test_clickhouse.troy
```diff
@@ -0,0 +1,52 @@
+define flow clickhouse
```
Licenser, line 1: this file should probably not be in the PR :)

Agares, line 1: I would put this as an example in the docs, but remove from the repo, I don't think we generally keep examples in the repo.

Agares (commented):
Great stuff! 🥳 🎉

src/connectors/tests/clickhouse/more_complex_test.rs
```diff
@@ -0,0 +1,404 @@
...
+    connectors::{
+        impls::clickhouse,
+        tests::{free_port, ConnectorHarness},
+    },
+    errors::{Error, Result},
+};
+
+const CONTAINER_NAME: &str = "clickhouse/clickhouse-server";
+const CONTAINER_VERSION: &str = "22.3.3.44";
+const SERVER_PORT: u16 = 9000;
```
Licenser, line 71: Would it make sense to move those one level up since they're shared between tests and currently redefined multiple times?

src/connectors/impls/clickhouse/conversion.rs
```diff
@@ -0,0 +1,457 @@
...
+
+        DummySqlType::Int32 => get_and_wrap(context, ValueAccess::as_i32, CValue::Int32),
+
+        DummySqlType::Int64 => get_and_wrap(context, ValueAccess::as_i64, CValue::Int64),
+
+        DummySqlType::String => get_and_wrap(context, ValueAccess::as_str, |s| {
+            CValue::String(Arc::new(s.as_bytes().to_vec()))
+        }),
+
+        DummySqlType::Ipv4 => convert_string_or_array(
```
Licenser, line 95: would it make sense to allow u32 encoded integers as IPs as well?

src/connectors/impls/clickhouse/conversion.rs
```diff
@@ -0,0 +1,457 @@
...
+        }),
+
+        DummySqlType::DateTime64Nanos => get_and_wrap(context, ValueAccess::as_i64, |timestamp| {
+            CValue::DateTime64(timestamp, (9, UTC))
+        }),
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+struct ConversionContext<'a, 'b, 'c, 'd> {
```
Licenser, line 142: we try to avoid single letter lifetimes as they make it quite hard to explain what each lifetime does, could we expand them to something more descriptive? (same for below)

src/connectors/impls/clickhouse/conversion.rs
```diff
@@ -0,0 +1,457 @@
...
+    column_name: &'a str,
+    value: &'b TValue<'c>,
+    expected_type: &'d DummySqlType,
+}
+
+fn wrap_getter_error<'a, 'b, F, T>(context: ConversionContext<'_, 'a, 'b, '_>, f: F) -> Result<T>
```
Licenser, line 148: Could we use more descriptive names for the Type parameters? especially in the functions below they're hard to make out where the `getter` is called `F` and the `wrapper` called `G` it makes the code quite a bit harder to understand

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,491 @@
...
+#[derive(Default, Debug)]
+pub(crate) struct Builder {}
+
+#[async_trait::async_trait]
+impl ConnectorBuilder for Builder {
+    fn connector_type(&self) -> ConnectorType {
+        "clickhouse".into()
+    }
+
+    async fn build(&self, alias: &str, raw_config: &ConnectorConfig) -> Result<Box<dyn Connector>> {
```
Licenser, line 37: we can use `build_cfg` here (it's new) and save the error case for a missing config

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,491 @@
...
+
+        format!("{host}/{path}?compression={compression}")
+    }
+}
+
+#[derive(Deserialize)]
+struct ClickhouseConfig {
+    url: Url<ClickHouseDefaults>,
+    compression: Option<Compression>,
+    database: Option<String>,
```
Licenser, line 99: Is it possible to not define a database and still get a working connector?

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,491 @@
...
+        let compression = self.config.compression.unwrap_or_default();
+
+        format!("{host}/{path}?compression={compression}")
+    }
+}
+
+#[derive(Deserialize)]
+struct ClickhouseConfig {
+    url: Url<ClickHouseDefaults>,
+    compression: Option<Compression>,
```
Licenser, line 98: Since we `unwrap_or_default` this field later, would it work to use `#[serde(default = Compression::default)]` here and mark it a non option?

src/connectors/impls/clickhouse.rs
mfelsche: Yes, clickhouse has a "default" database, you will write to if this isn't defined.

Licenser: Makes sense!

src/connectors/impls/clickhouse.rs
```diff
@@ -0,0 +1,491 @@
...
+    ) -> Result<Vec<(String, clickhouse_rs::types::Value)>> {
+        let mut rslt = Vec::new();
+
+        let object = input
+            .as_object()
+            .ok_or_else(|| Error::from(ErrorKind::ExpectedObjectEvent(input.value_type())))?;
+
+        for (column_name, expected_type) in columns.iter() {
+            // If the value is not present, then we can replace it by null.
+            const NULL: Value = Value::Static(StaticNode::Null);
```
Licenser, line 216: you can use `Value::const_null()` instead :) it's a const fn doing exactly this

Licenser (commented):
This is absolutely amazing work and a huge contribution! Incredible job :D! All the solutions are spot on and well thought out, clever without being obfuscating! Plus I love your coding style, it makes very good use of rust's features and tools!

Cargo.toml
scrabsha: As can be seen in the [`Cargo.lock`](https://github.com/scrabsha/tremor-runtime/blob/a3e16a2585b9963a0a46afd9e27dbe0b582d6f38/Cargo.lock#L1243) file, cargo defaulted it to the last commit of the `async-await` branch at the time I added the dependency. As a result, it won't "automatically update" if a commit is pushed in this branch.

From [the cargo references](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories):
> Once a git dependency has been added, Cargo will lock that dependency to the latest commit at the time. New commits will not be pulled down automatically once the lock is in place. However, they can be pulled down manually with cargo update.

That being said, I would totally understand if you prefer having a specific `rev` instead of a `branch` here :)

Cargo.toml
```diff
@@ -253,6 +269,7 @@ integration-docker = [
   "s3-integration",
   "kafka-integration",
   "gcp-integration",
+  "ch-integration",
```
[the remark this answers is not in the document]
scrabsha: Definitely! It is addressed by commit [0a2f3b2](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/0a2f3b287869a1d9875eac1d3c63a42c43b44656). Thanks

src/connectors/impls/clickhouse/conversion.rs
scrabsha: Thanks for pointing this out! I did some renaming and we able to remove two lifetimes (I was actually being overly generic :sweat_smile:) in commit 9a8f36b1.

Cargo.toml
Licenser: Oh, awesome :) learned something new! I think a `rev` might still be slightly better since I know at times we end up with a lock file so messed up from a merge :sob: that we delete and re-create it.

scrabsha: That's a very fair point. It should be fixed by commit b5082120.

src/connectors/tests/clickhouse/mod.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021, The Tremor Team
```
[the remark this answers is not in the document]
scrabsha: I hate that there's no way to consistently enforce such conventions :sob:
This is fixed by commit 03516222.

src/connectors/tests/clickhouse/more_complex_test.rs
scrabsha: It does. It is fixed by commit ccfb85c6.

src/connectors/impls/clickhouse/conversion.rs
scrabsha: Thank you for pointing this out. It should be better as of 32eda5c0.

src/connectors/impls/clickhouse.rs
scrabsha: Thanks for adding this function! Commit [8b41100](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/8b41100aab463fc4bf8aec3c36572579eedc1d36) should address this.

src/connectors/impls/clickhouse.rs
scrabsha: Nice catch. The resulting code (as of [75350df](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/75350dff63d363abf7d9f1f26b41eb2155775440)) is waaaaay cleaner.

src/connectors/impls/clickhouse.rs
scrabsha: Nice catch. It is fixed by commit [6ff650f](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/6ff650f9c429b085a2e708f12c9b6c76753c3ed5).

test_clickhouse.troy
scrabsha: I wrote a similar example in the docs. It was just a file I used when I was experimenting locally. It has been removed in [fb93fb7](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/fb93fb75452b35cfd008c5586faa7735d1dfdacd).

scrabsha: Correct! I removed it in [fb93fb7](https://github.com/tremor-rs/tremor-runtime/pull/1538/commits/fb93fb75452b35cfd008c5586faa7735d1dfdacd).

Licenser (approved):
🚀 great work!

src/connectors/impls/clickhouse/conversion.rs
```diff
@@ -0,0 +1,457 @@
...
+
+        DummySqlType::Nullable(expected_inner_type) if value.is_null() => {
+            // Null value (when we expect it).
+            Ok(CValue::Nullable(Either::Left(
+                expected_inner_type.as_ref().into(),
+            )))
+        }
+
+        DummySqlType::Nullable(expected_inner_type) => {
+            // TODO: we don't want to `?` here because the expected type it
```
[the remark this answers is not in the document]
scrabsha: It is

Sadly the resulting code would be super ugly. I think it's better to leave it *as is*, despite generating slightly useless error messages.

scrabsha:
After discussion with @mfelsche, we decided that this branch will be merged as is

I'll add a changelog entry at lunch time, so that the checklist is completed.

Licenser (dismissed):
🎸 Woooh! Great job :D well done!

Licenser (approved):

PR #1542, opened 2022-04-13
Improve merge behaviour

Licenser:
# Pull request

## Description

Ensure that we merge objects in a sensible way and not overwrite the target with a non object.

## Checklist

* [x] The code is tested


darach (dismissed):
LGTM - good catch 👍

tests/script_errors/bad_merge/script.tremor
```diff
@@ -0,0 +1,3 @@
+patch {"key": 42} of
```
mfelsche, line 1: the corresponding `error.txt` is empty. Does this mean this shouldn't error?

Licenser: Oh, no just that I forgot to delete the files, I had to move that to the runtime-errors check, patch isn't constant folded (to a degree makes sense to another it'd be nice to have it just because but not a can of worms I wanted to open pre 12.0)

mfelsche (approved):

PR #1566, opened 2022-04-21
Google Big Query Connector

Agares:
Signed-off-by: Ramona Luczkiewicz <rluczkiewicz@wayfair.com>

# Pull request

## Description
This of course is a very early prototype that does not do anything useful, except for writing the same row over and over.
You can get the token by running `gcloud auth print-access-token` (this will be replaced with proper google cloud auth).
```
define flow gbqtest
flow
    define pipeline passthrough
    pipeline
    end;

    define connector gbq from gbq
    with
        config = {
            "token": "...",
            "table_id": "projects/tremor/datasets/test/tables/streaming_test"
        }
    end;

    define connector metronome from metronome
    with
        config = {
            "interval": 5000000000
        }
    end;

    create connector metronome;
    create connector gbq;

    create pipeline passthrough;

    connect /connector/metronome to /pipeline/passthrough;
    connect /pipeline/passthrough to /connector/gbq;
end;

deploy flow gbqtest;
```

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/179)

## Checklist

* [x] The code is tested


src/connectors.rs
```diff
@@ -1178,6 +1178,7 @@ pub(crate) fn builtin_connector_types() -> Vec<Box<dyn ConnectorBuilder + 'stati
         Box::new(impls::http::server::Builder::default()),
         Box::new(impls::otel::client::Builder::default()),
         Box::new(impls::otel::server::Builder::default()),
+        Box::new(impls::gbq::Builder::default()),
```
darach, line 1181: Perhaps namespace this to `impls::gbq::writer::Builder::default()` as this will make adding `reader` orthogonal in due course?

Agares:
Left to do:
1. Properly handle authorization - we already depend on the `gouth` crate, so will use that
2. Handle disconnections/reconnections
3. Do not panic if the JSON types do not match the expected types - I think we should just treat this as an error, we can always add automated type conversions later, if a need arises, but it's hard to make changes to that logic once it's in place (without breaking backwards compatibility). Thoughts?

src/connectors/impls/gbq/sink.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+impl GbqSink {
+    pub async fn new(config: Config) -> Result<Self> {
+        let token = Token::new()?.header_value()?;
+
+        let token_metadata_value = MetadataValue::from_str(token.as_str())?;
+
+        let tls_config = ClientTlsConfig::new()
+            .ca_certificate(Certificate::from_pem(googapis::CERTIFICATES))
+            .domain_name("pubsub.googleapis.com");
```
mfelsche, line 240:
```suggestion
            .domain_name("bigquerystorage.googleapis.com");
```

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,443 @@
...
+struct JsonToProtobufMapping {
+    fields: HashMap<String, Field>,
+    descriptor: DescriptorProto,
+}
+
+fn map_field(
+    schema_name: &str,
+    raw_fields: &Vec<TableFieldSchema>,
+) -> (DescriptorProto, HashMap<String, Field>) {
+    let mut nested_types = vec![];
```
Licenser, line 75: It'd be nice to use Vec::with_capacity() here since we kind of know the size already - this will save allocations.

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,443 @@
...
+fn encode_field(val: &Value, field: &Field, result: &mut Vec<u8>) -> Result<()> {
+    let tag = field.tag;
+
+    // fixme check which fields are required and fail if they're missing
+    // fixme do not panic if the tremor type does not match
+    match field.table_type {
+        TableType::Double => prost::encoding::double::encode(
+            tag,
+            &val.as_f64().ok_or_else(|| {
+                ErrorKind::BigQueryTypeMismatch("f64", format!("{:?}", val.value_type()))
```
Licenser, line 184: Since the `BigQueryTypeMismatch`'s second argument seems to always be a value_type, would it be possible to use that as the 2nd argument for `BigQueryTypeMismatch` instead of converting it to a string each time?

Licenser: As a site note, perhaps a `try_as_*` function in the value_traits that returns the actual type as an error would be super nice I'll throw in a ticket for that in the value_trait repo

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,443 @@
...
+                }),
+                rows: Some(ProtoRows {
+                    serialized_rows: vec![mapping.map(event.data.parts().0)?],
+                }),
+            })),
+        };
+
+        let append_response = client
+            .append_rows(stream::iter(vec![request]))
+            .timeout(Duration::from_secs(10))
```
Licenser, line 347: hardcoding a timeout here seems to be risky. Would it be possible to omit it or make it configurable to omit / or select how long it is.

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,443 @@
...
+            ctx.notifier.connection_lost().await?;
+
+            return Ok(SinkReply::FAIL);
+        };
+
+        let mut append_response = append_response?.into_inner();
+
+        if let Ok(x) = append_response
+            .next()
+            .timeout(Duration::from_secs(10))
```
Licenser, line 362: see above

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,443 @@
...
+                    "The write stream is not available",
+                ))?;
+        let mapping = self
+            .mapping
+            .as_ref()
+            .ok_or(ErrorKind::BigQueryClientNotAvailable(
+                "The mapping is not available",
+            ))?;
+
+        let request = AppendRowsRequest {
```
Licenser, line 331: We will need to handle batching here, right now a batched (coming from the batch operator) request would fail the encoding. I think ES / HTTP have some good examples on how to best do this. It'd also allow multiple rows to be inserted at once via batching wich would be a nice feature.

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+            &val.as_bool()
+                .ok_or_else(|| ErrorKind::BigQueryTypeMismatch("bool", val.value_type()))?,
+            result,
+        ),
+        TableType::String
+        | TableType::Date
+        | TableType::Time
+        | TableType::Datetime
+        | TableType::Timestamp
+        | TableType::Numeric
```
mfelsche, line 205: shouldn't we also allow any number type for `Numeric` and `Bignumeric` too?

Agares: `Numeric` and `Bignumeric` are decimal-precision, so I'd avoid doing that, because if somebody tries to push a float/double into them, it's likely a (potentially costly) mistake.

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+            }
+        } else {
+            ctx.notifier.connection_lost().await?;
+
+            Ok(SinkReply::FAIL)
+        }
+    }
+
+    async fn connect(&mut self, _ctx: &SinkContext, _attempt: &Attempt) -> Result<bool> {
+        error!("Connecting to BigQuery");
```
mfelsche, line 384: add the `ctx` as first parameter so users will see in the logs which exact connector this log message belongs to.

```
error!("{ctx} Connecting to BigQuery");
```

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+        let token = Token::new()?.header_value()?;
```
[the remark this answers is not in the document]
Agares: oh, that's a good point, we should be passing the `Token` struct into the interceptor, and calling `header_value` on every request (it checks for expiration and refreshes as needed)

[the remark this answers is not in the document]
mfelsche: You'd need to put the `Token` into the `AuthInterceptor` and for each request call `Token::header_value` this will do the refreshing bits: https://docs.rs/gouth/0.2.1/src/gouth/token.rs.html#24-38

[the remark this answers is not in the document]
mfelsche: Didn't see your comment! :) We were thinking the same!

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+        let table_type =
+            if let Some(table_type) = table_field_schema::Type::from_i32(raw_field.r#type) {
+                warn!("Found a field of unknown type: {}", raw_field.name);
+                table_type
+            } else {
+                continue;
+            };
+
+        let grpc_type = match table_type {
+            table_field_schema::Type::Int64 => field_descriptor_proto::Type::Int64,
```
mfelsche, line 94:
```suggestion
            TableType::Int64 => field_descriptor_proto::Type::Int64,
```

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+                    "The write stream is not available",
+                ))?;
+        let mapping = self
+            .mapping
+            .as_ref()
+            .ok_or(ErrorKind::BigQueryClientNotAvailable(
+                "The mapping is not available",
+            ))?;
+
+        let mut serialized_rows = vec![];
```
mfelsche, line 330:
```suggestion
        let mut serialized_rows = Vec::with_capacity(event.len());
```

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+            AuthInterceptor {
+                token: token_metadata_value,
+            },
+        );
+
+        let write_stream = client
+            .create_write_stream(CreateWriteStreamRequest {
+                parent: self.config.table_id.clone(),
+                write_stream: Some(WriteStream {
+                    name: "".to_string(),
```
mfelsche, line 410: the name of the write stream should be configurable for the connector.

mfelsche: Lets add a comment that for the `createWriteStream` call only the `type` is important, all other info will be filled in to the `response`. Took a while until it made `click` in my head.

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+
+        Ok(result)
+    }
+
+    pub fn descriptor(&self) -> &DescriptorProto {
+        &self.descriptor
+    }
+}
+impl GbqSink {
+    pub async fn new(config: Config) -> Result<Self> {
```
mfelsche, line 291:
```suggestion
    pub fn new(config: Config) -> Result<Self> {
```

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+        let descriptor = map_field("table", vec);
+
+        Self {
+            descriptor: descriptor.0,
+            fields: descriptor.1,
+        }
+    }
+
+    pub fn map(&self, value: &Value) -> Result<Vec<u8>> {
+        let mut result = vec![];
```
mfelsche, line 274: it is usually more effective to pre-allocate a chunk, even if it is too big, to avoid reallocations while filling the buffer.
One thing we could do is to store the final capacity of the created vec and use this for the next one too, assuming the value sizes will remain roughly the same.

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -0,0 +1,660 @@
...
+            serialized_rows.push(mapping.map(data)?);
+        }
+
+        let request = AppendRowsRequest {
+            write_stream: write_stream.name.clone(),
+            offset: None,
+            trace_id: "".to_string(),
+            rows: Some(append_rows_request::Rows::ProtoRows(ProtoData {
+                writer_schema: Some(ProtoSchema {
+                    proto_descriptor: Some(mapping.descriptor().clone()),
```
mfelsche, line 342: this only needs to be provided upon the first request (after connect). We could store the full `ProtoSchema` in an `Option<ProtoSchema>` and extract it with a `self.proto_schema.take()` here and have it be a `None` all the time, saving 1 clone. :)
We could do the same with `trace_id` if we would have it configured on the connector.

mfelsche: interesting, very interesting... :oldmanshakesfistatcloud:

Agares: The underlying client is `tonic`, which is pretty good about handling long-lived connections, as far as I know. On the other hand - we're using HTTP here, so I expect the connection might not survive forever.

Licenser: Oh if it's HTTP not HTTP2 then that explains it, http will be counted as 1 connection per request, using http2/quick would probably solve that

src/connectors/impls/gbq/writer/sink.rs
darach: Add a comment to this effect with the reasoning here would be a good change

src/connectors/impls/gbq/writer/sink.rs
darach: We should be able to use grpc status codes for this purpose - perhaps thats something worth investigating and optimising in a separate ticket ( this would impact otel similarly and in future other grpc endpoints )

Licenser (approved):
🚀

darach (approved):
LGTM 🛩️

src/connectors/impls/gbq/writer.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::connectors::impls::gbq::writer::sink::GbqSink;
+use crate::connectors::prelude::*;
+use crate::connectors::{Connector, ConnectorBuilder, ConnectorConfig, ConnectorType};
+use serde::Deserialize;
+use tremor_pipeline::ConfigImpl;
+
+#[derive(Deserialize, Clone)]
+pub(crate) struct Config {
+    pub table_id: String,
+    pub connect_timeout: u64,
```
mfelsche, line 26: we have all timeouts in all the other connectors (and everywhere else) in nanoseconds. Would you mind changing these to nanos as well? I know seconds might be a little more applicable for such timeouts, but for the sake of consistency etc.

Licenser: Good catch!

Agares: Good point, seconds might be too much, I guess people might want to set timeouts which are in tens of ms. I hope some day we'll have some nice syntax for passing timeouts, because manually counting zeroes isn't fun :).
Anyway, it's done, switched to nanos.

mfelsche: We do have all the functions in `std::time::nanos` e.g. `from_seconds`

Licenser (approved):
🚀 great work!

PR #1582, opened 2022-04-29
Fix bug that tremor run would always ever run for exactly 150s

Licenser:
# Pull request

## Description

Tremor run had originally a timeout running for exactly 150s, this makes no sense of all, so we join world and let the runtime decide when to stop.

## Checklist

* [x] The code is tested

## Performance

no runtime changes


mfelsche (dismissed):
:facepalm: lets merge this quickly, so no one sees it.

mfelsche (approved):

mfelsche:
I think in this case the `codecov/patch` metric is not applicable. Those lines weren't tested before.

Would you mind adding a test for the `run_troy_source` function?

mfelsche:
Should we add a coverage exclude for `tremor-cli`?

Licenser:
idk that feels a bit "wrong" I want to try to see if we can use a wrapper to include more of the integration tests, that'd be really handy.

Licenser:
Okay I got it working but cargo-llvm-cov has a limitation that prevents us from using it: https://github.com/taiki-e/cargo-llvm-cov/issues/160

mfelsche (commented):
At least now we know that the tests in tremor-cli have been successful at one point in time.

mfelsche (approved):

PR #1609, opened 2022-05-04
CI Improvements - Publishing packages

PrimalPimmy:
# Pull request

## Description

- Added drop down menu feature on Draft-Release workflow which has options: major, minor, patch, rc
- Improved release process by no longer requiring to input old version and various other small changes
- All workspaces have the same version now
- Publish feature to crates.io (only tremor-common package for now)

## Checklist

* [x] The code is tested


.github/workflows/create-release.yaml
```diff
@@ -47,4 +41,13 @@ jobs:
           prerelease: false
           body: ${{ steps.extract-release-notes.outputs.release_notes }}
         env:
-          GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
\ No newline at end of file
+          GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
+      - name: Publish packages to crates.io
```
Licenser, line 45: I think the name here isn't quite correct

PrimalPimmy: what should it be?

.github/workflows/draft-release.yaml
```diff
@@ -0,0 +1,77 @@
+name: "Draft new release"
+
+on:
+  workflow_dispatch:
+    inputs:
+      new-version:
+        type: choice
+        description: "Which version you'd like to release?"
+        options:
+        - major
```
Licenser, line 10: This is fine for `-rc.2` but the workflow a;sp needs to allow to go from version -> version+1-rc.1 -...> version+1-rc.N -> version+1

Licenser (approved):
🚀 nice!

.github/workflows/draft-release.yaml
PrimalPimmy: I'll look into it in the next PR! thanks for pointing out!

PR #1620, opened 2022-05-06
Google PubSub - subscriptions

Agares:
# Pull request

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/184/files)

## Checklist

* [x] The code is tested


src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,191 @@
...
+                    .to_string(),
+                // fixme config?
+                max_messages: 1,
+                ..Default::default()
+            })
+            .await?
+            .into_inner();
+
+        if response.received_messages.len() == 0 {
+            return self.pull_data(pull_id, ctx).await;
```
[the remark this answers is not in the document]
Agares: I think I'll do it similarly to the way Kafka is doing it - spawn a background task that throws the messages on a channel.
0 messages does not meet anything, the API documentation isn't very clear on when does it happen, but it has happened to me when Tremor was running for a while with no new messages being published.

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,191 @@
...
+                // fixme use an error for GPubSub
+                "The client is not connected",
+            ))?;
+
+        let response = client
+            .pull(PullRequest {
+                subscription: "projects/wf-gcp-us-tremor-sbx/subscriptions/test-subscription-a"
+                    .to_string(),
+                // fixme config?
+                max_messages: 1,
```
mfelsche, line 115: Hmmm... the signature of `pull_data` only allows this to be 1, i see. Does this have an obvious performance impact? Should we maybe think about returning a `Result<Vec<SourceReply>>`? cc @Licenser

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,279 @@
...
+            }
+        };
+
+        *pull_id = ack_id;
+
+        Ok(SourceReply::Data {
+            origin_uri: EventOriginUri::default(),
+            data,
+            meta: None,
+            stream: None,
```
mfelsche, line 226: If gsub has the notion of a partition, we should somehow use 1 stream per partition, as we do in kafka. The reason is that we only have order inside of single streams.

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,279 @@
...
+                    )),
+                }),
+            },
+        );
+
+        let client = Arc::new(Mutex::new(client));
+
+        let (tx, rx) = async_std::channel::bounded(QSIZE.load(Ordering::Relaxed));
+
+        async_std::task::spawn(consumer_task(
```
mfelsche, line 179: keep track of the handle and delete any running one at the start of this function. Otherwise you end up having several consuming tasks running in concurrently.

src/connectors/impls/gpubsub/gsub.rs
Agares: There's no partitions in gpubsub, messages can have an ordering key, but then the ordering is handled internally by the platform (https://cloud.google.com/pubsub/docs/ordering), i.e. the messages received by sending PullRequest will always be in order. I don't think there's an easy way to split it into separate streams, since there's no pre-defined set of ordering keys, they can come and go.

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,279 @@
...
+) {
+    let mut ack_counter = 0;
+
+    loop {
+        let response_with_potential_timeout = client
+            .lock()
+            .await
+            .pull(PullRequest {
+                subscription: subscription_id.clone(),
+                max_messages: 1000,
```
mfelsche, line 87: do we maybe wanna move it to the config? Should this be smaller than the `sender` capacity?

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,279 @@
...
+    sender: Sender<AsyncTaskMessage>,
+    ack_ids: Arc<DashMap<u64, String>>,
+    subscription_id: String,
+    request_timeout: Duration,
+) {
+    let mut ack_counter = 0;
+
+    loop {
+        let response_with_potential_timeout = client
+            .lock()
```
mfelsche, line 83: i am afraid that this lock will block the `Source` from handling `ack/fail`. I think you can simply clone the `client` and pass a clone into this task, no? This would avoid the `Arc<Mutex<...>>` and leave the dashmap as the only synchronization point.

src/connectors/impls/gpubsub/gsub.rs
mfelsche: then lets just use `Some(DEFAULT_STREAM_ID)` for the time being

src/connectors/impls/gpubsub/gsub.rs
Agares: Turns out that the client can't be cloned, but it seems like a sensible workaround would be to create two clients using a cloned channel (according to tonic's docs, this will still use one connection handled in a background task)

src/connectors/impls/gpubsub/gsub.rs
Agares: I think setting this to QSIZE should be good enough here, I don't think this is something that users should have control of, as this really depends mostly on tremor internals and not outside factors

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,284 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+use tonic::codegen::InterceptedService;
+use tonic::transport::{Certificate, Channel, ClientTlsConfig};
+use tonic::Status;
+use tremor_pipeline::ConfigImpl;
+
+#[derive(Deserialize, Clone)]
+struct Config {
+    pub connect_timeout: u64,
```
mfelsche, line 23: Pls add the request_timeout as a config option.

mfelsche: And the `ack_deadline` which specifies after how many seconds an unacked message will be replayed.

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,284 @@
...
+
+        let client_background = connect_to_pubsub()?;
+
+        let (tx, rx) = async_std::channel::bounded(QSIZE.load(Ordering::Relaxed));
+
+        let join_handle = async_std::task::spawn(consumer_task(
+            client_background,
+            tx,
+            self.ack_ids.clone(),
+            self.config.subscription_id.clone(),
```
mfelsche, line 189: How should we treat subscriptions? Should we do a `GetOrCreate` operation (reuse an existing subscription or create one if none is there)?

src/connectors/impls/gpubsub/gsub.rs
Agares: The ack_deadline is set on the subscription, idk if we should actually create subscriptions, as that leaves some questions - what should we do if the subscription already exists? What should we do if it exists, but with different settings than we want?

src/connectors/impls/gpubsub/gsub.rs
Agares: Replied above :) https://github.com/tremor-rs/tremor-runtime/pull/1620#discussion_r871265835

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,284 @@
...
+    sender: Sender<AsyncTaskMessage>,
+    ack_ids: Arc<DashMap<u64, String>>,
+    subscription_id: String,
+    request_timeout: Duration,
+) {
+    let mut ack_counter = 0;
+
+    loop {
+        let response_with_potential_timeout = client
+            .pull(PullRequest {
```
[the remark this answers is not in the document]
Agares: The streaming call would maybe give us some savings, as we'd send less requests, but I'm not convinced it's worth the complexity (we'd likely need another async task to handle producing the request stream). There will be only one connection anyway, the only saving here is a bit less requests (but we still need to send requests with ACKs, so it's not that much)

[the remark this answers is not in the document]
Licenser: streaming also reduces latency which can be a significant impact, plus given the amount of messages gpubsub is likely to see savings do make some sense.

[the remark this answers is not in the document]
mfelsche: Agreed, might not be worth it.

src/connectors/impls/gpubsub/gsub.rs
mfelsche: yeah, lets do it like this and watch how it behaves.

src/connectors/impls/gpubsub/gsub.rs
mfelsche: Would this logic make sense?

If a subscription with the given name already exists, use it after verifying that the config is identical, if not throw an error, otherwise create one with the given config.

src/connectors/impls/gpubsub/gsub.rs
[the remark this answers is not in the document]
mfelsche: Just saw @Licenser commenting. Yeah, it might be worth it checking the diff, especially with high volumes.

Making this work would require to have 2 channels, one for acks into the background task and one from the background task to the source. This might help get rid of the `dashmap` as sync point and move the mapping from pull_id to ack_id into the background task.

But i am also happy with creating a separate issue for this enhancement and tackle it later. We are going to fetch `QSIZE` messages at a time, so latency might not be suffering too much.

src/connectors.rs
```diff
@@ -1176,6 +1177,7 @@ pub(crate) fn builtin_connector_types() -> Vec<Box<dyn ConnectorBuilder + 'stati
         Box::new(impls::otel::client::Builder::default()),
         Box::new(impls::otel::server::Builder::default()),
         Box::new(impls::gbq::writer::Builder::default()),
+        Box::new(impls::gpubsub::gsub::Builder::default()),
```
Licenser, line 1180: would it make sense to encapsulate that into a `consumer` sub module to mirror the kafka one?

src/lib.rs
```diff
@@ -39,6 +39,7 @@ extern crate lazy_static;
 #[macro_use]
 extern crate pretty_assertions;
 
+extern crate core;
```
mfelsche, line 42: why is this needed?

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,331 @@
...
+use tremor_pipeline::ConfigImpl;
+
+#[derive(Deserialize, Clone)]
+struct Config {
+    pub connect_timeout: u64,
+    pub ack_deadline: u64,
+    pub subscription_id: String,
+    #[serde(default = "default_endpoint")]
+    pub endpoint: String,
+    #[serde(default = "default_skip_authentication")]
```
Licenser, line 41: Is this used outside of tests? If not can we switch this to a #[cfg(test)] or something?

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,331 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+use tonic::codegen::InterceptedService;
+use tonic::transport::{Certificate, Channel, ClientTlsConfig};
+use tonic::Status;
+use tremor_pipeline::ConfigImpl;
+
+#[derive(Deserialize, Clone)]
+struct Config {
+    pub connect_timeout: u64,
```
mfelsche, line 36: should those settings be required for each connector? Or should they rather be an `Option<...>`?

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,331 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl Source for GSubSource {
+    async fn connect(&mut self, _ctx: &SourceContext, _attempt: &Attempt) -> Result<bool> {
+        let url = Url::<HttpsDefaults>::parse(self.config.endpoint.as_str())?;
```
mfelsche, line 187: this could/should be done in the `ConnectorBuilder` above. Then we get errors for an invalid config as early as possible.

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,331 @@
...
+
+type PubSubClient = SubscriberClient<InterceptedService<Channel, AuthInterceptor>>;
+type AsyncTaskMessage = Result<(u64, Vec<u8>)>;
+
+struct GSubSource {
+    config: Config,
+    client: Option<PubSubClient>,
+    receiver: Option<Receiver<AsyncTaskMessage>>,
+    ack_sender: Option<Sender<u64>>,
+    ack_ids: Arc<DashMap<u64, String>>,
```
mfelsche, line 85: i think you don't need this anymore here, do you? You could change the map in the `consumer_task` to be a simple hashmap, as it is contained in that task. Or would this make us lose `ack_ids` in case of a reconnect?

src/connectors/impls/gpubsub/gsub.rs
Agares: These can be made optional, my only question would be "what defaults are reasonable?"

src/connectors/impls/gpubsub/gsub.rs
Agares: yep, we would lose the ACK IDs, OTOH it's not necessairly a big issue, since then we'd just process the messages again, and as reconnects are rather rare, I don't think this would be an issue (we don't guarantee exactly once delivery anyway, as far as I know)

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,331 @@
...
+                };
+            }
+        };
+
+        *pull_id = ack_id;
+
+        Ok(SourceReply::Data {
+            origin_uri: EventOriginUri::default(),
+            data,
+            meta: None,
```
mfelsche, line 289: It would be super nice to have some metadata with the message `id`, the `ordering_key`, `publish_time` and the attributed from the [`PubsubMessage`](https://github.com/mechiru/googapis/blob/v0.6.0/googapis/genproto/google.pubsub.v1.rs#L409).

Similar to how it is done in kafka: https://github.com/tremor-rs/tremor-runtime/blob/main/src/connectors/impls/kafka/consumer.rs#L343

src/connectors.rs
Agares: It's already in the `gsub` submodule (and there will be `gpub` in another PR), but I can change the nomenclature to `consumer`/`producer`, I don't really have a strong opinion here, though I like `gsub`/`gpub` as connector names better than `gpubsub-consumer`/`gpubsub-producer`

Licenser: I think for recognizability and pettern repetition it's better to have a consumer/producer.

I've never heard to it referred to as `gpub` only as google pub/sub (gpubsub) so throwing in `gpub` and `gsub` will make discovery for users harder.

For the `producer/consumer` it's nice to have repeatable patterns for approximately equal services, pubsub, kafka, (whatever other message bus we add) if they all have a producer and consumer people can get used to that pattern, same with `client` and `server` or `reader` and `writer`

src/connectors/impls/gpubsub/gsub.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    let ack_ids_clone = ack_ids.clone();
+    let request_stream = async_stream::stream! {
+        yield StreamingPullRequest {
+            subscription: subscription_id.clone(),
+            ack_ids: vec![],
+            modify_deadline_seconds: vec![],
+            modify_deadline_ack_ids: vec![],
+            stream_ack_deadline_seconds: i32::try_from(ack_deadline.as_secs()).unwrap_or(10),
+            client_id: "".to_string(),
```
mfelsche, line 120: In kafka, we do assign our own client id: https://github.com/tremor-rs/tremor-runtime/blob/main/src/connectors/impls/kafka/consumer.rs#L92 should we do the same here?

src/connectors/impls/gpubsub/consumer.rs
```diff
@@ -0,0 +1,399 @@
...
+    10_000_000_000u64 // 10 seconds
+}
+
+#[derive(Debug, Default)]
+pub(crate) struct Builder {}
+
+#[async_trait::async_trait]
+impl ConnectorBuilder for Builder {
+    fn connector_type(&self) -> ConnectorType {
+        "gpubsub-consumer".into()
```
Licenser, line 75: Only a tiny thing, I prefer the `-` writing but we don't allow `-` in idents so anyone would have to escape this when using it, therefore we use `_` in the connector names (as ugly as it looks :sob:)
```suggestion
        "gpubsub_consumer".into()
```

darach:
LGTM - 🤞 the CI/CD gods will bless the tests!

src/connectors/impls/gpubsub/consumer.rs
```diff
@@ -0,0 +1,399 @@
...
+struct Config {
+    #[serde(default = "default_connect_timeout")]
+    pub connect_timeout: u64,
+    #[serde(default = "default_ack_deadline")]
+    pub ack_deadline: u64,
+    pub subscription_id: String,
+    #[serde(default = "default_endpoint")]
+    pub endpoint: String,
+    #[cfg(test)]
+    #[serde(default = "default_skip_authentication")]
```
Licenser, line 47: Is this used outside of tests? If not can we switch this to a #[cfg(test)] or something?

Agares: there's a `#[cfg(test)]`, line 46 :)

Licenser: never mind I'm blind, I looked at the old file :sob: sorry

Licenser (dismissed):

Licenser (approved):
🚀

src/connectors/impls/gpubsub/consumer.rs
```diff
@@ -0,0 +1,399 @@
...
+    attributes: HashMap<String, String>,
+) -> Value<'static> {
+    let mut attributes_value = Value::object_with_capacity(attributes.len());
+    for attr in attributes {
+        attributes_value
+            .as_object_mut()
+            .map(|x| x.insert(Cow::from(attr.0), Value::from(attr.1)));
+    }
+    literal!({
+        "pubsub_consumer": {
```
mfelsche, line 227:
```suggestion
        "gpubsub_consumer": {
```

src/connectors/impls/gpubsub/blue_green_hashmap.rs
```diff
@@ -0,0 +1,114 @@
...
+/// (now-expiration*2, now-expiration], therefore satisfying the minimum expiration time.
+///
+#[allow(unused)]
+pub struct BlueGreenDashMap<K: Send, V: Send> {
+    expiration: Duration,
+    hashmap_blue: (DashMap<K, V>, SystemTime),
+    hashmap_green: (DashMap<K, V>, SystemTime),
+}
+
+impl<K, V> BlueGreenDashMap<K, V>
```
darach, line 37: FlipFlopDataMop 🥳

darach (commented):
LGTM

src/connectors/impls/gpubsub/consumer.rs
```diff
@@ -0,0 +1,404 @@
...
+    mut client: PubSubClient,
+    client_id: String,
+    sender: Sender<AsyncTaskMessage>,
+    subscription_id: String,
+    ack_deadline: Duration,
+    ack_receiver: Receiver<u64>,
+) {
+    let mut ack_counter = 0;
+
+    let ack_ids = Arc::new(RwLock::new(BlueGreenDashMap::new(
```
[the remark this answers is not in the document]
mfelsche: We also don't need the `BlueGreenDashMap` to be based on `DashMap` and can instead base it on `HashMap`, this might lower the synchronization overhead and go even more brrrrr.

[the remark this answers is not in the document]
Agares: we do, because the async_stream generates it's own task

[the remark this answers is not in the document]
mfelsche: But one of those synchronization layers could go away, right? As we need to mess with those two maps, i guess it is the RwLock, right? So we could base it on two `halfbrown::HashMap`s, couldn't we?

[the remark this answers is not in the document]
Agares: good point, I switched to vanilla HashMap

mfelsche (dismissed):

src/connectors/impls/gpubsub/consumer.rs
```diff
@@ -0,0 +1,404 @@
...
+    ordering_key: String,
+    publish_time: Option<Duration>,
+    attributes: HashMap<String, String>,
+) -> Value<'static> {
+    let mut attributes_value = Value::object_with_capacity(attributes.len());
+    for attr in attributes {
+        attributes_value
+            .as_object_mut()
+            .map(|x| x.insert(Cow::from(attr.0), Value::from(attr.1)));
+    }
```
Licenser, line 231:
```suggestion
    for (name, value) in attributes {
        attributes_value
            .as_object_mut()
            .map(|x| x.insert(Cow::from(name), Value::from(value)));
    }
```

src/connectors/impls/gpubsub/consumer.rs
```diff
@@ -0,0 +1,404 @@
...
+
+            Ok(SubscriberClient::with_interceptor(
+                channel.clone(),
+                AuthInterceptor {
+                    token: Box::new(move || match token.header_value() {
+                        Ok(val) => Ok(val),
+                        Err(_) => Err(Status::unavailable(
+                            "Failed to retrieve authentication token.",
+                        )),
+                    }),
```
Licenser, line 287:
```suggestion
                    token: Box::new(move || token.header_value().map_err(|_| Err(Status::unavailable( "Failed to retrieve authentication token.")))),
```

Licenser (approved):

src/connectors/impls/gpubsub/consumer.rs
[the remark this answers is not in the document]
Licenser: Not for today but it'd be intersting to benchmark the difference, I suspect the DashMap will be faster then the RwLock but it doubtfully matters in this context, just curiosity :)

Licenser (approved):

PR #1632, opened 2022-05-10
Module based docgen (including troy modules)

mfelsche:
# Pull request

## Description

Generate stdlib docs from modules, also include troy modules now.

## Checklist

* [x] The code is tested


tests/flow_error.rs
```diff
@@ -53,7 +53,9 @@ macro_rules! test_cases {
 
                 match parse(&contents) {
                     Err(e) => {
-                        let got = Dumb::error_to_string(&e)?;
```
Licenser, line 56: Why do we spell out the highlighter generation?

tremor-cli/src/doc.rs
```diff
@@ -12,65 +12,167 @@
...
+use tremor_script::arena::Arena;
+use tremor_script::module::{Id, Module};
 
+fn push_line(line: &str, buf: &mut String) {
+    buf.push_str(line);
+    if !line.ends_with('\n') {
+        buf.push('\n');
+    }
+}
+#[allow(clippy::too_many_lines)]
```
Licenser, line 30: Could we refactor this to not require an allow?

tests/flow_error.rs
mfelsche: What exactly do you mean?

Licenser: this seems to unroll `error_to_string` and I wondered why

mfelsche: It covers the uncovered function `Deploy::format_error_with`. Sneaky, i know!

Licenser: alternatively, we could just `format_error_with`? it seems to be a copy of `format_error`. This feels like a harmful example of code for coverage's sake :(

Licenser: `format_error_with`, as far as I can tell, is never called anywhere so removing it seems the right thing to do.

src/connectors/impls/stdio.rs
```diff
@@ -34,6 +34,7 @@ lazy_static! {
             let mut stream = stdin();
             let mut buffer = [0_u8; INPUT_SIZE_BYTES];
             while let Ok(len) = stream.read(&mut buffer).await {
+                dbg!(len);
```
Licenser, line 37: weeeh I found a dbg :)
```suggestion
```

darach (approved):
LGTM 👍

Licenser (approved):
🚀

PR #1640, opened 2022-05-11
Remove coveralls upload

Licenser:
# Pull request

## Description

This removes the coveralls integration, we ran it in parallel with codecov and already switched to the CI demanding codecov instead of coveralls. We don't need this any longer and it makes PR's less spammy.

## Checklist

* [x] The code is tested

## Performance

no code changes

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -7,7 +7,7 @@ on:
       - main
 
 jobs:
-  tests-linux:
+  tests-linux:Ferror_to_string
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -7,7 +7,7 @@ on:
       - main
 
 jobs:
-  tests-linux:
+  tests-linux:Ferror_to_string
```
mfelsche, line 10:
```suggestion
  tests-linux:
```

mfelsche (approved):

.github/workflows/tests.yaml
Licenser: hah good catch! :D thanks.

darach (approved):
LGTM 📸

PR #1651, opened 2022-05-12
Fix bench connector not stopping

mfelsche:
# Pull request

## Description

The bench connector actually wasn't stopping when `stop_after_secs` wasn't configured, only `iters`.
Adapted it to stop after the expected number of events in that case.
Also added tests and switched it to not terminate the process directly, but to stop the world.

## Checklist

* [x] The code is tested

## Performance

Didn't test yet.

### DIFF
--- a/src/connectors/impls/bench.rs
+++ b/src/connectors/impls/bench.rs
@@ -15,16 +15,15 @@
 // #![cfg_attr(coverage, no_coverage)] // This is for benchmarking and testing
 
 use crate::connectors::prelude::*;
+use crate::errors::Kind as ErrorKind;
+use crate::system::{ShutdownMode, World};
 use hdrhistogram::Histogram;
+use std::io::{stdout, Write};
 use std::{
     cmp::min,
     fmt::Display,
     io::{BufRead as StdBufRead, BufReader, Read},
     time::Duration,
-};
-use std::{
-    io::{stdout, Write},
-    process,
 };
 use tremor_common::{file, time::nanotime};
 use xz2::read::XzDecoder;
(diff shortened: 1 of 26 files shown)


src/connectors/impls/bench.rs
```diff
@@ -247,38 +271,50 @@ impl Source for Blaster {
     }
 }
 
+#[derive(Clone, Debug)]
+struct StopAfter {
```
Licenser, line 275: Could this be Copy?

mfelsche: It could

src/connectors/impls/bench.rs
```diff
@@ -292,21 +328,13 @@ impl Sink for Blackhole {
...
-            if self.structured {
-                let v = self.to_value(2);
-                v.write(&mut stdout())?;
-            } else {
-                self.write_text(stdout(), 5, 2)?;
-            }
-            // ALLOW: This is on purpose, we use blackhole for benchmarking, so we want it to terminate the process when done
-            process::exit(0);
-        };
+        dbg!(&self.count);
```
Licenser, line 336: I found a dbg :D

src/connectors/impls/bench.rs
mfelsche: Should it though?

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -1109,7 +1110,9 @@ mod test {
             "config": {}
         });
 
-        let result = ConnectorHarness::new(function_name!(), "gbq", &config).await;
+        let result =
+            ConnectorHarness::new(function_name!(), &gbq::writer::Builder::default(), &config)
```
Licenser, line 1114: ohhh I love that!

src/connectors/tests/bench.rs
```diff
@@ -0,0 +1,110 @@
...
+};
+use tempfile::NamedTempFile;
+use tremor_common::ports::IN;
+use tremor_value::prelude::*;
+
+#[async_std::test]
+async fn stop_after_events() -> Result<()> {
+    let _ = env_logger::try_init();
+
+    let mut file = NamedTempFile::new()?;
```
Licenser, line 31: instead of creating a new file, could we use a fixture here?

mfelsche: We could, but i didnt like to depend on the current working directory while executing tests...

Licenser: I don't like creating files :D bu I think it comes down to "like" this is your PR so your like matters!

Licenser (approved):
🪨

Licenser (approved):
🚀

PR #1653, opened 2022-05-13
Raw payload support

### LINKED ISSUE #1652: `doc_as_upsert` support for ES update

Add support for `doc_as_upsert` for ES connector.

This is an additional key in the payload for `update` so it looks like:

`{"doc": {...}, "doc_as_upsert": true}`

This should be handled over metadata

**Notes**

see: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update

Licenser:
# Pull request

## Description

This adds `elastic.raw_payload` to the elastic connector. For `update` actions this will not wrap the body into `{"doc": ...}` but instead let the user provide the whole raw body with any additional items they desire.

## Related

* Related Issues: fixes #1652 

## Checklist

* [x] The code is tested

## Performance

The added overhead is one `if` which will not impact performance on an elastic search endpoint

### DIFF
--- a/src/connectors/tests/elastic.rs
+++ b/src/connectors/tests/elastic.rs
@@ -155,6 +155,71 @@ async fn connector_elastic() -> Result<()> {
         event.data.suffix().value()
     );
 
+    // upsert
+
+    let data = literal!({
+        "doc": {
+            "field1": 13.4,
+            "field2": "strong",
+            "field3": [false, true],
+            "field4": {
+                "nested": false
+            }
+        },
+        "doc_as_upsert": true,
+    });
+    let meta = literal!({
+        "elastic": {
+            "_index": "my_index",
+            "_id": "1234",
+            "action": "update",
+            "raw_payload": true
+        },
+    });
+    let event_not_batched = Event {
+        id: EventId::default(),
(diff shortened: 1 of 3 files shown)


src/connectors/tests/elastic.rs
```diff
@@ -155,6 +155,71 @@ async fn connector_elastic() -> Result<()> {
...
+                "result": "created",
+                "_id": "1234",
+                "_seq_no": 1,
+                "status": 201,
+                "_version": 1
+            }
+        }),
+        event.data.suffix().value()
+    );
+
```
mfelsche, line 222: It would be cool to actually verify that the `doc_as_upsert` works, by upserting another document with different fields and check that the resulting document is a mix/merge of the inserted and updated doc.

Licenser: It's verifying it by updating a document that doesn't exist, which update would fail, and succeeding. We're not validating content for `insert` either I don't think adding this requirement here makes sense.

mfelsche: Alrighty!

mfelsche (approved):

PR #1663, opened 2022-05-16
Add support for several elasticsearch auth options, TLS and more options

mfelsche:
# Pull request

## Description

New options for `elastic` connector:

- `headers`
- `timeout`
- `auth`
- `tls`

## Checklist

* [x] The code is tested

### DIFF
--- a/src/connectors/impls/elastic.rs
+++ b/src/connectors/impls/elastic.rs
@@ -13,38 +13,66 @@
 // limitations under the License.
 
 use std::fmt::Display;
+use std::time::Duration;
 
-use crate::connectors::prelude::*;
-use crate::connectors::sink::concurrency_cap::ConcurrencyCap;
-use crate::errors::{Error, Kind as ErrorKind, Result};
+use crate::{
+    connectors::{
+        impls::http::utils::Header, prelude::*, sink::concurrency_cap::ConcurrencyCap,
+        utils::tls::TLSClientConfig,
+    },
+    errors::{Error, Kind as ErrorKind, Result},
+};
 use async_std::channel::{bounded, Receiver, Sender};
-use elasticsearch::http::transport::{SingleNodeConnectionPool, TransportBuilder};
-use elasticsearch::http::Url;
-use elasticsearch::params::Refresh;
-use elasticsearch::{cluster::ClusterHealthParts, BulkDeleteOperation};
-use elasticsearch::{http::response::Response, Bulk};
-use elasticsearch::{BulkOperation, BulkOperations, BulkParts, Elasticsearch};
+use either::Either;
+use elasticsearch::{
+    cert::{Certificate, CertificateValidation},
+    cluster::ClusterHealthParts,
+    http::{
+        response::Response,
+        transport::{SingleNodeConnectionPool, TransportBuilder},
+        Url,
+    },
+    params::Refresh,
+    Bulk, BulkDeleteOperation, BulkOperation, BulkOperations, BulkParts, Elasticsearch,
+};
+use halfbrown::HashMap;
 use tremor_common::time::nanotime;
 use tremor_script::utils::sorted_serialize;
 use tremor_value::value::StaticValue;
 use value_trait::Mutable;
 
+use super::http::auth::Auth;
+
 #[derive(Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields)]
-pub struct Config {
+pub(crate) struct Config {
     /// list of elasticsearch cluster nodes
-    pub nodes: Vec<String>,
+    nodes: Vec<String>,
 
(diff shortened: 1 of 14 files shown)


src/connectors/impls/elastic.rs
```diff
@@ -54,6 +82,10 @@ fn default_concurrency() -> usize {
     DEFAULT_CONCURRENCY
 }
 
+fn default_auth() -> Auth {
```
Licenser, line 85: could we make Auth 'Default'? that way we avoid extra functions for this

src/connectors/impls/http/auth.rs
```diff
@@ -12,14 +12,30 @@
...
 /// Authorization methods
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(rename_all = "lowercase")]
 pub enum Auth {
     #[serde(alias = "basic")]
     Basic { username: String, password: String },
+    #[serde(alias = "bearer")]
+    Bearer(String),
+    #[serde(alias = "client_certificate")]
+    ClientCertificate {
```
Licenser, line 31: Would make sense to move this into `utils::tls` and keep the same naming conventions we have in other TLS related configs?

mfelsche: Why move it out of this enum? Or would you wrap this struct in a newtype and move the struct to tls?

Licenser: Ja I was thinking to create a struct fir client certificates, since they do exist for all TLS endpoints, not just HTTP, and wrap this here as a serde-flattened struct. Moving it to the `tls` utils will make it easier ti keep naming uniform through places, like not having `private_key` in one place and `key` in another

src/connectors/impls/http/auth.rs
```diff
@@ -12,14 +12,30 @@
...
+        /// path to a file containing one or more certificates in DER encoding that form a chain, ordered from leaf to root cert
+        certificate_chain: String,
+        /// path to a file containing the private key in either RSA or PKCS8 encoded form
+        private_key: String,
+    },
+    #[serde(alias = "elastic_api_key")]
```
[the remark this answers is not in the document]
Licenser: I'm worried that this will grow to a bunch of sink specific implementations that will end up being confusing.

src/connectors/impls/http/auth.rs
mfelsche: Yeah makes sense. Will do.

src/connectors/impls/http/auth.rs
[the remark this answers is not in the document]
Licenser: I mean if we decide that's what we want, 👍 , but lets not have that creep in pice by pice without a conscious decision to make it this.

src/connectors/impls/elastic.rs
mfelsche: Done

src/connectors/impls/http/auth.rs
mfelsche: Done, `Auth` is no solely for populating the `Authorization` header.

src/connectors/impls/http/auth.rs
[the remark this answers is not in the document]
mfelsche: I don't think this is too confusing. It is some special kind of auth that elastic pulled, and it actually is usable with http as well. I don't have a good idea how to refactor this one enum variant so it is easy to use and self-explanatory. I wouldn't want users to have to base64 their config on their own just for us having some leaner code. I would vote for keeping it this way.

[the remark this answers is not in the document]
Licenser: works 👍

Licenser (approved):
🚀

src/postprocessor.rs
```diff
@@ -61,6 +61,7 @@ pub fn lookup_with_config(config: &PostprocessorConfig) -> Result<Box<dyn Postpr
     match config.name.as_str() {
         "compress" => Ok(Box::new(Compress::from_config(config.config.as_ref())?)),
         "separate" => Ok(Box::new(separate::Separate::from_config(&config.config)?)),
+        "lines" => Ok(Box::new(separate::Separate::default())),
```
Licenser, line 64: This kind of doesn't belong in this PR and having it twice isn't really great.

Licenser (approved):
🚀 🇬🇵

Licenser (approved):
I shall not be dismissed like this! :joy: (or at least my approval shall not be dismissed :P)

mfelsche:
it is not working yet anyways. For some reason on CI the elasticsearch docker image is not coming up properly with http ssl settings :( Still investigating with some heavy printf debugging

Licenser (dismissed):
Not today github! Approved!

Licenser (approved):
Approval!

PR #1664, opened 2022-05-16
Fix for duplicate in k8s when using configmap

ekarlso:
This fixes the main issue described here which is not related to Helm or
k8s per say but a side-effect of how kubernetes mounts a configMap into
a path: https://github.com/tremor-rs/tremor-k8s-helm/issues/8

# Pull request

## Related

## Checklist

* [x] The code is tested


Licenser (changes requested):
We can't really do this, that'll magically outlaw some random filenames and will lead to pain down the road.

darach:
I think we can remove these generated folders via the helm charts:

https://stackoverflow.com/questions/68331724/kubernetes-volumemount-path-contains-timestamp

I'm going to try the above suggestion

Licenser:
Perhaps an option would be to ignore any hidden files? It's less arbitrary then "whatever k8s decided to name it on Monday".

docker/entrypoint.sh
```diff
@@ -39,7 +39,7 @@ else
 
 
     # Load *.troy files
-    TROYS=$(find ${CFG_DIR} -name '*.troy' -print 2>/dev/null | sort)
+    TROYS=$(find /etc/tremor/config/ -name '*.troy' | egrep -v '\.\.[0-9]{4}_[0-9]{2}_[0-9]{2}_[0-9]{2}_[0-9]{2}_[0-9]{2}' 2>/dev/null | sort)
```
Licenser, line 42: lets make this ignore all hidden files:
```suggestion
    TROYS=$(find /etc/tremor/config/ -not -path '*/.*' -type f -name '*.troy' -print 2>/dev/null | sort)
```

darach: 👍 This is cleaner and easier to document as well. @ekarlso would the above suggestion work in your case? If so we can test in the next release candidate...

ekarlso: Actually that doesnt fly, the "files" in kubernetes are symlinks rather :) just try it in one of any running it in a pod and you'll see :)

ekarlso:
Tried to update it to look for botjh File and for Link: 
find /etc/tremor/config/ -not -path '*/.*' \(-type l -o type f\) -name '*.troy' -print 2>/dev/null | sort
find /etc/tremor/config/ -not -path '*/.*' -type l -o type f -name '*.troy' -print 2>/dev/null | sort

But none of them are returning anyhing
https://www.unix.com/shell-programming-and-scripting/176162-bash-find-multiple-types-solved.html

darach (dismissed):
LGTM 🎉  - Thanks for the fix Endre!

docker/entrypoint.sh
Licenser, line 42: It seems like something went wrong with rebasing this is back to egrap :)

Licenser:
> Tried to update it to look for botjh File and for Link:
> ```
> find /etc/tremor/config/ -not -path '_/._' (-type l -o type f) -name '_.troy' -print 2>/dev/null | sort
> find /etc/tremor/config/ -not -path '_/._' -type l -o type f -name '_.troy' -print 2>/dev/null | sort
> ```
> 
> But none of them are returning anyhing https://www.unix.com/shell-programming-and-scripting/176162-bash-find-multiple-types-solved.html

there seem to be some odd `_` in there, have you given this a try:
```
find /etc/tremor/config/ -not -path '*/.*' -type f,l -name '*.troy' -print 2>/dev/null | sort
```

Licenser (approved):
you 🚀 ! Thanks :)

PR #1666, opened 2022-05-17
Allow 0.1% coverage drop to handle property based tests

Licenser:
# Pull request

## Description

The property based tests can be mean and stop covering a line they covered before randomly, this means we have some variation coverage. This PR should allow for that to make everyone's life easier, so I hope.

## Related

* every dependabot pr ever
* Related Issues: I got so many issues

## Checklist

* [x] The code is tested

## Performance

no code change

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,25 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+  status: # see https://docs.codecov.com/docs/commit-status
+    project:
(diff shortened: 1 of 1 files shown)


mfelsche (approved):
I don't know what the other sections here do, but lets give it a try. I am all in for that threshold!

Licenser:
neither do I :D they are the [default config](https://docs.codecov.com/docs/codecov-yaml#default-yaml) so I put them in as I am not sure, and couldn't find if the yaml i merged with the default or replaces it

PR #1670, opened 2022-05-18
Remove ack-on-drop and ack-on-dead-end.

mfelsche:
Those behaviours were producing edge cases where error events (from source connectors) were acked, though not handled and thus acked the original failed event.

# Pull request

## Checklist

* [x] The code is tested

## Performance
The performance should be back to where it was before, before we introduced `ack-on-drop` etc.

### DIFF
--- a/tremor-pipeline/src/op/trickle/script.rs
+++ b/tremor-pipeline/src/op/trickle/script.rs
@@ -62,18 +62,6 @@ impl Operator for Script {
             }
         });
 
-        Ok(if let Some(port) = port {
-            EventAndInsights {
-                events: vec![(port, event)],
-                ..EventAndInsights::default()
-            }
-        } else if event.transactional {
-            EventAndInsights {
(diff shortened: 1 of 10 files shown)


Licenser:
On a quick glance I didn't see it, does this also un-ack's `drop`?

Licenser (approved):
🚀 yes it does I was just blind :D

tremor-pipeline/src/op/trickle/script.rs
```diff
@@ -62,18 +62,6 @@ impl Operator for Script {
             }
         });
 
-        Ok(if let Some(port) = port {
-            EventAndInsights {
-                events: vec![(port, event)],
-                ..EventAndInsights::default()
-            }
-        } else if event.transactional {
```
mfelsche, line 70: by removing this we do not ack on `drop` in scripts

PR #1717, opened 2022-06-03
Fix race condition during quiescence.

mfelsche:
The spawn_task function was issueing a connection_lost so that the connector state went to Disconnected and
the source stopped pulling data from the source, ergo it wasn't getting sourcereplies about ending streams
and thus was never finishing quiescence but always timing out after 5 seconds.

This shaves off > 20s of integration test time.

# Pull request

## Checklist

* [x] The code is tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,6 @@ tremor-value/proptest-regressions/
 *.release
 tremor-script/current_counterexample.eqc
 /docs
-lalrpop-docgen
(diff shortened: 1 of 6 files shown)


Licenser (approved):

Licenser (approved):

Licenser (approved):

mfelsche:
:partying_face:

PR #1718, opened 2022-06-03
Stop integration tests after 5 minutes

mfelsche:
Stop integration tests after 5 minutes and print their log files for better debuggability on CI and to avoid infinite runs hogging machines.

## Checklist

* [x] The code is tested

### DIFF
--- a/tremor-cli/src/main.rs
+++ b/tremor-cli/src/main.rs
@@ -22,6 +22,7 @@
     clippy::unnecessary_unwrap,
     clippy::pedantic
 )]
+#![cfg_attr(coverage, feature(no_coverage))]
 
 #[macro_use]
 extern crate serde_derive;
--- a/tremor-cli/src/main.rs
+++ b/tremor-cli/src/main.rs
@@ -22,8 +22,6 @@
     clippy::unnecessary_unwrap,
     clippy::pedantic
 )]
-#![cfg_attr(coverage, feature(no_coverage))]
-
 #[macro_use]
 extern crate serde_derive;
 
--- a/tremor-cli/src/test.rs
+++ b/tremor-cli/src/test.rs
@@ -160,7 +160,6 @@ async fn suite_integration(
     }
 }
 
-#[cfg_attr(coverage, no_coverage)]
 async fn run_integration(
     root: &Path,
     config: &TestConfig,
--- a/tremor-cli/src/test.rs
+++ b/tremor-cli/src/test.rs
@@ -43,8 +43,6 @@ pub mod stats;
 pub mod tag;
 mod unit;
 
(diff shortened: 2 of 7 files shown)


tremor-cli/src/test.rs
```diff
@@ -178,7 +183,33 @@ async fn run_integration(
         status::tags(&tags, Some(&matched), Some(&config.excludes))?;
 
         // Run integration tests
-        let test_report = process::run_process("integration", base, root, &tags).await?;
+        let test_report = match process::run_process("integration", base, root, &tags)
```
Licenser, line 186: could we make this a command-line argument instead of a hard-coded kill? I worry this will bite us at one point. Something like `tremor test --timeout-secs 300` ...

Licenser (approved):

tremor-cli/src/main.rs
```diff
@@ -22,6 +22,7 @@
     clippy::unnecessary_unwrap,
     clippy::pedantic
 )]
+#![cfg_attr(coverage, feature(no_coverage))]
```
Licenser, line 25: 👀

Licenser (dismissed):

tremor-cli/src/main.rs
mfelsche: See: https://github.com/taiki-e/cargo-llvm-cov#exclude-function-from-coverage

mfelsche: Damn, we can't use this on stable rust. :sob:

Licenser (dismissed):

Licenser (dismissed):

tremor-cli/src/test.rs
```diff
@@ -178,7 +181,37 @@ async fn run_integration(
...
+        let test_report = if let Some(timeout) = config.timeout {
+            match process::run_process("integration", base, root, &tags)
+                .timeout(timeout)
+                .await
+            {
+                Err(_) => {
+                    // timeout
+                    // walk the entire dir and print all *.log files for debuggability
+                    error!("Timeout running integration test {}", root.display());
+                    let walker = GlobWalkerBuilder::new(root, "*.log")
```
Licenser, line 193: I don't think we need this, if the test fails we upload the log files in CI, removing the printer could solve the coverage?

mfelsche: yeah, makes sense. i didn't see that we store the log files. It should be enough to error out here. Good point!

Licenser: It's well hidden, and I think we enver used it so it's a secret :) shhhh don't tell anyone!

Licenser (dismissed):

Licenser (approved):

PR #1719, opened 2022-06-03
Add the Google PubSub Publisher

Agares:
# Pull request

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/204)

## Checklist

* [x] The code is tested

### DIFF
--- a/src/connectors/impls/gpubsub/producer.rs
+++ b/src/connectors/impls/gpubsub/producer.rs
@@ -185,8 +185,8 @@ impl Sink for GpubSink {
 
         for (value, meta) in event.value_meta_iter() {
             for payload in serializer.serialize(value, event.ingest_ns)? {
-
-                let ordering_key = ctx.extract_meta(meta)
+                let ordering_key = ctx
+                    .extract_meta(meta)
                     .get("ordering_key")
                     .as_str()
                     .map_or_else(|| "".to_string(), ToString::to_string);
--- a/src/connectors/impls/gpubsub/producer.rs
+++ b/src/connectors/impls/gpubsub/producer.rs
@@ -6,11 +6,14 @@ use crate::connectors::sink::Sink;
 use crate::connectors::utils::url::HttpsDefaults;
 use crate::connectors::{
     CodecReq, Connector, ConnectorBuilder, ConnectorConfig, ConnectorContext, ConnectorType,
+    Context,
 };
 use crate::errors::Result;
+use async_std::prelude::FutureExt;
 use googapis::google::pubsub::v1::publisher_client::PublisherClient;
 use googapis::google::pubsub::v1::{PublishRequest, PubsubMessage};
 use gouth::Token;
+use std::collections::HashMap;
 use std::time::Duration;
 use tonic::codegen::InterceptedService;
 use tonic::transport::{Certificate, Channel, ClientTlsConfig};
@@ -21,6 +24,8 @@ use tremor_pipeline::{ConfigImpl, Event};
 pub struct Config {
     #[serde(default = "default_connect_timeout")]
     pub connect_timeout: u64,
+    #[serde(default = "default_request_timeout")]
+    pub request_timeout: u64,
     #[serde(default = "default_endpoint")]
     pub endpoint: String,
     pub topic: String,
@@ -31,6 +36,9 @@ fn default_endpoint() -> String {
 }
 fn default_connect_timeout() -> u64 {
     1_000_000_000u64 // 1 second
+}
+fn default_request_timeout() -> u64 {
+    10_000_000_000u64 // 10 seconds
 }
 
 impl ConfigImpl for Config {}
@@ -112,7 +120,7 @@ impl Sink for GpubSink {
         let token = Token::new()?;
 
         let client = PublisherClient::with_interceptor(
-            channel.clone(),
+            channel,
             AuthInterceptor {
                 token: Box::new(move || {
                     token.header_value().map_err(|_| {
@@ -131,7 +139,7 @@ impl Sink for GpubSink {
         &mut self,
         _input: &str,
         event: Event,
-        _ctx: &SinkContext,
+        ctx: &SinkContext,
         serializer: &mut EventSerializer,
         _start: u64,
     ) -> Result<SinkReply> {
@@ -144,23 +152,41 @@ impl Sink for GpubSink {
 
         for e in event.value_iter() {
             messages.push(PubsubMessage {
-                data: serializer.serialize(e, event.ingest_ns).unwrap()[0].clone(),
-                attributes: Default::default(),
+                data: serializer.serialize(e, event.ingest_ns)?[0].clone(),
+                attributes: HashMap::new(),
                 message_id: "".to_string(),
                 publish_time: None,
                 ordering_key: "".to_string(),
             });
         }
 
-        client
+        if let Ok(inner_result) = client
             .publish(PublishRequest {
                 topic: self.config.topic.clone(),
                 messages,
             })
+            .timeout(Duration::from_nanos(self.config.request_timeout))
             .await
-            .unwrap();
+        {
+            if let Err(error) = inner_result {
+                error!("Failed to publish a message: {}", error);
+                ctx.swallow_err(
+                    ctx.notifier.connection_lost().await,
+                    "Failed to notify about PubSub connection loss",
+                );
 
-        Ok(SinkReply::ACK)
+                Ok(SinkReply::FAIL)
+            } else {
+                Ok(SinkReply::ACK)
+            }
+        } else {
+            ctx.swallow_err(
+                ctx.notifier.connection_lost().await,
+                "Failed to notify about PubSub connection loss",
+            );
+
+            Ok(SinkReply::FAIL)
+        }
     }
 
     fn auto_ack(&self) -> bool {
--- a/src/connectors/impls/gpubsub/producer.rs
+++ b/src/connectors/impls/gpubsub/producer.rs
@@ -28,6 +28,7 @@ use googapis::google::pubsub::v1::publisher_client::PublisherClient;
 use googapis::google::pubsub::v1::{PublishRequest, PubsubMessage};
 use gouth::Token;
 use std::collections::HashMap;
+use std::sync::Arc;
 use std::time::Duration;
 use tonic::codegen::InterceptedService;
 use tonic::transport::{Certificate, Channel, ClientTlsConfig};
@@ -43,6 +44,8 @@ pub struct Config {
     #[serde(default = "default_endpoint")]
     pub endpoint: String,
     pub topic: String,
+    #[cfg(test)]
+    pub skip_authentication: bool,
 }
 
 fn default_endpoint() -> String {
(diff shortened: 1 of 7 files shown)


src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,213 @@
...
+    async fn connect(&mut self, _ctx: &SinkContext, _attempt: &Attempt) -> Result<bool> {
+        let mut channel = Channel::from_shared(self.config.endpoint.clone())?
+            .connect_timeout(Duration::from_nanos(self.config.connect_timeout));
+        if self.url.scheme() == "https" {
+            let tls_config = ClientTlsConfig::new()
+                .ca_certificate(Certificate::from_pem(googapis::CERTIFICATES))
+                .domain_name(
+                    self.url
+                        .host_str()
+                        .ok_or_else(|| Status::unavailable("The endpoint is missing a hostname"))?
```
Licenser, line 126: could we validate this at definition time?

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,213 @@
...
+        let client = self.client.as_mut().ok_or(ErrorKind::ClientNotAvailable(
+            "PubSub",
+            "The publisher is not connected",
+        ))?;
+
+        let mut messages = vec![];
+
+        for e in event.value_iter() {
+            messages.push(PubsubMessage {
+                data: serializer.serialize(e, event.ingest_ns)?[0].clone(),
```
Licenser, line 169: We don't allow `[]` access in hot code since it can panic

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,234 @@
...
+        ctx: &SinkContext,
+        serializer: &mut EventSerializer,
+        _start: u64,
+    ) -> Result<SinkReply> {
+        let client = self.client.as_mut().ok_or(ErrorKind::ClientNotAvailable(
+            "PubSub",
+            "The publisher is not connected",
+        ))?;
+
+        let mut messages = vec![];
```
mfelsche, line 186:
```suggestion
        let mut messages = Vec::with_capacity(event.len());
```

src/connectors/impls/gpubsub/producer.rs
mfelsche: Yeah. I think it is safe to concat all the vecs from a single value and put it into 1 PubSubMessage.

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        for e in event.value_iter() {
+            messages.push(PubsubMessage {
+                data: serializer.serialize(e, event.ingest_ns)?[0].clone(),
+                attributes: HashMap::new(),
+                message_id: "".to_string(),
```
mfelsche, line 192: this and the ordering_key should come from the event metadata, if available.

src/connectors/impls/gpubsub/producer.rs
Licenser: Ja, I think that's what we do in kafka isn't it? It should probably act the same.

Agares: Oh, dang, thank you for reminding me. What does it actually mean when there's more than 1 result from the serializer? It seems like Kafka is actually sending each as a separate message.

src/connectors/impls/gpubsub/producer.rs
Agares: The `message_id` is generated on the server side (I'll add a comment in the code). I'll get the `ordering_key` from the metadata though

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,234 @@
...
+                publish_time: None,
```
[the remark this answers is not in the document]
Agares: yep, the value passed here is ignored

src/connectors/impls/gpubsub/producer.rs
Agares: Okay, sounds good, I'll do it this way.
BTW Unless I'm reading the code incorrectly, kafka will issue a separate message per payload: https://github.com/tremor-rs/tremor-runtime/blob/main/src/connectors/impls/kafka/producer.rs#L259-L294

Licenser: You are totally right, then we should probably keep it that way

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,284 @@
...
+            "PubSub",
+            "The publisher is not connected",
+        ))?;
+
+        let mut messages = Vec::with_capacity(event.len());
+
+        for (value, meta) in event.value_meta_iter() {
+            for payload in serializer.serialize(value, event.ingest_ns)? {
+                let ordering_key = meta
+                    .get("gpubsub_producer")
```
mfelsche, line 189: you can use `ctx.extract_meta(meta)` to get the inner meta if you like.

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,284 @@
...
+                    .map_or_else(|| "".to_string(), ToString::to_string);
+
+                messages.push(PubsubMessage {
+                    data: payload,
+                    attributes: HashMap::new(),
+                    // publish_time and message_id will be ignored in the request and set by server
+                    message_id: "".to_string(),
+                    publish_time: None,
+
+                    // fixme get from the metadata
```
mfelsche, line 201:
```suggestion
                    // from the metadata
```

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,284 @@
...
+                    ctx.notifier.connection_lost().await,
+                    "Failed to notify about PubSub connection loss",
+                );
+
+                Ok(SinkReply::FAIL)
+            } else {
+                Ok(SinkReply::ACK)
+            }
+        } else {
+            ctx.swallow_err(
```
mfelsche, line 227: Do we always want to assume a lost connection if we hit a timeout? Or should we simply fail? Calling the `connection_lost()` method will have the effect that the circuit breaker will be triggered and we will attempt a reconnect. If we just fail on the other hand, and we hit repeated timeouts, we might never reconnect.

Maybe we should discuss a golden strategy here that we can/should apply to all sink-connectors?

We do have the `qos::backpressure` and `qos::percentile` operators who can be used to either stop upstreams from sending using the circuit-breaker or just drop some messages.

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,284 @@
...
+
+        if let Ok(inner_result) = client
+            .publish(PublishRequest {
+                topic: self.config.topic.clone(),
+                messages,
+            })
+            .timeout(Duration::from_nanos(self.config.request_timeout))
+            .await
+        {
+            if let Err(error) = inner_result {
```
mfelsche, line 215: Is it known what possible errors we can get and could we use this info to only call `connection_lost()` when we know something is wrong with the connection, and not e.g. the message being to big or something?

src/connectors/tests/gpubsub/gpub.rs
```diff
@@ -0,0 +1,296 @@
...
+        testcontainers::images::google_cloud_sdk_emulators::CloudSdk::pubsub();
+    let runnable_image = RunnableImage::from((pubsub, pubsub_args));
+    let container = runner.run(runnable_image);
+
+    let port = container
+        .get_host_port_ipv4(testcontainers::images::google_cloud_sdk_emulators::PUBSUB_PORT);
+    let endpoint = format!("http://localhost:{}", port);
+
+    let mut env = EnvHelper::new();
+    env.remove_var("GOOGLE_APPLICATION_CREDENTIALS");
```
mfelsche, line 93: This test might fail if we run this test on a gcloud VM or there is a credentials file somewhere where `gcloud` expects it. According to the documentation here: https://crates.io/crates/gouth Is this correct?

src/connectors/tests/gpubsub/gpub.rs
```diff
@@ -0,0 +1,296 @@
...
+        iter_count += 1;
+    }
+
+    if received_messages.len() != 100 {
+        panic!(
+            "Received an unexpected number of messages - {}",
+            received_messages.len()
+        );
+    }
+
```
mfelsche, line 211: don't forget to stop the harness

src/connectors/tests/gpubsub/gpub.rs
```diff
@@ -0,0 +1,296 @@
...
+        ..Event::default()
+    };
+    harness.send_to_sink(event, IN).await?;
+    harness
+        .wait_for_state(State::Failed)
+        .timeout(Duration::from_secs(10))
+        .await?
+        .unwrap();
+
+    Ok(())
```
mfelsche, line 295: same here, pls stop the harness :scream_cat:

src/connectors/tests/gpubsub/gpub.rs
Agares: true,  though I'm not sure if there's a foolproof way to enforce a failure to find the token, so maybe it's a risk worth taking

mfelsche: You could, at the start of the test try to get a token manually and only run the test and the asserts in it if this was failing, otherwise "skip" it with just an `Ok(())`.

Agares: But I'm afraid that if we do that, then if the tests end up running on GCP, it will not work and just go unnoticed. It will fails, it will be mildly annoying, but at least we will know that action needs to be taken.

mfelsche: Ok, lets add a hook that is calling you whenever this test fails. :D

src/connectors/impls/gpubsub/producer.rs
mfelsche: I would say we should just `FAIL` here and not call `connection_lost`. But pls leave a comment that this might need fixing.

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,298 @@
...
+        if self.url.scheme() == "https" {
+            let tls_config = ClientTlsConfig::new()
+                .ca_certificate(Certificate::from_pem(googapis::CERTIFICATES))
+                .domain_name(self.hostname.clone());
+
+            channel = channel.tls_config(tls_config)?;
+        }
+
+        let channel = channel.connect().await?;
+        #[allow(unused_assignments, unused_mut)]
```
Licenser, line 135: I really find this super confusing to read, could we re-structure this part to, for example, have a create_publisher() function that has one implementation in `test` and one in `not(test)` to make it clear what code is executed when?

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,298 @@
...
+                    Code::Aborted,
+                    Code::Cancelled,
+                    Code::DataLoss,
+                    Code::DeadlineExceeded,
+                    Code::Internal,
+                    Code::ResourceExhausted,
+                    Code::Unavailable,
+                    Code::Unknown,
+                ]
+                .contains(&error.code())
```
Licenser, line 228: We don't need to perform an array traversal for this

```suggestion
                if matches!(error.code(),
                    Code::Aborted |
                    Code::Cancelled |
                    Code::DataLoss |
                    Code::DeadlineExceeded |
                    Code::Internal |
                    Code::ResourceExhausted |
                    Code::Unavailable |
                    Code::Unknown
                )
```

Licenser (dismissed):

src/connectors/impls/gpubsub/producer.rs
mfelsche: this is the last thing that needs fixing

mfelsche (dismissed):

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -0,0 +1,297 @@
...
+pub struct Config {
+    #[serde(default = "crate::connectors::impls::gpubsub::default_connect_timeout")]
+    pub connect_timeout: u64,
+    #[serde(default = "crate::connectors::impls::gpubsub::default_request_timeout")]
+    pub request_timeout: u64,
+    #[serde(default = "crate::connectors::impls::gpubsub::default_endpoint")]
+    pub endpoint: String,
+    pub topic: String,
+    #[cfg(test)]
+    #[serde(default = "crate::connectors::impls::gpubsub::default_skip_authentication")]
```
Licenser, line 48: from what I can tell this field is not used anymore?

Licenser (approved):
🚀

PR #1738, opened 2022-06-10
Make compression levels and other parameters in processors configurable

### LINKED ISSUE #1726: Make compression levels and other parameters in processors configurable

It would be useful to some production users to be able to configure and optimise compression levels. Currently
we have postprocessors such as `zstd` which hardcode a compression level ( `0` ) where more configurability
is desireable.

Make compression algorithms more configurable perhaps via the `compress` postprocessor and add
appropriate parameters for setting optimization level by algorithm. Currently only `algorithm` is tunable.

**Notes**

omkar-mohanty:
Signed-off-by: omkar-mohanty <franzohouser@gmail.com>

# Pull request

## Description

This pull request aims to make compression levels more configurable.

## Related

* Related Issues: fixes #1726

### DIFF
--- a/src/postprocessor/compress.rs
+++ b/src/postprocessor/compress.rs
@@ -21,6 +21,7 @@ use std::{
 };
 use tremor_script::Value;
 
+#[derive(Debug, PartialEq)]
 enum Algorithm {
     Gzip,
     Zlib,
@@ -50,18 +51,9 @@ impl Algorithm {
     pub fn into_postprocessor(self, config: Option<&Value>) -> Result<Box<dyn Postprocessor>> {
         match config.get_i64("level") {
             Some(compression_level) => match self {
-                Algorithm::Xz2 => match Xz2::with_config(compression_level) {
-                    Ok(codec) => Ok(codec),
-                    Err(msg) => Err(msg),
-                },
-                Algorithm::Zstd => match Zstd::with_config(compression_level) {
-                    Ok(codec) => Ok(codec),
-                    Err(msg) => Err(msg),
-                },
-                Algorithm::Lz4 => match Lz4::with_config(compression_level) {
-                    Ok(codec) => Ok(codec),
-                    Err(msg) => Err(msg),
-                },
+                Algorithm::Xz2 => Xz2::with_config(compression_level),
+                Algorithm::Zstd => Zstd::with_config(compression_level),
+                Algorithm::Lz4 => Lz4::with_config(compression_level),
                 _ => Err("compression level not supported for given algorithm".into()),
             },
             None => {
@@ -114,7 +106,7 @@ struct Xz2 {
     compression_level: u32,
 }
 impl Xz2 {
-    fn with_config(level: i64) -> Result<Box<Self>> {
+    fn with_config(level: i64) -> Result<Box<dyn Postprocessor>> {
         if !(0..=9).contains(&level) {
             return Err(format!(
                 "Xz2 supports compression level between 0 and 9 but {level} was given"
@@ -169,7 +161,7 @@ struct Lz4 {
     compression_level: u32,
 }
 impl Lz4 {
-    pub fn with_config(level: i64) -> Result<Box<Self>> {
+    pub fn with_config(level: i64) -> Result<Box<dyn Postprocessor>> {
         if level < 0 {
             return Err("Lz4 compression level cannot be less than 0".into());
         }
@@ -207,10 +199,10 @@ struct Zstd {
     compression_level: i32,
 }
 impl Zstd {
-    pub fn with_config(level: i64) -> Result<Box<Self>> {
+    pub fn with_config(level: i64) -> Result<Box<dyn Postprocessor>> {
         if !(-7..=22).contains(&level) {
             return Err(format!(
-                "Xz2 supports compression level between 0 and 9 but {level} was given"
+                "Zstd supports compression level between -7 and 22 but {level} was given"
             )
             .into());
         }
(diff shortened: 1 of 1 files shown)


mfelsche:
Thanks for you stab at this task! :+1:

You added the options to set/get the compression level on the `Postprocessor` trait. This is somewhat wrong, as not all postprocessors should have this flag or should support this. The `Postprocessor` (same as the `Preprocessor`) they only should have the process function that takes a byte-array and returns zeor or more byte-arrays. What the processor does inside should not be exposed over the trait.

The way we configure postprocessors is through this function: https://github.com/tremor-rs/tremor-runtime/blob/main/src/postprocessor/compress.rs#L119 There it gets an `Option<&Value>` which contains the config for the specific processor (if any). In general, you can specify post- and pre-processors either simply by name or via a full blown name-and-config style. See an example here: https://www.tremor.rs/docs/edge/reference/preprocessors/separate (We should make that more clear in the docs).

So what we actually want is to adapt the configuration of the `compress` postprocessor to take values for configuring the few algorithms it supports. We can Add support per algorithm one at a time, they all have different options, i guess.

So the tasks for his PR would be the following (roughly, just a suggested guideline):

* Find out what options there are for each algorithm and choose which ones we want/need to support
* Find out how this should be configured by the user (i.e. find a good format that is quite common for all algorithms)
* Pass on the config to the specific postprocessor and consume it there, setting the specific options.

An example format for configuring could be:

```js
postprocessors = [
  {
    "name": "compress",
    "config": {
      "algorithm": "zstd",
      "config": {
        "level": 21
      }
    }
  }
]
```

This has the ugly doubling of `config`, maybe you find something better.

omkar-mohanty:
> You added the options to set/get the compression level on the Postprocessor trait. This is somewhat wrong, as not all postprocessors should have this flag or should support this. The Postprocessor (same as the Preprocessor) they only should have the process function that takes a byte-array and returns zeor or more byte-arrays. What the processor does inside should not be exposed over the trait.

Thanks for letting me know @mfelsche.

src/postprocessor/compress.rs
```diff
@@ -18,6 +18,44 @@ use simd_json::ValueAccess;
...
+    pub fn postprocessor(config: Option<&Value>, algorithm:Algorithm) -> Box<dyn Postprocessor> {
+        match config.get_str("level") {
+            Some(compression_level) => {
+                let compression_level = compression_level.parse::<i64>().unwrap(); 
+
+                let codec:Box<dyn Postprocessor> = match algorithm {
+                    Algorithm::Xz2  => Box::new(Xz2::with_config(compression_level as u32)),
+                    Algorithm::Zstd => Box::new(Zstd::with_config(compression_level as i32)),
+                    Algorithm::Lz4 => Box::new(Lz4::with_config(compression_level as u32)),
+                    _ => panic!("compression level not supported"),
```
mfelsche, line 40: Lets not panic the whole process here, but make the method return a `Result<Box<dyn Postprocessor>>` and return an `Err` here.

src/postprocessor/compress.rs
```diff
@@ -92,39 +153,61 @@ impl Postprocessor for Lz4 {
...
     pub(crate) fn from_config(config: Option<&Value>) -> Result<Self> {
         let codec: Box<dyn Postprocessor> =
             match config.get_str("algorithm").ok_or("Missing algorithm")? {
-                "gzip" => Box::new(Gzip::default()),
-                "zlib" => Box::new(Zlib::default()),
-                "xz2" => Box::new(Xz2::default()),
-                "snappy" => Box::new(Snappy::default()),
-                "lz4" => Box::new(Lz4::default()),
-                "zstd" => Box::new(Zstd::default()),
+                "gzip" => Algorithm::postprocessor(config, Algorithm::Gzip),
```
mfelsche, line 205: idea: What about implementing https://doc.rust-lang.org/std/str/trait.FromStr.html for `Algorithm`, use it here and then maybe implement a method like `fn into_postprocessor(self, config: Option<&Value>) -> Result<Box<dyn Postprocessor>>` and then use it here as well.

Licenser: I like the idea of having a Algoritm enum it's neat, it's much cleaner than the string matching we had before! So Matthias has a point `FromStr` would be neat here or `Deserialize` and a CompressorConfig to combine algorithm and level, could be an option if levels exist for all compression algorithms?

omkar-mohanty: I suppose so, but what about algorithms like "snappy" which do not have levels?

Licenser: That's a good point then `Deserialize` won't work too well.

src/postprocessor/compress.rs
```diff
@@ -15,9 +15,69 @@
...
+                "zstd" => Ok(Algorithm::Zstd),
+                other => return Err(format!("Unknown compression algorithm: {other}").into()),
+            }
+    }
+    
+}
+
+impl Algorithm {
+    pub fn into_postprocessor(config: Option<&Value>, algorithm:Algorithm) -> Result<Box<dyn Postprocessor>> {
+        match config.get_str("level") {
```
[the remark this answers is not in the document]
omkar-mohanty: Oh I was not aware of a `get_u64` function, thanks!!

src/postprocessor/compress.rs
```diff
@@ -92,42 +174,63 @@ impl Postprocessor for Lz4 {
...
+}
 
-#[derive(Clone, Default, Debug)]
-struct Zstd {}
+#[derive(Clone, Debug)]
+struct Zstd {
+    compression_level: i32,
+}
+impl Zstd {
+    pub fn with_config(level: i32) -> Self {
```
Licenser, line 197: Adding this here but it applies to all compressors. The constructor should probably valid the compression level to be correct, we don't want a runtime failure because someone defines level 999999 and only 20 is valid.

src/postprocessor/compress.rs
```diff
@@ -49,20 +110,42 @@ impl Postprocessor for Zlib {
...
-#[derive(Default)]
-struct Xz2 {}
+struct Xz2 {
+    compression_level: u32,
+}
+impl Xz2 {
+    fn with_config(level: i64) -> Result<Box<Self>> {
+        if !(0..=9).contains(&level) {
+            return Err(format!(
+                "Xz2 supports compression level between 0 and 9 but {level} was given"
```
mfelsche, line 120: :heart_eyes:

src/postprocessor/compress.rs
```diff
@@ -92,42 +187,63 @@ impl Postprocessor for Lz4 {
...
+
+#[derive(Clone, Debug, Default)]
+struct Zstd {
+    compression_level: i32,
+}
+impl Zstd {
+    pub fn with_config(level: i64) -> Result<Box<Self>> {
+        if !(-7..=22).contains(&level) {
+            return Err(format!(
+                "Xz2 supports compression level between 0 and 9 but {level} was given"
```
mfelsche, line 213:
```suggestion
                "Zstd supports compression level between -7 and 22 but {level} was given"
```

Licenser:
There isn't really a place in the code comments to put this, and reviews can easily feel cold and abrasive, so I'll put it here: 👍 great work so far :D, we really appreciate it!

src/postprocessor/compress.rs
```diff
@@ -49,20 +110,42 @@ impl Postprocessor for Zlib {
...
+    fn with_config(level: i64) -> Result<Box<Self>> {
+        if !(0..=9).contains(&level) {
+            return Err(format!(
+                "Xz2 supports compression level between 0 and 9 but {level} was given"
+            )
+            .into());
+        }
+
+        Ok(Box::new(Self {
+            compression_level: level.try_into().expect("could not parse compression level"),
```
mfelsche, line 126: expect always panics, lets just return an `Err` instead:

```suggestion
            compression_level: level.try_into()?,
```

src/postprocessor/compress.rs
```diff
@@ -82,8 +165,20 @@ impl Postprocessor for Snappy {
...
+    compression_level: u32,
+}
+impl Lz4 {
+    pub fn with_config(level: i64) -> Result<Box<Self>> {
+        if level < 0 {
+            return Err("Lz4 compression level cannot be less than 0".into());
+        }
+
+        Ok(Box::new(Self {
+            compression_level: level.try_into().expect("could not parse compression level"),
```
mfelsche, line 178:
```suggestion
            compression_level: level.try_into()?,
```

src/postprocessor/compress.rs
```diff
@@ -92,42 +187,63 @@ impl Postprocessor for Lz4 {
...
+            return Err(format!(
+                "Xz2 supports compression level between 0 and 9 but {level} was given"
+            )
+            .into());
+        }
 
-#[derive(Clone, Default, Debug)]
-struct Zstd {}
+        Ok(Box::new(Self {
+            compression_level: level.try_into().expect("could not parse compression level"),
```
mfelsche, line 219:
```suggestion
            compression_level: level.try_into()?,
```

mfelsche:
This looks pretty good. We are so close

src/postprocessor/compress.rs
```diff
@@ -15,9 +15,70 @@
...
+}
+
+impl Algorithm {
+    pub fn into_postprocessor(self, config: Option<&Value>) -> Result<Box<dyn Postprocessor>> {
+        match config.get_i64("level") {
+            Some(compression_level) => match self {
+                Algorithm::Xz2 => match Xz2::with_config(compression_level) {
+                    Ok(codec) => Ok(codec),
+                    Err(msg) => Err(msg),
+                },
```
Licenser, line 56: I think you can remove the match here
```suggestion
                Algorithm::Xz2 => Xz2::with_config(compression_level),
```

Licenser:
Awesome this looks great :) I think the only item left is the little copy and pasta error Matthias flagged for the `Zstd` config.

src/postprocessor/compress.rs
```diff
@@ -92,42 +179,63 @@ impl Postprocessor for Lz4 {
...
-struct Zstd {}
+#[derive(Clone, Debug, Default)]
+struct Zstd {
+    compression_level: i32,
+}
+impl Zstd {
+    pub fn with_config(level: i64) -> Result<Box<dyn Postprocessor>> {
+        if !(-7..=22).contains(&level) {
+            return Err(format!(
+                "Xz2 supports compression level between -7 and 22 but {level} was given"
```
Licenser, line 205: This still says Xz2, it should be Zstd here :)
```suggestion
                "Zstd supports compression level between -7 and 22 but {level} was given"
```

Licenser (dismissed):
Great work! 👍

mfelsche (approved):

PR #1751, opened 2022-06-15
Ci package test

Licenser:
# Pull request

## Description

This adds a package build test to CI so we can ensure that our ancient centos7 still builds stuff :sob: it also downgrades in malloc as it requires a kernel that is not older than 8 years. (I wish we could upgrade our kernel to something released in the past decade ...)

## Checklist

* [x] The code is tested

## Performance

🤷

### DIFF
--- a/.github/workflows/validate-packages.yaml
+++ b/.github/workflows/validate-packages.yaml
@@ -1,0 +1,17 @@
+name: Validate package creation
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Build releases
+        run: make builds
+      - name: Creating release packages
+        run: make packages
+      - name: List files
+        run: ls packaging/out/*
--- a/tremor-cli/Cargo.toml
+++ b/tremor-cli/Cargo.toml
@@ -39,10 +39,7 @@ serde_yaml = "0.8"
 signal-hook = "0.3"
 signal-hook-async-std = "0.2"
 simd-json = { version = "0.5", features = ["known-key"] }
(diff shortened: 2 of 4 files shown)


tremor-cli/Cargo.toml
```diff
@@ -39,13 +39,7 @@ serde_yaml = "0.8"
...
 signal-hook-async-std = "0.2"
 simd-json = { version = "0.5", features = ["known-key"] }
-# we need to stick with 0.2.26 as it includes its own libc
-# which allows us to build on older systems like centos 7
-# issues to track until we can loosen those restrictions:
-#   - https://github.com/microsoft/snmalloc/issues/328
-#   - https://github.com/SchrodingerZhu/snmalloc-rs/issues/145
-snmalloc-rs = { version = "=0.3.1", optional = false }
-snmalloc-sys = { version = "=0.3.1", optional = false }
+snmalloc-rs = { version = "0.2" }
```
mfelsche, line 42: sneaky little hobbitses. Can you add a comment on why the downgrade is necessary?

.github/workflows/validate-packages.yaml
```diff
@@ -0,0 +1,17 @@
...
+  package:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Build releases
+        run: make builds
+      - name: Creating release packages
+        run: make packages
+      - name: List files
+        run: ls packaging/out/*
```
mfelsche, line 17: would it be possible to actually assert/expect some files here and fail the action if they are not present?

Licenser: If they are not present the build fails to start with

mfelsche (approved):

PR #1761, opened 2022-06-17
Google Cloud Storage - writer

Agares:
# Pull request

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/223)

## Checklist

* [x] The code is tested

### DIFF
--- a/src/connectors/impls/gcs/api_client.rs
+++ b/src/connectors/impls/gcs/api_client.rs
@@ -526,13 +526,7 @@ mod tests {
         let client = MockHttpClient {
             config: Default::default(),
             handle_request: Box::new(|mut req| {
-                let body = block_on(req.body_bytes()).unwrap();
-                assert_eq!(vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9], body);
-                assert_eq!(req.header("Content-Range").unwrap()[0], "bytes 0-9/*");
-
-                let mut response = Response::new(StatusCode::Ok);
-                response.insert_header("Range", "bytes=0-10");
-                Ok(response)
+                Ok(Response::new(StatusCode::InternalServerError))
             }),
             simulate_failure: Arc::new(AtomicBool::new(true)),
             simulate_transport_failure: Arc::new(AtomicBool::new(true)),
@@ -546,7 +540,7 @@ mod tests {
             sessions_per_file,
             client,
             backoff_strategy: ExponentialBackoffRetryStrategy {
-                max_retries: 1,
+                max_retries: 2,
                 base_sleep_time: Duration::from_nanos(1),
             },
         };
@@ -642,17 +636,12 @@ mod tests {
 
     #[async_std::test]
     async fn fails_when_retries_are_exhausted() {
-        let request_handled = Arc::new(AtomicBool::new(false));
-        let request_handled_clone = request_handled.clone();
-
         let response = retriable_request(
-            &ExponentialBackoffRetryStrategy::new(1, Duration::from_nanos(1)),
+            &ExponentialBackoffRetryStrategy::new(3, Duration::from_nanos(1)),
             &mut MockHttpClient {
                 config: Default::default(),
                 handle_request: Box::new(move |_req| {
-                    request_handled_clone.swap(true, Ordering::Acquire);
-
-                    Ok(Response::new(StatusCode::Ok))
+                    Ok(Response::new(StatusCode::InternalServerError))
                 }),
                 simulate_failure: Arc::new(AtomicBool::new(true)),
                 simulate_transport_failure: Arc::new(Default::default()),
--- a/src/connectors/impls/gcs/api_client.rs
+++ b/src/connectors/impls/gcs/api_client.rs
@@ -678,4 +678,21 @@ mod tests {
         assert!(request_handled.load(Ordering::Acquire));
         assert_eq!(response.status(), StatusCode::Ok);
     }
+
+    #[test]
+    pub fn mock_http_client_config() {
+        let mut client = MockHttpClient {
+            config: Default::default(),
+            handle_request: Box::new(|_| Ok(Response::new(StatusCode::Ok))),
+            simulate_failure: Arc::new(Default::default()),
+            simulate_transport_failure: Arc::new(Default::default()),
+        };
+
+        let mut config = http_client::Config::new();
+        config.timeout = Some(Duration::from_secs(1000));
+
+        client.set_config(config.clone()).unwrap();
+
+        assert_eq!(client.config().timeout, config.timeout);
+    }
 }
--- a/src/connectors/impls/gcs/api_client.rs
+++ b/src/connectors/impls/gcs/api_client.rs
@@ -315,14 +315,12 @@ mod tests {
             (self.handle_request)(req)
         }
 
-        #[cfg_attr(coverage, no_coverage)] // This function is required by the trait, but never called in tests
         fn set_config(&mut self, config: http_client::Config) -> http_types::Result<()> {
             self.config = config;
 
             Ok(())
         }
 
-        #[cfg_attr(coverage, no_coverage)] // This function is required by the trait, but never called in tests
         fn config(&self) -> &http_client::Config {
             &self.config
         }
--- a/src/connectors/impls/gcs/api_client.rs
+++ b/src/connectors/impls/gcs/api_client.rs
@@ -82,7 +82,7 @@ async fn retriable_request<
     make_request: TMakeRequest,
 ) -> Result<Response> {
     let max_retries = backoff_strategy.max_retries();
-    for i in 1..max_retries + 1 {
+    for i in 1..=max_retries {
         let request = make_request();
         let error_wait_time = backoff_strategy.wait_time(i);
 
--- a/src/connectors/impls/gcs/api_client.rs
+++ b/src/connectors/impls/gcs/api_client.rs
@@ -82,7 +82,7 @@ async fn retriable_request<
     make_request: TMakeRequest,
 ) -> Result<Response> {
     let max_retries = backoff_strategy.max_retries();
-    for i in 0..max_retries {
+    for i in 1..max_retries + 1 {
         let request = make_request();
         let error_wait_time = backoff_strategy.wait_time(i);
 
@@ -93,14 +93,14 @@ async fn retriable_request<
                 match result {
                     Ok(mut response) => {
                         if response.status().is_server_error() {
+                            let response_body = response
+                                .body_string()
+                                .await
+                                .unwrap_or_else(|_| "<unable to read body>".to_string());
+
                             warn!(
                                 "Request {}/{} failed - Server error: {}",
-                                i + 1,
-                                max_retries,
-                                response
-                                    .body_string()
-                                    .await
-                                    .unwrap_or_else(|_| "<unable to read body>".to_string())
+                                i, max_retries, response_body
                             );
                             sleep(error_wait_time).await;
                             continue;
@@ -109,7 +109,7 @@ async fn retriable_request<
                         return Ok(response);
                     }
                     Err(error) => {
-                        warn!("Request {}/{} failed: {}", i + 1, max_retries, error);
+                        warn!("Request {}/{} failed: {}", i, max_retries, error);
                         sleep(error_wait_time).await;
                         continue;
                     }
@@ -118,9 +118,7 @@ async fn retriable_request<
             Err(error) => {
                 warn!(
                     "Request {}/{} failed to be created: {}",
-                    i + 1,
-                    max_retries,
-                    error
+                    i, max_retries, error
                 );
                 sleep(error_wait_time).await;
                 continue;
@@ -532,6 +530,55 @@ mod tests {
     }
 
     #[async_std::test]
+    pub async fn upload_data_fails_on_failed_request() -> Result<()> {
+        let done_until = Arc::new(AtomicUsize::new(0));
+        let client = MockHttpClient {
+            config: Default::default(),
+            handle_request: Box::new(|mut req| {
+                let body = block_on(req.body_bytes()).unwrap();
+                assert_eq!(vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9], body);
+                assert_eq!(req.header("Content-Range").unwrap()[0], "bytes 0-9/*");
+
+                let mut response = Response::new(StatusCode::Ok);
+                response.insert_header("Range", "bytes=0-10");
+                Ok(response)
+            }),
+            simulate_failure: Arc::new(AtomicBool::new(true)),
+            simulate_transport_failure: Arc::new(AtomicBool::new(true)),
+        };
+        let mut sessions_per_file = HashMap::new();
+        sessions_per_file.insert(
+            FileId::new("bucket", "my_file"),
+            url::Url::parse("https://example.com/session").unwrap(),
+        );
+        let mut api_client = DefaultApiClient {
+            sessions_per_file,
+            client,
+            backoff_strategy: ExponentialBackoffRetryStrategy {
+                max_retries: 1,
+                base_sleep_time: Duration::from_nanos(1),
+            },
+        };
+        let result = api_client
+            .handle_http_command(
+                done_until,
+                &Url::parse("http://example.com/upload").unwrap(),
+                HttpTaskCommand::UploadData {
+                    file: FileId::new("bucket", "my_file"),
+                    data: BufferPart {
+                        data: vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
+                        start: 0,
+                    },
+                },
+            )
+            .await;
+
+        assert!(result.is_err());
+
+        Ok(())
+    }
+
+    #[async_std::test]
     pub async fn can_finish_upload() -> Result<()> {
         let done_until = Arc::new(AtomicUsize::new(0));
         let client = MockHttpClient {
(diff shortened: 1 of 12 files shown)


src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#[async_trait::async_trait]
+impl Connector for GCSWriterConnector {
+    async fn create_sink(
+        &mut self,
+        sink_context: SinkContext,
+        builder: SinkManagerBuilder,
+    ) -> Result<Option<SinkAddr>> {
+        let url = Url::<HttpsDefaults>::parse(&self.config.endpoint)?;
```
Licenser, line 67: this ideally would be checked in `build_cfg` already so we get an error as soon as possible

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,682 @@
...
+        let sink = GCSWriterSink {
+            client: None,
+            url: self.url.clone(),
+            config: self.config.clone(),
+            buffers: ChunkedBuffer::new(self.config.buffer_size),
+            current_name: None,
+            current_session_url: None,
+        };
+
+        let addr = builder.spawn(sink, sink_context)?;
```
mfelsche, line 110:
```suggestion
        builder.spawn(sink, sink_context).map(Some)
```

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,682 @@
...
+        Ok(Some(addr))
+    }
+
+    fn codec_requirements(&self) -> CodecReq {
+        CodecReq::Required
+    }
+}
+
+struct ChunkedBuffer {
```
mfelsche, line 119: This reminds me a lot of https://docs.rs/bytes/latest/bytes/struct.BytesMut.html

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,682 @@
...
+        self.data = Vec::from(self.data.get(bytes_to_remove..).ok_or(
+            ErrorKind::GoogleCloudStorageError("Not enough data in the buffer"),
+        )?);
+        self.buffer_start += bytes_to_remove;
+
+        Ok(())
+    }
+
+    pub fn read_current_block(&self) -> Option<&[u8]> {
+        self.data.get(..self.block_size)
```
mfelsche, line 149: they way the data is used below (the `Body` impl clones the slice again), it might be more efficient to use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_off and return self and assign the returned vec to `self.data` (don't forget to set the capacity to `self.block_size * 2` afterwards).

To avoid `split_off` panicking, you could use it like this `self.data.split_off(self.block_size.min(self.data.len())`

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,682 @@
...
+                            break;
+                        }
+
+                        sleep(Duration::from_millis(25u64 * 2u64.pow(i))).await;
+                    }
+                }
+
+                if let Some(response) = response {
+                    if response.status().is_server_error() {
+                        return Err("Received server errors from Google Cloud Storage".into());
```
mfelsche, line 263: does the response body contain any useful error payload?

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,682 @@
...
+            self.start_upload_if_needed(meta, name).await?;
+
+            let serialized_data = serializer.serialize(value, event.ingest_ns)?;
+            for item in serialized_data {
+                self.buffers.write(&item);
+            }
+
+            if let Some(data) = self.buffers.read_current_block() {
+                let mut response = None;
+                for i in 0..3 {
```
mfelsche, line 215: Do we really want to keep this retry logic? We don't do similar things e.g. with the `http_client` connector. If people want retries, they can put a WAL in between or make their events `transactional` with other means. What do you think?

mfelsche (commented):
Good stuff! I am a bit hesitant if we should keep the request retry logic though and/or make it configurable?

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,682 @@
...
+        }
+    }
+
+    pub fn mark_done_until(&mut self, position: usize) -> Result<()> {
+        if position < self.buffer_start {
+            return Err("Buffer was marked as done at index which is not in memory anymore".into());
+        }
+
+        let bytes_to_remove = position - self.buffer_start;
+        self.data = Vec::from(self.data.get(bytes_to_remove..).ok_or(
```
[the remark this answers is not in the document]
Agares: Yes, but this should never happen, so that would potentially hide bugs in other places.

src/connectors/impls/gcs/writer.rs
Agares: We could remove it, though in my experience, the errors here happen quite often.

src/connectors/impls/gcs/writer.rs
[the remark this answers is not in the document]
mfelsche: Ah got it, because we are tracking the current cursor position in the overall bytes here too. :heavy_check_mark:

src/connectors/impls/gcs/writer.rs
mfelsche: The one big problem i have with this is that if we sleep here waiting for a retry, we are blocking the whole `sink` task from making progress. The http_client spawns a task for every http request sent. Could we somehow do the same here, keep track of the tasks/joinhandles/futures and chain those futures, so the uploads actually happen in order?

Agares: hmm... that could work, we could also have a background task that deals with the communication with GCS, but this brings another issue - how do we deal with ACKs then? AFAIK the ACK has to happen as `on_event` returns, unless I'm missing something

mfelsche: Here is an example from the http_client: 
* https://github.com/tremor-rs/tremor-runtime/blob/main/src/connectors/impls/http/client.rs#L164
* https://github.com/tremor-rs/tremor-runtime/blob/main/src/connectors/impls/http/client.rs#L373-L378

src/connectors/impls/gcs/writer.rs
Agares: I'm afraid we can't do this, because there's no guarantee that google will persist all of the data that we're sending to it. I.e. we might send 256kiB, but google can persists only 128kiB of that (or 0), so we need to keep it in memory until we receive a response and know how much we can throw away. 

```
If the request succeeds, the server responds with 308 Resume Incomplete. The response contains a [Range header](https://cloud.google.com/storage/docs/json_api/v1/parameters#range). Use the upper value in this header to determine where to start the next chunk. You should not assume that the server received all bytes sent in the request.
```
(https://cloud.google.com/storage/docs/performing-resumable-uploads#chunked-upload)

mfelsche: Got it, that's why we need to have the actual cutting and removal of old stuff separate. Good point! :)

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,831 @@
...
+#[cfg(test)]
+use tests::MockHttpClient;
+use tremor_pipeline::{ConfigImpl, Event};
+use tremor_value::Value;
+use value_trait::ValueAccess;
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct Config {
+    #[serde(default = "default_endpoint")]
+    endpoint: String,
```
Licenser, line 47: we generalized to `url`  for everything that takes  url I think we should do it here too and not use `String` but `Url::<HttpsDefaults>` as a type.

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,831 @@
...
+    async fn create_sink(
+        &mut self,
+        sink_context: SinkContext,
+        builder: SinkManagerBuilder,
+    ) -> Result<Option<SinkAddr>> {
+        let default_bucket = self
+            .config
+            .bucket
+            .as_ref()
+            .map(|bucket_name| Value::from(bucket_name.as_str()).into_static());
```
Licenser, line 114:
```suggestion
            .cloned()
            .map(Value::from);
```

src/connectors/impls/gcs/writer.rs
Agares: ohhh, right, I didn't think to even try deserializing this, but it's a good idea, will make the code a bit simpler :)

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,831 @@
...
+
+        builder.spawn(sink, sink_context).map(Some)
+    }
+
+    fn codec_requirements(&self) -> CodecReq {
+        CodecReq::Required
+    }
+}
+
+struct ChunkedBuffer {
```
Licenser, line 133: This could need a bit of documentation on what it does and what the intent is, I've a hard time figuring that out just from the code

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,831 @@
...
+    }
+
+    pub fn final_block(&self) -> &[u8] {
+        self.data.as_slice()
+    }
+}
+
+enum HttpTaskCommand {
+    FinishUpload {
+        name: String,
```
Licenser, line 185: We use `bucket` in `StartUpload` so could this collide with having files with the same name but different buckets?

Licenser: looking at the code below, yap that would clash

src/connectors/impls/gcs/writer.rs
Licenser: Since we have the retry logic in 3 places here, would it make sense to abstract it out, make it configurable and allow it be used elsewhere too?

hardcoded 3 tries at 25ms^2*i seems like a recepie for pain :joy:

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,831 @@
...
+                    "The file name in metadata is not a string",
+                ))?;
+
+            self.finish_upload_if_needed(name).await?;
+
+            self.start_upload_if_needed(meta, name).await?;
+
+            let serialized_data = serializer.serialize(value, event.ingest_ns)?;
+            for item in serialized_data {
+                self.buffers.write(&item);
```
Licenser, line 483: the tasks seem to be bucket/name based, allowing for multiple names to be uploaded, but there seems to be only one buffer, I feel like I'm missing something here. are the bucket/name tasks simply "future-proofing" and not relevant in `on_event`?

Agares: The current way of working is that we're uploading all the events to a single buffer, until the name (and/or bucket) changes, and the `http_task` processes the events sequentially, so there's no need for multiple buffers.
I hope in the future we can support uploading multiple files in parallel, but this will require some careful analysis.

src/connectors/impls/gcs/writer.rs
Agares: Yep, all good points, I'll revamp the error handling and retries.

src/connectors/impls/gcs/writer.rs
Licenser: makes perfect sense!

src/connectors/impls/gcs/api_client.rs
```diff
@@ -0,0 +1,476 @@
...
+        if let Some(mut response) = response {
+            if response.status().is_server_error() {
+                error!(
+                    "Error from Google Cloud Storage: {}",
+                    response.body_string().await?
+                );
+                return Err("Received server errors from Google Cloud Storage".into());
+            }
+
+            if let Some(raw_range) = response.header("range") {
```
Licenser, line 143: This is nicer written as:

```rust
let new_range = response.header("range").ok_or("No range header")?;
```

it removes a level of nesting and puts the error right next to where it occurs instead of 20 lines below :)

src/connectors/impls/gcs/api_client.rs
```diff
@@ -0,0 +1,476 @@
...
+                    )
+                    .ok_or(ErrorKind::GoogleCloudStorageError(
+                        "Unable to get the end of the Range",
+                    ))?;
+
+                // NOTE: The data can only be thrown away to the point of the end of the Range header,
+                // since google can persist less than we send in our request.
+                //
+                // see: https://cloud.google.com/storage/docs/performing-resumable-uploads#chunked-upload
+                done_until.store(range_end.parse()?, Ordering::Release);
```
Licenser, line 169: Is done_until really shared over multiple threads (do we require an Atomic here) if not, is the ordering the right one? A comment on the sharing logic explaining it would be helpful to avoid confusion in a few months.

src/connectors/impls/gcs/api_client.rs
```diff
@@ -0,0 +1,476 @@
...
+        Ok(())
+    }
+
+    async fn start_upload(&mut self, url: &Url<HttpsDefaults>, file: FileId) -> Result<()> {
+        for i in 0..3 {
+            let url = url::Url::parse(&format!(
+                "{}/b/{}/o?name={}&uploadType=resumable",
+                url, file.bucket, file.name
+            ))?;
+            #[cfg(not(test))]
```
Licenser, line 186: can we isolate the `test` and `non test` code into a separate function that's swapped out? mixing `cfg(test)` into the code makes it extremely hard to follow.

src/connectors/impls/gcs/api_client.rs
```diff
@@ -0,0 +1,476 @@
...
+            #[cfg(test)]
+            let request = Request::new(Method::Post, url);
+            #[cfg(not(test))]
+            {
+                request.insert_header("Authorization", self.token.header_value()?.to_string());
+            }
+
+            let response = self.client.send(request).await;
+
+            if let Ok(mut response) = response {
```
Licenser, line 197: It would be nice to log the error if one occurred, right now, it seems to be quietly swallowed and retried.

src/connectors/impls/gcs/api_client.rs
```diff
@@ -0,0 +1,476 @@
...
+                    data.start + data.len() - 1,
+                    data.start + data.len()
+                ),
+            );
+
+            request.set_body(data.data.clone());
+
+            let response = self.client.send(request).await;
+
+            if let Ok(mut response) = response {
```
Licenser, line 257: Same as above, it'd be nice not to swallow this error completely if something goes wrong.

src/connectors/impls/gcs/api_client.rs
```diff
@@ -0,0 +1,476 @@
...
+                error!(
+                    "Error from Google Cloud Storage: {}",
+                    response.body_string().await?
+                );
+            }
+
+            sleep(Duration::from_millis(25u64 * 2u64.pow(i))).await;
+        }
+
+        Ok(())
```
Licenser, line 270: Same for the above functions.

If I read the code correctly, if we fail three times to upload then we silently succeed this seems wrong, shouldn't the default response be an Error

src/connectors/impls/gcs/chunked_buffer.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    pub fn write(&mut self, data: &[u8]) {
+        self.data.extend_from_slice(data);
+    }
+
+    pub fn start(&self) -> usize {
+        self.buffer_start
+    }
+
+    // TODO: Consume self instead?
```
Licenser, line 79: That would be nice, saves us a clone :D

src/connectors/impls/gcs/writer.rs
```diff
@@ -0,0 +1,949 @@
...
+    endpoint: Url<HttpsDefaults>,
+    #[serde(default = "default_connect_timeout")]
+    connect_timeout: u64,
+    #[serde(default = "default_buffer_size")]
+    buffer_size: usize,
+    bucket: Option<String>,
+}
+
+#[allow(clippy::unwrap_used)]
+fn default_endpoint() -> Url<HttpsDefaults> {
```
Licenser, line 53: Would be nice to add a test-case for this to guarantee if ever an error sneaks in here we find it during testing

Licenser (changes requested):
A lot of the errors are plain strings and not wrapped into `ErrorKind::GoogleCloudStorageError` it would be good to unify this so people know where the error comes from.


I am a bit unsure about handling errors during writes, I think the way the connector acts right now we introduce the possibility for file corruption on write, let me give an example of the failure condition I have in mind:

- we have 10 events 0..9 that contain the chunks of the file
- we receive then I'm quick succession so that when event 10 arrives at the connector not all events have been written yet, say only 1 and 2 have been written
- now event 3 fails catastrophically (as in retries don't work)
- we send a CB fail event for 3 causing the upstrea system to resend everything from 3 onwards
- we write event 4 to 10 successfully
- the re-sends for 3-10 arrive, we now add them at the end of the file

This would lead to a corrupted file that tremor tells the user was uploaded successfully. Or I am missing the code that handles this condition :D it's a big PR could totally be.

I'm not sure how to best handle this the only way I see is not acknowledging any writes until the entire file is written, but that itself has problems for "endless streaming" use cases like appending logs. An alternative would be to make the connector as non-transactional and ignore all write failures turning it into a udp-like fire and forget one?

src/connectors/impls/gcs/api_client.rs
Agares: it's shared between multiple tasks (i.e. the background task writes, and the sink task reads), AFAIK we can't make any guarantees about which core will it run on.
I'll add some comments though.

Licenser (approved):
🚀 weeeeh!

PR #1787, opened 2022-06-27
Remove old tokui 0.3 dep

Licenser:
# Pull request

## Description

Removes old tokio 0.3 dependency

## Checklist

* [x] The code is tested

## Performance

Seven Buritos

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ async-compression = { version = "0.3", features = [
 async-std = { version = "1.12.0", features = [
   "unstable",
   "attributes",
-  "tokio03",
   "tokio1",
 ] }
 async-stream = "0.3"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -226,9 +225,7 @@ uuid = { version = "1.1", features = ["v4"] }
 qwal = { git = "https://github.com/tremor-rs/qwal" }
 
 [dev-dependencies]
-serial_test = { version = "=0.8", features = [
-  "logging",
-] }
+serial_test = { version = "=0.8", features = ["logging"] }
```
Agares, line 228: is this supposed to be in this PR? seem unrelatedd

Agares (approved):

Cargo.toml
Licenser: ja it sucks when that happens -.- we've not found a way yet to unify this the same way it works with `cargo fmt` :(

PR #1808, opened 2022-07-01
Make otel logs less strict and handle optional fields as optional

mfelsche:
# Pull request

## Description

This might be required for metrics and traces as well.

## Checklist

* [ ] The code is tested


src/connectors/impls/otel/client.rs
```diff
@@ -147,17 +147,26 @@ impl Sink for OtelSink {
             for value in event.value_iter() {
                 let err = if self.config.metrics && value.contains_key("metrics") {
                     let request = ExportMetricsServiceRequest {
-                        resource_metrics: metrics::resource_metrics_to_pb(Some(value))?,
+                        resource_metrics: ctx.bail_err(
```
Licenser, line 150: would a heper function that includes `value` in the error be helpful here? Something like

```rust
fn try_convert<T>(value: Value, converter: Fn<(Value) ->T>, err: &str) -> Result<T> {...}
```

Licenser: I think we already do that, the logging framework lets people set levels based on module, so if we classify invalid content as warnings or even info they can just opt out for them. Also we could send them to the error port, together with the work that Rebeca is doing that could solve the whole problem once an for all :)

src/connectors/impls/otel/logs.rs
```diff
@@ -158,7 +168,7 @@ pub(crate) fn resource_logs_to_json(request: ExportLogsServiceRequest) -> Result
 
 pub(crate) fn resource_logs_to_pb(json: &Value<'_>) -> Result<Vec<ResourceLogs>> {
     json.get_array("logs")
-        .ok_or("Invalid json mapping for otel logs message - cannot convert to pb")?
+        .ok_or("Missing \"logs\" array")?
```
Licenser, line 171:
```suggestion
        .ok_or("Missing `logs` array")?
```

tremor-cli/src/target_process.rs
```diff
@@ -131,6 +131,14 @@ impl TargetProcess {
             };
             buf.canonicalize()?
         };
+        // env var replacement
+        let mut args: Box<dyn Iterator<Item = String>> = Box::new(args.iter().cloned());
+        for (var, value) in env {
+            args = Box::new(args.map(|s| s.replace(&format!("${}", var.clone()), value)));
```
Licenser, line 137: I don't think we need the clone here, as it's in a  format it can be borrowed.

Licenser (approved):

tremor-cli/src/target_process.rs
mfelsche: I wouldn't have added it, if the borrow checker didn't tell me it was necessary. The thing is that this is an iterator, which only lazily applies the closure to the base-iterator, so `var` needs to be cloned to outlive the containing loop iteration.

darach (changes requested):

tremor-cli/tests/integration/otel-client/tmp
```diff
@@ -0,0 +1,34 @@
+{
```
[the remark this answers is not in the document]
Licenser: fixed this

PR #1811, opened 2022-07-05
Add initial GCP logging write connector

darach:

# Pull request

Adds GCP cloud logging support for writing log entries using json payloads.
Initial experimental revision for soliciting feedback.

## Description

Adds basic support for GCP cloud logging

## Related

None.

## Checklist

* [x] The code is tested

## Performance

This is net new functionality with no performance testing conducted to date.

### DIFF
--- a/src/connectors/impls/gcl.rs
+++ b/src/connectors/impls/gcl.rs
@@ -1,0 +1,77 @@
+// Copyright 2022, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use rand::Rng;
+use tremor_script::{tremor_fn, Registry, Value};
+
+pub(crate) mod writer;
+
+// The span ID within the trace associated with the log entry.
+//  For Trace spans, this is the same format that the Trace API v2
+//  uses: a 16-character hexadecimal encoding of an 8-byte array,
+// such as 000000000000004a
+fn random_span_id_value(ingest_ns_seed: u64) -> Value<'static> {
+    let mut rng = tremor_common::rand::make_prng(ingest_ns_seed);
+    let span_id: String = (0..8)
+        .map(|_| rng.gen::<u8>())
+        .map(|b| format!("{:02x}", b))
+        .collect();
+    Value::from(span_id)
+}
+
+fn random_trace_id_value(ingest_ns_seed: u64) -> Value<'static> {
+    let mut rng = tremor_common::rand::make_prng(ingest_ns_seed);
+    let span_id: String = (0..16)
+        .map(|_| rng.gen::<u8>())
+        .map(|b| format!("{:02x}", b))
+        .collect();
+    Value::from(span_id)
+}
+
+/// Extend function registry with `GCP Cloud Logging` support
+pub fn load(registry: &mut Registry) {
+    registry
+        .insert(tremor_fn! (google_logging|gen_span_id_string(ctx) {
+            Ok(random_span_id_value(ctx.ingest_ns()))
+        }))
+        .insert(tremor_fn! (google_logging|gen_trace_id_string(ctx) {
+            Ok(random_trace_id_value(ctx.ingest_ns()))
+        }));
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use test_case::test_case;
+    use tremor_script::{registry, EventContext};
+
+    #[test_case("gen_span_id_string" ; "google_logging::gen_span_id_string")]
+    #[test_case("gen_trace_id_string" ; "google_logging::gen_trace_id_string")]
+    fn test_tremor_fns(fun_name: &str) {
+        let mut ok_count = 0;
+        let mut reg = registry::registry();
+        load(&mut reg);
+
+        let f = reg.find("google_logging", fun_name);
+
+        if let Ok(f) = f {
+            let context = EventContext::default();
+            let r = f.invoke(&context, &[]);
+            assert!(r.is_ok());
+            ok_count += 1;
+        }
+
+        assert_eq!(1, ok_count);
+    }
+}
--- a/src/connectors/impls/gcl/writer.rs
+++ b/src/connectors/impls/gcl/writer.rs
@@ -97,7 +97,7 @@ pub(crate) struct Config {
     pub default_severity: i32,
 
     /// This setting sets a default set of labels that can be overriden on a per event
-    /// // basis through metadata
+    /// basis through metadata
     #[serde(default = "Default::default")]
     pub labels: HashMap<String, String>,
 }
@@ -180,6 +180,7 @@ fn value_to_monitored_resource(
         Some(from) => {
             let vt = from.value_type();
             match from {
+                // Consider refactoring for unwrap_or_default when #1819 is resolved
                 OwnedValue::Object(from) => {
                     let kind = from.get("type");
                     let kind = kind.as_str();
--- a/src/connectors/impls/gcl/writer.rs
+++ b/src/connectors/impls/gcl/writer.rs
@@ -1,0 +1,318 @@
+// Copyright 2022, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub(crate) mod meta;
+mod sink;
+
+use crate::connectors::impls::gcl::writer::sink::GclSink;
+use crate::connectors::prelude::*;
+use crate::connectors::{Connector, ConnectorBuilder, ConnectorConfig, ConnectorType};
+use crate::errors::Error;
+use googapis::google::api::MonitoredResource;
+use googapis::google::logging::r#type::LogSeverity;
+use serde::Deserialize;
+use simd_json::OwnedValue;
+use std::collections::HashMap;
+use tremor_pipeline::ConfigImpl;
+
+#[derive(Deserialize, Clone)]
+pub(crate) struct Config {
+    /// The default `log_name` for this configuration or `default` if not provided.
+    /// The `log_name` field can be overridden in metadata.
+    ///
+    /// From the protocol buffers specification:
+    ///
+    /// Optional. A default log resource name that is assigned to all log entries in entries that do not specify a value for log_name:
+    ///
+    /// "projects/\[PROJECT_ID]/logs/[LOG_ID\]"
+    /// "organizations/\[ORGANIZATION_ID]/logs/[LOG_ID\]"
+    /// "billingAccounts/\[BILLING_ACCOUNT_ID]/logs/[LOG_ID\]"
+    /// "folders/\[FOLDER_ID]/logs/[LOG_ID\]"
+    /// \[LOG_ID\] must be URL-encoded. For example:
+    ///
+    /// "projects/my-project-id/logs/syslog"
+    /// "organizations/1234567890/logs/cloudresourcemanager.googleapis.com%2Factivity"
+    /// The permission logging.logEntries.create is needed on each project, organization, billing account, or folder that is receiving new log entries, whether the resource is specified in logName or in an individual log entry.
+    pub log_name: Option<String>,
+
+    /// The default monitored resource for the log entries being written.
+    ///
+    /// From the protocol buffers specification:
+    ///
+    /// Optional. A default monitored resource object that is assigned to all log entries in entries that do not specify a value for resource. Example:
+    ///
+    /// ```json
+    /// { "type": "gce_instance",
+    /// "labels": {
+    ///   "zone": "us-central1-a", "instance_id": "00000000000000000000" }}
+    /// ```    
+    #[serde(default = "Default::default")]
+    pub resource: Option<simd_json::OwnedValue>,
+
+    /// This setting sets the behaviour of the connector with respect to whether valid entries should
+    /// be written even if some entries in a batch set to Google Cloud Logging are invalid. Defaults to
+    /// false
+    ///
+    /// From the protocol buffers specification:
+    ///
+    /// Optional. Whether valid entries should be written even if some other entries fail due to INVALID_ARGUMENT or PERMISSION_DENIED errors. If any entry is not written, then the response status is the error associated with one of the failed entries and the response includes error details keyed by the entries' zero-based index in the entries.write method.
+    #[serde(default = "default_partial_success")]
+    pub partial_success: bool,
+
+    /// This setting enables a sanity check for validating that log entries are well formed and
+    /// and valid by exercising the connector to write entries where resulting entries are not
+    /// persisted. Useful primarily during initial exploration and configuration or after
+    /// large configuration changes as a sanity check.
+    ///
+    /// From the protocol buffers specification:
+    ///
+    /// Optional. If true, the request should expect normal response, but the entries won't be persisted nor exported. Useful for checking whether the logging API endpoints are working properly before sending valuable data.
+    #[serde(default = "default_dry_run")]
+    pub dry_run: bool,
+
+    /// Sets a connection timeout for connections to the GCP logging endpoint
+    /// that applies during connection
+    #[serde(default = "default_connect_timeout")]
+    pub connect_timeout: u64,
+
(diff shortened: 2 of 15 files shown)


src/connectors/impls/gcl.rs
```diff
@@ -0,0 +1,77 @@
...
+pub(crate) mod writer;
+
+// The span ID within the trace associated with the log entry.
+//  For Trace spans, this is the same format that the Trace API v2
+//  uses: a 16-character hexadecimal encoding of an 8-byte array,
+// such as 000000000000004a
+fn random_span_id_value(ingest_ns_seed: u64) -> Value<'static> {
+    let mut rng = tremor_common::rand::make_prng(ingest_ns_seed);
+    let span_id: String = (0..8)
+        .map(|_| rng.gen_range(0_u8..=255_u8))
```
Licenser, line 27:
```suggestion
        .map(|_| rng.gen<u8>())
```

src/connectors/impls/gcl.rs
```diff
@@ -0,0 +1,77 @@
...
+        .map(|b| format!("{:02x}", b))
+        .collect();
+    Value::from(span_id)
+}
+
+fn random_trace_id_value(ingest_ns_seed: u64) -> Value<'static> {
```
Licenser, line 33: would it make sense to have a `random_id` function that both `random_trace_id_value` and `random_span_id_value` call

Licenser: Narrator: it wouldn't since they have a different size ...

src/connectors/impls/gcl/writer.rs
```diff
@@ -0,0 +1,239 @@
...
+use googapis::google::api::MonitoredResource;
+use googapis::google::logging::r#type::LogSeverity;
+use serde::Deserialize;
+use simd_json::OwnedValue;
+use std::collections::HashMap;
+use tremor_pipeline::ConfigImpl;
+
+#[derive(Deserialize, Clone)]
+pub(crate) struct Config {
+    // Optional. A default log resource name that is assigned to all log entries in entries that do not specify a value for log_name:
```
Licenser, line 31: `///` would turn all those into comments the compiler understands :)

src/connectors/impls/gcl/writer.rs
```diff
@@ -0,0 +1,239 @@
...
+    // "projects/my-project-id/logs/syslog"
+    // "organizations/1234567890/logs/cloudresourcemanager.googleapis.com%2Factivity"
+    // The permission logging.logEntries.create is needed on each project, organization, billing account, or folder that is receiving new log entries, whether the resource is specified in logName or in an individual log entry.
+    pub log_name: Option<String>,
+
+    // Optional. A default monitored resource object that is assigned to all log entries in entries that do not specify a value for resource. Example:
+    //
+    // { "type": "gce_instance",
+    // "labels": {
+    //   "zone": "us-central1-a", "instance_id": "00000000000000000000" }}
```
Licenser, line 48:
```suggestion
    /// ```json
    /// { "type": "gce_instance",
    /// "labels": {
    ///   "zone": "us-central1-a", "instance_id": "00000000000000000000" }}
    /// ```    
```

src/connectors/impls/gcl/writer.rs
```diff
@@ -0,0 +1,239 @@
...
+    #[serde(default = "default_resource")]
```
Licenser, line 49: can we save `default_resource`? Option defaults to `None`

Licenser:
```suggestion
    #[serde(default = "Default::default")]
```

Licenser: the serde default might not even be needed at all, I think anything that's a Option can be omitted

src/connectors/impls/gcl/writer.rs
```diff
@@ -0,0 +1,239 @@
...
+    #[serde(default = "default_dry_run")]
+    pub dry_run: bool,
+
+    // pub table_id: String,
+    #[serde(default = "default_connect_timeout")]
+    pub connect_timeout: u64,
+
+    // Default Log severity
+    #[serde(default = "default_log_severity")]
+    pub default_severity: Option<i32>,
```
Licenser, line 66: Is there a difference between `None` and `LogSeverity::Info` if not we could just make this a i32 and save the option dance later on?

Licenser: Found where it's used, it seems there is a difference, should the default be `None` here instead of Info? As this is written now even if `default_severity` isn't defined in the config it will overwrite it with `Info`

src/connectors/impls/gcl/writer.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    #[serde(default = "default_labels")]
+    pub labels: Option<HashMap<String, String>>,
```
Licenser, line 69: It seems like using 

```suggestion
    #[serde(default = "Default::default")]
    pub labels: HashMap<String, String>,
```

instead will simplify some of the code below the `default_labels` function is no longer needed that way

src/connectors/impls/gcl/writer.rs
```diff
@@ -0,0 +1,239 @@
...
+) -> Result<Option<MonitoredResource>> {
+    match from {
+        None => Ok(None),
+        Some(from) => {
+            let vt = from.value_type();
+            match from {
+                OwnedValue::Object(from) => {
+                    let kind = from.get("type");
+                    let kind = kind.as_str();
+                    let maybe_labels = from.get("labels");
```
Licenser, line 110: `unwrap_or_default` would save the match and nesting below

src/connectors/impls/gcl/writer/meta.rs
```diff
@@ -0,0 +1,604 @@
...
+    v2::{LogEntryOperation, LogEntrySourceLocation},
+};
+use halfbrown::HashMap;
+use std::collections::HashMap as StdHashMap;
+use tremor_value::Value;
+use value_trait::ValueAccess;
+
+use super::Config;
+
+pub(crate) fn default_log_name(config: &Config) -> String {
```
Licenser, line 28: this should probably be a function of `Config` not a stand alone function

src/connectors/impls/gcl/writer/meta.rs
```diff
@@ -0,0 +1,604 @@
...
+    }
+
+    // NOTE user supplied defaults supported at this time
+
+    "".to_string() // Allow GCL to auto assign a unique id for this event
+}
+
+pub(crate) fn http_request(meta: Option<&HashMap<Cow<str>, Value>>) -> Option<HttpRequest> {
+    // Override for a specific per event trace
+    if let Some(has_meta) = meta {
```
Licenser, line 91: This can be un nested by using:

```rust
fn ... {
  let meta = meta?
  let http_request = has_meta.get("http_request")?;

  Soime(...)
}
```

that would make the function a lot more readable

src/connectors/impls/gcl/writer/meta.rs
```diff
@@ -0,0 +1,604 @@
...
+    labels
+}
+
+pub(crate) fn operation(meta: Option<&HashMap<Cow<str>, Value>>) -> Option<LogEntryOperation> {
+    // Override for a specific per event trace
+    if let Some(has_meta) = meta {
+        if let Some(operation @ Value::Object(_)) = has_meta.get("operation") {
+            return Some(LogEntryOperation {
+                id: operation.get("id").as_str().unwrap_or("").to_string(),
+                producer: operation.get("producer").as_str().unwrap_or("").to_string(),
```
Licenser, line 184:
```suggestion
                id: operation.get_str("id").unwrap_or("").to_string(),
                producer: operation.get_str("producer").unwrap_or("").to_string(),
```

src/connectors/impls/gcl/writer/meta.rs
```diff
@@ -0,0 +1,604 @@
...
+    if let Some(has_meta) = meta {
+        if let Some(trace) = has_meta.get("trace") {
+            return trace.to_string();
+        }
+    }
+
+    // NOTE user supplied defaults supported at this time
+
+    "".to_string()
+}
```
Licenser, line 206: Passing in `Option<&Value>` could simplify this and the other functions a lot:
```suggestion
pub(crate) fn get_key_or_dflt(meta: Option<&Value>, key: &str) -> String {

    // Override for a specific per event trace
    // NOTE user supplied defaults supported at this time
    meta.get_str(key).unwrap_or_default().to_string()
}
```

src/connectors/impls/gcl/writer/meta.rs
```diff
@@ -0,0 +1,604 @@
...
+    if let Some(has_meta) = meta {
+        if let Some(trace_sampled) = has_meta.get("trace_sampled") {
+            #[allow(clippy::redundant_else)] // Not redundant
+            if let Some(trace_sampled) = trace_sampled.as_bool() {
+                return Ok(trace_sampled);
+            } else {
+                return Err(
+                    "Unable to parse `trace_sampled` from event metadata - expected bool".into(),
+                );
+            }
```
Licenser, line 232: The way this is written, the else in fact is redundant it can be simplified w/o the need for an allow:
```suggestion
            return  if let Some(trace_sampled) = trace_sampled.as_bool() {
                Ok(trace_sampled);
            } else {
                Err(
                    "Unable to parse `trace_sampled` from event metadata - expected bool".into(),
                );
            }
```

src/connectors/impls/gcl/writer/meta.rs
```diff
@@ -0,0 +1,604 @@
...
+pub(crate) fn source_location(
+    meta: Option<&HashMap<Cow<str>, Value>>,
+) -> Option<LogEntrySourceLocation> {
+    // Override for a specific per event trace
+    if let Some(has_meta) = meta {
+        if let Some(loc @ Value::Object(_)) = has_meta.get("source_location") {
+            return Some(LogEntrySourceLocation {
+                file: loc.get("file").as_str().unwrap_or("").to_string(),
+                line: loc.get("line").as_i64().unwrap_or(0),
+                function: loc.get("function").as_str().unwrap_or("").to_string(),
```
Licenser, line 250:
```suggestion
                file: loc.get_str("file").unwrap_or("").to_string(),
                line: loc.get_i64("line").unwrap_or(0),
                function: loc.get_str("function").unwrap_or("").to_string(),
```

src/connectors/impls/gcl/writer/sink.rs
```diff
@@ -0,0 +1,348 @@
...
+impl GclSink {
+    pub fn new(config: Config) -> Self {
+        Self {
+            client: None,
+            config,
+        }
+    }
+
+    #[cfg(test)]
+    #[allow(dead_code)] // NOTE Actually used for unit testing
```
Licenser, line 73: Since it's guarded behind `cfg(test)` this should no longer be needed
```suggestion

```

src/connectors/impls/gcl/writer/sink.rs
```diff
@@ -0,0 +1,348 @@
...
+            "The client is not connected",
+        ))?;
+
+        let mut entries = Vec::with_capacity(event.len());
+        for (data, meta) in event.value_meta_iter() {
+            let meta = meta.get_object("gcl_writer").or(None);
+            #[allow(clippy::cast_precision_loss)]
+            #[allow(clippy::cast_possible_wrap)]
+            let mut timestamp = Timestamp {
+                seconds: event.ingest_ns as i64 / 1_000_000_000i64,
```
Licenser, line 103:
```suggestion
                seconds: (event.ingest_ns / 1_000_000_000) as i64,
```

src/connectors/impls/gcl/writer/sink.rs
```diff
@@ -0,0 +1,348 @@
...
+                log_name: meta::default_log_name(&self.config),
+                resource: super::value_to_monitored_resource(self.config.resource.as_ref())?,
+                labels: meta::default_labels(&self.config),
+                entries,
+                partial_success: self.config.partial_success,
+                dry_run: self.config.dry_run,
+            })
+            .await;
+
+        match log_entries_response {
```
Licenser, line 121:
```suggestion
        match log_entries_response {
```

src/connectors/impls/gcl/writer.rs
darach: Yep. I'll likely rephrase those - they're verbatim from the proto files. I'll replace with doc comments for now

src/connectors/impls/gcl/writer.rs
darach: Oh it gets worse - the proto bufs define `LogSeverity.DEFAULT` which seems to map to `Info` from what I've seen in the logs explorer - probably want a steer from Zubin et al here

src/connectors/impls/gcl/writer/sink.rs
```diff
@@ -0,0 +1,348 @@
...
+            let mut timestamp = Timestamp {
+                seconds: event.ingest_ns as i64 / 1_000_000_000i64,
+                nanos: (event.ingest_ns % 1_000_000_000) as i32,
+            };
+            timestamp.normalize();
+            entries.push(value_to_log_entry(timestamp, &self.config, data, meta)?);
+        }
+
+        let log_entries_response = client
+            .write_log_entries(WriteLogEntriesRequest {
```
Agares, line 111: I think we should add a timeout here

src/connectors/impls/gcl/writer.rs
darach: I went for `LogSeverity.Default` and whacked the Option for now - we can tweak based on feedback when the connector is POC'd by Zubin et al

src/connectors/impls/gcl/writer.rs
```diff
@@ -0,0 +1,239 @@
...
+impl ConfigImpl for Config {}
+
+fn value_to_monitored_resource(
+    from: Option<&simd_json::OwnedValue>,
+) -> Result<Option<MonitoredResource>> {
+    match from {
+        None => Ok(None),
+        Some(from) => {
+            let vt = from.value_type();
+            match from {
```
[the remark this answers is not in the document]
darach: The value type is only used on error, it isn't used for matching - as such even if the representation changes, only the serialization is used in the error:

```rust
                        Some(labels) => labels
                            .as_object()
                            .ok_or_else(|| {
                                Error::from(ErrorKind::GclTypeMismatch("Value::Object", vt))
                            })?
```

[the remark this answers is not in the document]
Licenser: Ah yes https://github.com/simd-lite/value-trait/pull/25 will make this **a lot** easier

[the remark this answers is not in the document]
Licenser: the as object version looks a lot neater :)

src/connectors/impls/gcl/writer.rs
darach: Didn't attend to this one, not seeing how unwrap_or_default improves readability here

src/connectors/impls/gcl/writer/meta.rs
darach: This was a good change, and it simplified some tests and removed some untested branches for the win

src/connectors/impls/gcl/writer/sink.rs
darach: Added

src/connectors/impls/gcl/writer.rs
Licenser: It would allow collapsing the next statment down to:
```rust
                    let labels: HashMap<_, _> = from.try_get_object("labels")?
                            .cloned()
                            .unwrap_or_default()
                            .iter()
                            .map(|(key, value)| {
                                let key = key.to_string();
                                let value = value.to_string();
                                (key, value)
                            })
                            .collect();
```

Licenser: I just realiuze why this makes no sense :) in my mind #1819 is already merged and this code legal, right now it isn't so it can't seem useful yet

darach: Added #1819 to a comment to consider unknown_or_default in due course - closing for now

src/connectors/impls/gcl.rs
```diff
@@ -0,0 +1,77 @@
...
+use rand::Rng;
+use tremor_script::{tremor_fn, Registry, Value};
+
+pub(crate) mod writer;
+
+// The span ID within the trace associated with the log entry.
+//  For Trace spans, this is the same format that the Trace API v2
+//  uses: a 16-character hexadecimal encoding of an 8-byte array,
+// such as 000000000000004a
+fn random_span_id_value(ingest_ns_seed: u64) -> Value<'static> {
```
Agares, line 24: I know it's pedantic, and the probability of this happening is practically zero, but should we reroll the ID if it's all zeros? (since it's a special value)

Agares (approved):

src/connectors/impls/gcl.rs
```diff
@@ -0,0 +1,77 @@
...
+    let mut rng = tremor_common::rand::make_prng(ingest_ns_seed);
+    let span_id: String = (0..8)
+        .map(|_| rng.gen::<u8>())
+        .map(|b| format!("{:02x}", b))
+        .collect();
+    Value::from(span_id)
+}
+
+fn random_trace_id_value(ingest_ns_seed: u64) -> Value<'static> {
```
mfelsche, line 33: How are these different from `cncf::otel::gen_trace_id_string()` and the` span_id`-equivalent?It is nice to have them in this connector-scoped-namespace, but the implementations could be de-duplicated.

mfelsche (approved):
LGTM!

Everything totally straightforward.

src/connectors/impls/gcl.rs
darach: They refer to Cloud Trace not Otel

PR #1817, opened 2022-07-07
Initial work for pluggable logging - exploring tremor-pipeline

znwar:
Signed-off-by: Perle_nwar <rebeccaabli3@gmail.com>

# Pull request

## Related

## Checklist

* [x] The code is tested


tremor-pipeline/src/query.rs
```diff
@@ -660,6 +660,7 @@ impl Query {
                 insights: Vec::new(),
                 dot: format!("{}", dot),
                 metrics_channel: METRICS_CHANNEL.tx(),
+                logging_channel: LOGGING_CHANNEL.tx(),
```
Licenser, line 663: It might be worth considering a subscriber that subscribes directly to the logging framework instead of weaving the sender throughout the entire runtime - that could save a lot of effort and allow the `error!`, `info!`, `warn!` etc macro to be used without having to re-write every log message.

Licenser: marking this as resolve as there is now a appended :)

tremor-pipeline/src/executable_graph.rs
```diff
@@ -530,6 +535,18 @@ impl ExecutableGraph {
                     error!("Failed to send metrics: {}", e);
                 };
             }
+            for value in m.to_value(metric_name, &mut tags, ingest_ns) {
```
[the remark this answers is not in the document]
PerFleur: Yes, I was just trying out some stuff
I'll delete the unnecessary parts

tremor-cli/src/main.rs
```diff
@@ -126,6 +123,10 @@ async fn run(cli: Cli) -> Result<()> {
         Command::Run(r) => r.run().await,
         Command::Doc(d) => d.run(),
         Command::New { name } => create_template(std::env::current_dir()?, &name),
+        Command::PluggableLogging {} => {
```
Agares, line 126: We can remove this command, as pluggable logging is not a stand-alone thing, but works with the `--pluggable-logging` option for all other commands

tremor-pipeline/src/lib.rs
```diff
@@ -72,6 +76,45 @@ pub use op::{ConfigImpl, InitializableOperator, Operator};
...
+#[derive(Debug)]
+///Pluggable-logging Appender
+pub struct PluggableLoggingAppender {
+    /// async Sender
+    pub tx: Sender<LoggingMsg>,
+}
+
+impl Append for PluggableLoggingAppender {
+    fn append(&self, record: &log::Record) -> anyhow::Result<()> {
+        let vec = (r#"{"level": ""#.to_owned()
```
Agares, line 88: I think it would be better to use the `literal!()` macro here, so we get a `Value`, instead of creating a string.
This way can cause problems, eg. when the line contains `"`.
Also, please remove the `expect`, as Tremor should never panic in production, it can be replaced e.g. with `unwrap_or`

tremor-pipeline/src/lib.rs
```diff
@@ -131,13 +224,34 @@ impl MetricsMsg {
...
+            origin_uri,
+            language,
+        }
+    }
+}
 
 /// Sender for metrics
 pub type MetricsSender = Sender<MetricsMsg>;
 
+// TODO FIX ME NOTE Add LogsSender
```
Agares, line 246: I believe this comment is not needed anymore :)

tremor-pipeline/src/logging.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::errors::Result;
+use tremor_script::registry::Registry;
+use tremor_script::tremor_fn;
+
+/// Install's common functions into a registry
+///
+/// # Errors
+///  * if we can't install extensions
+pub fn load(reg: &mut Registry) -> Result<()> {
+    reg.insert(tremor_fn!(logging|info(_context) {
```
Agares, line 24: I think this should call the `info!` macro, we should also add functions for other levels (error, warn, debug, etc.)

tremor-script/src/registry.rs
```diff
@@ -118,7 +118,9 @@ pub fn registry() -> Registry {
         }))
         .insert(tremor_fn!(system|version(_context) {
             Ok(Value::String(VERSION.into()).into_static())
-        }));
+        }))
+        .insert(tremor_fn!(logging|info(_context) {
```
Agares, line 122: I think this function is already being registered in `tremor-pipeline/src/logging.rs`, so no need to register it twice.

Agares (changes requested):
Great work! I left some comments, but those should hopefully not be too hard to resolve :)

Cargo.toml
```diff
@@ -84,6 +84,7 @@ indexmap = { version = "1", features = ["serde-1"] }
 lazy_static = "1"
 libflate = "1.2"
 log = { version = "0.4", features = ["kv_unstable"] }
+log4rs = "1"
```
Licenser, line 87: I think this dev dependency not a runtime dependency, or did I missed its use outside of tests?

src/connectors/impls/logging.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+#[derive(Debug, Default)]
+pub(crate) struct Builder {}
+
+#[async_trait::async_trait]
+impl ConnectorBuilder for Builder {
+    fn connector_type(&self) -> ConnectorType {
+        "logging".into()
```
Licenser, line 40: We call the `metrics` connector `metrics` to follow the same naming scheme for consistency would mean `logs` is a better name for this connector.

src/connectors/impls/logging.rs
```diff
@@ -0,0 +1,127 @@
...
+    rx: Receiver<LoggingMsg>,
+    origin_uri: EventOriginUri,
+}
+
+impl LoggingSource {
+    pub(crate) fn new(rx: Receiver<LoggingMsg>) -> Self {
+        Self {
+            rx,
+            origin_uri: EventOriginUri {
+                scheme: "tremor-logging".to_string(),
```
Licenser, line 88: same as with the connector name `tremor-logs` would be a more expected name given we already have `tremor-metrics`

src/connectors/impls/logging.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+#[async_trait::async_trait()]
+impl Source for LoggingSource {
+    async fn pull_data(&mut self, _pull_id: &mut u64, _ctx: &SourceContext) -> Result<SourceReply> {
+        let msg = self
+            .rx
+            .recv()
+            .await
+            .map_err(|e| Error::from(format!("error: {}", e)))?;
```
Licenser, line 104: @mfelsche discovered recently that the `OverFlow` error has to be handle with a  retry on pull, this was a bug he fixed in #1912

src/connectors/tests/logging.rs
```diff
@@ -0,0 +1,135 @@
...
+    config::{Appender, Config, Root},
+    Logger,
+};
+use serial_test::serial;
+use tremor_pipeline::{PluggableLoggingAppender, LoggingChannel, LoggingMsg};
+use tremor_script::literal;
+
+fn append_channel() -> (Logger,  Receiver<LoggingMsg>) {
+
+	let ( tx, rx) = broadcast(128);
```
Licenser, line 28: <3

tremor-cli/src/cli.rs
```diff
@@ -53,6 +56,8 @@ pub(crate) enum Command {
         #[clap( value_parser = clap::value_parser!(String))]
         name: String,
     },
+    /// Pluggable-logging
```
Licenser, line 59: I don't think we want to introduce a whole subcommand for logging, this will confuse users as to what it does and provide very little benefit to them. I can totally see that it was helpful in development, but before merging this it should probably be deleted.

tremor-cli/src/main.rs
```diff
@@ -16,12 +16,7 @@
 // This isn't a external crate so we don't worry about docs
 // #![deny(missing_docs)]
 #![recursion_limit = "1024"]
-#![deny(
```
Licenser, line 19: Sorry those have to stay

tremor-cli/src/main.rs
```diff
@@ -107,8 +103,9 @@ async fn main() -> Result<()> {
 }
 
 async fn run(cli: Cli) -> Result<()> {
-    // Logging
-    if let Some(logger_config) = &cli.logger_config {
+    if cli.pluggable_logging {
```
Licenser, line 106: I'm not sure we want them to be exclusive, I can totally see a user wanting to have a rotating file along with ingesting the metrics to a pipeline, but I could make a use-case up that doesn't exit.

If they are exclusive, we definitely want to error when both are provided however otherwise users will have to deal with their logging config silently being ignored and that'd be bad :tm:

Looking at the code wouldn't it make more sense to allow `tremor` to be caused as an appended so the user can configure it? Having it forced to always be info for everything seems like a problem in usability, most of the time users will want to limit logs to `warning` or `error` to keep down noise. 

Looking at the other code, guardingthis is quite tricky as the functions and connectors still would be present, just not usable, the more I dig into it the more I get the impression this perhaps should be always on.

tremor-pipeline/src/executable_graph.rs
```diff
@@ -809,6 +813,7 @@ mod test {
             ops: vec![Value::null(), Value::null(), Value::null(), Value::null()],
         };
         let mut rx = METRICS_CHANNEL.rx();
+        let _rx_ = LOGGING_CHANNEL.rx();
```
Licenser, line 816: We don't seem to need this any more

tremor-pipeline/src/executable_graph.rs
```diff
@@ -723,7 +727,7 @@ mod test {
             config: NodeConfig::default(),
         }
     }
-
+    //RE
```
Licenser, line 730:
```suggestion
```

tremor-pipeline/src/lib.rs
```diff
@@ -19,7 +19,6 @@
 #![recursion_limit = "1024"]
 #![deny(
     clippy::all,
-    clippy::unwrap_used,
```
Licenser, line 22: We need to keep this

tremor-pipeline/src/lib.rs
```diff
@@ -44,6 +44,7 @@ use std::fmt;
 use std::fmt::Display;
 use std::iter::Iterator;
 use std::str::FromStr;
+//use std::{fmt::Debug, mem, pin::Pin, sync::Arc};
```
Licenser, line 47:
```suggestion
```

tremor-pipeline/src/lib.rs
```diff
@@ -90,12 +133,19 @@ pub struct MetricsChannel {
     tx: Sender<MetricsMsg>,
     rx: Receiver<MetricsMsg>,
 }
+/// Channel for plugging-logging messages
+pub struct LoggingChannel {
```
Licenser, line 137: Since `LoggingChannel` and `MetricsChannel` are effectively the same it would be nicer to implement them as:

```rust
struct OverflowingChannel<Msg> {
    tx: Sender<Msg>,
    rx: Receiver<Msg>
}
```

instead of having two identical copies of code

tremor-pipeline/src/lib.rs
```diff
@@ -121,6 +194,26 @@ pub struct MetricsMsg {
     pub origin_uri: Option<EventOriginUri>,
 }
 
+#[derive(Debug, Clone)]
+/// Language from which the logs are coming (Rust, Tremor, etc.)
+pub enum LanguageKind {
```
Licenser, line 199: Perhaps `LogSource` would be a better name here as it has nothing really to do with languages

tremor-pipeline/Cargo.toml
```diff
@@ -32,6 +32,8 @@ tremor-value = { version = "0.12.4", path = "../tremor-value" }
 url = "2.2.2"
 value-trait = "0.4"
 window = { git = "https://github.com/tremor-rs/window.git", tag = "v0.1.1" }
+anyhow = "1.0.58"
```
Licenser, line 35: Are those two unused dependencies, dev dependencies or actual runtime dependencies? I did not see them in the code.

tremor-cli/tests/integration/system-metrics/config.troy
```diff
@@ -6,7 +6,7 @@
 # - exit the flow when we collected metrics for a certain amount of events that passed through source, pipeline and sink
 define flow main
 flow
-  use integration;
+ 	use integration;
```
Licenser, line 9:
```suggestion
  use integration;
```

Licenser (commented):
This is overall great work 👍 just a few minor cleanups left :D, well done!

tremor-pipeline/Cargo.toml
znwar: Sorry but I didn't understand a comment.
If I delete **anyhow** from t**remor_pipeline/Cargo.toml** I would get an error in /**tremor-pipeline/src/lib.rs**  because the appender return a **anyhow::Result<()>**

Licenser: That answers the question :) so it's a full dependency! all good then :D

tremor-script/src/registry.rs
```diff
@@ -853,22 +853,31 @@ impl Aggr {
...
 
-    // Test utility to grab a function from the registry
-    pub fn fun<'event>(m: &str, f: &str) -> impl Fn(&[&Value<'event>]) -> FResult<Value<'event>> {
-        let f = registry()
-            .find(m, f)
-            .expect("could not find function")
-            .clone();
-        move |args: &[&Value]| -> FResult<Value> { f.invoke(&EventContext::new(0, None), args) }
-    }
+    // // Test utility to grab a function from the registry
```
Agares, line 873: If this code is not needed, let's delete it, commented out code loses its meaning quickly.

tremor-pipeline/src/lib.rs
```diff
@@ -73,6 +77,54 @@ pub use op::{ConfigImpl, InitializableOperator, Operator};
...
+            "line": record.line()
+        });
+
+        let msg = LoggingMsg {
+            language: lang,
+            payload: EventPayload::from(vec),
+            origin_uri: None,
+        };
+
+        println!("{:?}", msg);
```
Agares, line 117: do we need this `println!`, or was it only for debugging?

tremor-pipeline/src/lib.rs
```diff
@@ -73,6 +77,54 @@ pub use op::{ConfigImpl, InitializableOperator, Operator};
...
+        block_on(self.tx.broadcast(msg))?;
+        Ok(())
+    }
+
+    fn flush(&self) {
+        todo!("TODO logging impl Append for PluggableLoggingAppender :: fn flush");
```
Agares, line 123: Let's get rid of this `todo!` (I believe this function can be left empty, as the logs are going straight into the channel in `append`)

tremor-pipeline/src/lib.rs
znwar: No, it was for debugging, indeed

tremor-script/src/registry.rs
znwar: I really don't know why there are changes there, I don't remember making them

znwar: Oh my bad, the function was just moved above sorry :woman_facepalming:

tremor-cli/src/logging.rs
```diff
@@ -12,22 +12,31 @@
...
 
-    let config = Config::builder()
+    let config = match Config::builder()
         .appender(Appender::builder().build("stdout", Box::new(stdout)))
         .build(Root::builder().appender("stdout").build(LevelFilter::Debug))
-        .unwrap(); // TODO pluggable_logging see if unwrap should be removed
+    {
+        Ok(config) => config,
+        Err(_) => {
+            return Err("config could be built".into());
```
Licenser, line 34:
```suggestion
            return Err("log4rs config could not be built".into());
```

tremor-pipeline/src/logging.rs
```diff
@@ -149,7 +157,7 @@ impl TremorFn for PluggableLoggingFm {
                 err => err,
             }
         } else {
-            Err(FunctionError::RuntimeError{mfa: this_mfa(), error: "expected 1st parameter to format to be a format specifier e.g. to print a number use `string::format(\"{}\", 1)`".into()})
+            Err(FunctionError::RuntimeError{mfa: this_mfa(), error: "expected 1st parameter to format to be a format specifier e.g. to print a number use `string::format(\"{}\", 1)`".to_string()})
```
Licenser, line 160:
```suggestion
            Err(FunctionError::RuntimeError{mfa: this_mfa(), error: format!("expected 1st parameter to format to be a format specifier e.g. to print a number use `logging::{}(\"{{}}\", 1)`", self.name)})
```

CHANGELOG.md
```diff
@@ -104,6 +104,7 @@
 - Add `gcl_writer` Google Cloud Platform Cloud Logging connector for writing log entries, with in-flight concurrency support.
 - Add reverse functions and tests for arrays and strings, add sort test for arrays
 - Add a ClickHouse connector
+- Add a Logging connector
```
Licenser, line 107: This should move up to Unreleased

src/connectors/impls/logging.rs
```diff
@@ -0,0 +1,137 @@
...
+impl Source for LoggingSource {
+    async fn pull_data(&mut self, _pull_id: &mut u64, _ctx: &SourceContext) -> Result<SourceReply> {
+        loop {
+            match self.rx.recv().await {
+                Ok(msg) => {
+                    break Ok(SourceReply::Structured {
+                        payload: msg.payload,
+                        origin_uri: msg.origin_uri.unwrap_or_else(|| self.origin_uri.clone()),
+                        stream: DEFAULT_STREAM_ID,
+                        port: None,
```
Licenser, line 107: Not a change request but perhaps we can use port for levels? perhaps we can add a TODO comment for this?

tremor-pipeline/src/lib.rs
Licenser: this is moved down now :)

tremor-pipeline/src/lib.rs
```diff
@@ -73,6 +77,51 @@ pub use op::{ConfigImpl, InitializableOperator, Operator};
...
+#[derive(Debug, Clone)]
+///Pluggable-logging Appender
+pub struct PluggableLoggingAppender {
+    /// async Sender
+    pub tx: Sender<LoggingMsg>,
+}
+
+impl Append for PluggableLoggingAppender {
+    fn append(&self, record: &log::Record) -> anyhow::Result<()> {
+        let path: String;
```
Licenser, line 89: perhaps path could be Option<String> that way it's easily recognizable in the code as `null` or `string` to see if it was set :)

Licenser (commented):
Super nice!

src/connectors/impls/logging.rs
znwar: I think it is not possible to make this change (I tried it) because this field "port" from `SourceReply` has a default value of "out" (modulo wrappers like `Option`, `Cow`, etc...) and is already used.
When launching the test "pluggable_logging_operator" with the tremor server running, there is just no output.

Licenser: All good a `TODO` around it so we remember to investigate later is good enough :)

Licenser:
It seems like something went very wrong with rebasing o.O

tremor-pipeline/src/executable_graph.rs
```diff
@@ -257,6 +257,10 @@ pub struct ExecutableGraph {
     pub(crate) last_metrics: u64,
     pub(crate) metric_interval: Option<u64>,
     pub(crate) metrics_channel: MetricsSender,
+
+    #[allow(dead_code)]
```
Licenser, line 261: let's remove the allows

```suggestion
```

tremor-pipeline/src/executable_graph.rs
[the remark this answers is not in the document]
Licenser: this is updated so resolving this

tremor-cli/src/main.rs
Licenser: This seems to be gone, can we resolve this?

tremor-pipeline/src/lib.rs
Licenser: this seems to be fixed can we resolve this?

tremor-pipeline/src/logging.rs
Licenser: this seems to be fixed can we resolve this?

tremor-script/src/registry.rs
Licenser: this seems to be fixed can we resolve this?

src/connectors/impls/logging.rs
Licenser: this is named consistently now :) resolving the comment :D

tremor-cli/src/logging.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020-2021, The Tremor Team
```
Licenser, line 1: I think this file can go, we no longer need it :)

tremor-cli/src/cli.rs
Licenser: resolving the comment :D, nicely done!

tremor-pipeline/src/executable_graph.rs
Licenser: this is cleaned up thank you!

tremor-pipeline/src/lib.rs
Licenser: This too is updated, nice thank you  again :)

tremor-pipeline/src/lib.rs
Licenser: also cleaned up 👍

tremor-script/src/registry.rs
Licenser: this seems to be fixed, can we resolve the comment?

tremor-pipeline/src/lib.rs
Licenser: this seems to be fixed, can we resolve the comment?

tremor-pipeline/src/lib.rs
Licenser: this seems to be fixed, can we resolve the comment?

CHANGELOG.md
Licenser: thanks for moving it up!

src/connectors/impls/logging.rs
Licenser: perfect 👍

tremor-pipeline/src/lib.rs
Licenser: nicely done 👍 code looks much simpler now too :D

tremor-cli/src/logging.rs
znwar: This file is used in the main.rs

tremor-cli/src/main.rs
Licenser: yay denies are back!

Licenser (dismissed):

znwar:
I don't understand why some tests fail 

> A lot of tests and checks are still failing, we can't merge this until the CI passess

Licenser:
The elastic search tests did fail for me today as well, I think there seems to be something off about them that isn't related to your code. What we could do is creat a branch for you to merge into and then take fixing the ES tests from there on, the rest looks good 👍 so it shouldn't hold you back to be honest.

Only needs the DCO fixed ( which should be easy the commands are here: https://github.com/tremor-rs/tremor-runtime/pull/1817/checks?check_run_id=11054787211 )

Licenser:
don't mind the conflict I'll sort that once the DCO is green :) and then merge it

znwar:
@Licenser I have too many conflicts

Licenser:
You can ignore the conflicts, I think things got messed up when merging main instead of rebasing. But we can't merge contributions that are not signed off with a DCO

znwar:
So, what do I do?
Do I rebase the whole stack of commits containing the ones that I did not sign-off with something like
```
git rebase HEAD~202 -i --rebase-merges
```
and then this on conflicts:
```
git rebase --skip
```
and this on the said commit to sign-off?
```
git commit --sign-off
```

Licenser:
Ja exactly, that should do the trick 👍

znwar:
Okay but... `git` still stop on "bump" commits and other things as it says that there are conflicts (with `Cargo.lock` and other files sometimes)
I am confused: do I really have to skip maybe a few hundred of conflicts, if not a thousand, each with the skip command, or is there a quicker and more efficient way to do it?

Licenser:
I'll close this since the new PR was merged 👍 great job!

PR #1827, opened 2022-07-11
Allow kafka connectors to detect some more errors and reconnect.

mfelsche:
Also refactor kafka contexts a bit and deduplicate implementation.

# Pull request

## Checklist

* [x] The code is tested

## Performance

nope.

### DIFF
--- a/src/connectors/impls/kafka.rs
+++ b/src/connectors/impls/kafka.rs
@@ -24,7 +24,7 @@ use beef::Cow;
 use core::future::Future;
 use futures::future;
 use halfbrown::HashMap;
-use rdkafka::{error::KafkaError, util::AsyncRuntime, ClientContext};
+use rdkafka::{error::KafkaError, util::AsyncRuntime, ClientContext, Statistics};
 use rdkafka_sys::RDKafkaErrorCode;
 use std::{
     sync::atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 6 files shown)


src/connectors/impls/kafka.rs
```diff
@@ -76,3 +88,366 @@ fn is_failed_connect_error(err: &KafkaError) -> bool {
...
+            active: AtomicBool::new(true),
+        }
+    }
+
+    fn on_connection_lost(&self) -> JoinHandle<()> {
+        let ctx = self.ctx.clone();
+
+        // only actually notify the notifier if we didn't do so before
+        // otherwise we would flood the connector and reconnect multiple times
+        if self.active.fetch_and(false, Ordering::AcqRel) {
```
Agares, line 127: I think `swap` here would do the same thing as `fetch_and`, but be a bit more readable

src/connectors/impls/kafka.rs
```diff
@@ -76,3 +88,366 @@ fn is_failed_connect_error(err: &KafkaError) -> bool {
...
+                info!(target: "librdkafka", "{} librdkafka: {} {}", self.ctx, fac, log_message);
+            }
+            rdkafka::config::RDKafkaLogLevel::Debug => {
+                debug!(target: "librdkafka", "{} librdkafka: {} {}", self.ctx, fac, log_message);
+            }
+        }
+    }
+
+    // TODO: add full connector id to tags
+    #[allow(clippy::cast_sign_loss)]
```
Agares, line 169: Can you add a short comment here explaining why this is allowed?

Agares (approved):

PR #1831, opened 2022-07-13
Include flow-alias in pipeline- and connector-alias

### LINKED ISSUE #1802: Identify connectors correctly in metrics
Currently in metrics messages from connectors (and pipelines), we only track the connector type, not the connector instance name. We also do not track which flow instance this connector is part of, so multiple connectors of the same type get conflated, which makes metrics not as useful as they could be, as we have no granularity and precise measurements.

Also when looking at the logs, currently we only report the connector alias, not the flow alias a connector is part of. So logs are not as useful as they could be.

Track the flow instance id and the connector instance id in metrics and logs. The same is true for pipelines.

mfelsche:
# Pull request

## Description

Include flow-alias in pipeline- and connector-alias for deduplicating log and metrics output when multiple flows with the same connector aliases are deployed.

## Related

* Related Issues: fixes #1802 

## Checklist

* [x] The code is tested

### DIFF
--- a/src/connectors.rs
+++ b/src/connectors.rs
@@ -36,6 +36,7 @@ use self::utils::quiescence::QuiescenceBeacon;
 pub(crate) use crate::config::Connector as ConnectorConfig;
 use crate::instance::State;
 use crate::pipeline;
+use crate::system::flow::ConnectorAlias;
 use crate::system::{KillSwitch, World};
 use crate::{
     errors::{Error, Kind as ErrorKind, Result},
@@ -67,8 +68,8 @@ pub(crate) const ACCEPT_TIMEOUT: Duration = Duration::from_millis(100);
 /// connector address
 #[derive(Clone, Debug)]
 pub struct Addr {
-    /// connector instance url
-    pub(crate) alias: String,
+    /// connector instance alias
+    pub(crate) alias: ConnectorAlias,
     sender: Sender<Msg>,
     source: Option<SourceAddr>,
     pub(crate) sink: Option<SinkAddr>,
@@ -225,8 +226,8 @@ pub(crate) enum Msg {
 /// result of an async operation of the connector.
 /// bears a `url` to identify the connector who finished the operation
 pub(crate) struct ConnectorResult<T: std::fmt::Debug> {
-    /// the connector url
-    pub(crate) alias: String,
+    /// the connector alias
+    pub(crate) alias: ConnectorAlias,
     /// the actual result
     pub(crate) res: Result<T>,
 }
@@ -249,8 +250,8 @@ impl ConnectorResult<()> {
 
 /// context for a Connector or its parts
 pub(crate) trait Context: Display + Clone {
-    /// provide the url of the connector
-    fn alias(&self) -> &str;
+    /// provide the alias of the connector
+    fn alias(&self) -> &ConnectorAlias;
 
     /// get the quiescence beacon for checking if we should continue reading/writing
     fn quiescence_beacon(&self) -> &QuiescenceBeacon;
@@ -316,8 +317,8 @@ pub(crate) trait Context: Display + Clone {
 /// connector context
 #[derive(Clone)]
 pub(crate) struct ConnectorContext {
-    /// url of the connector
-    pub(crate) alias: String,
+    /// alias of the connector instance
+    pub(crate) alias: ConnectorAlias,
     /// type of the connector
     connector_type: ConnectorType,
     /// The Quiescence Beacon
@@ -333,7 +334,7 @@ impl Display for ConnectorContext {
 }
 
 impl Context for ConnectorContext {
-    fn alias(&self) -> &str {
+    fn alias(&self) -> &ConnectorAlias {
         &self.alias
     }
 
@@ -353,7 +354,9 @@ impl Context for ConnectorContext {
 /// Connector instance status report
 #[derive(Debug, Serialize)]
 pub struct StatusReport {
-    /// connector instance url
+    /// connector alias without the containing flow alias
+    /// when we report this report will happen in the context of a flow,
+    /// so we don't need to provide that extra context here again
     pub(crate) alias: String,
     /// state of the connector
     pub status: State,
@@ -397,7 +400,7 @@ pub(crate) type Known =
 /// # Errors
 /// if the connector can not be built or the config is invalid
 pub(crate) async fn spawn(
-    alias: &str,
+    alias: &ConnectorAlias,
     connector_id_gen: &mut ConnectorIdGen,
     builder: &dyn ConnectorBuilder,
     config: ConnectorConfig,
@@ -405,13 +408,7 @@ pub(crate) async fn spawn(
 ) -> Result<Addr> {
     // instantiate connector
     let connector = builder.build(alias, &config, kill_switch).await?;
-    let r = connector_task(
-        alias.to_string(),
-        connector,
-        config,
-        connector_id_gen.next_id(),
-    )
-    .await?;
+    let r = connector_task(alias.clone(), connector, config, connector_id_gen.next_id()).await?;
 
     Ok(r)
 }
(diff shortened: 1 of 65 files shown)


src/connectors.rs
```diff
@@ -353,7 +354,9 @@ impl Context for ConnectorContext {
 /// Connector instance status report
 #[derive(Debug, Serialize)]
 pub struct StatusReport {
-    /// connector instance url
+    /// connector alias without the containing flow alias
+    /// when we report this report will happen in the context of a flow,
+    /// so we don't need to provide that extra context here again
     pub(crate) alias: String,
```
Licenser, line 360: For consistency would it make sense to still carry over the Alias here instead of a `String`?

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -390,13 +390,14 @@ impl Sink for GbqSink {
...
 
+        let interceptor_ctx = ctx.clone();
         let mut client = BigQueryWriteClient::with_interceptor(
             channel,
             AuthInterceptor {
                 token: Box::new(move || match token.header_value() {
                     Ok(val) => Ok(val),
                     Err(e) => {
-                        error!("Failed to get token for BigQuery: {}", e);
+                        error!("{interceptor_ctx} Failed to get token for BigQuery: {}", e);
```
Licenser, line 400:
```suggestion
                        error!("{interceptor_ctx} Failed to get token for BigQuery: {e}");
```

src/connectors/impls/kv.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
         Codec,
     },
     connectors::prelude::*,
-    errors::err_conector_def,
+    errors::err_connector_def,
```
Licenser, line 22: this error should probably have been in the connectors prelude :)

src/pipeline.rs
```diff
@@ -439,25 +444,43 @@ fn maybe_send(r: Result<()>) {
     }
 }
 
+struct PipelineContext {
```
Licenser, line 447: What's the difference between `PipelineContext` and `PipelineAlias`, I'm a bit confused about that.

src/system/flow.rs
```diff
@@ -27,74 +27,110 @@ use async_std::{
...
+    ast::{self, ConnectStmt, DeployFlow, Helper},
     errors::not_defined_err,
 };
 
 /// unique identifier of a flow instance within a tremor instance
-#[derive(Debug, PartialEq, PartialOrd, Eq, Hash, Clone, Serialize)]
-pub(crate) struct Id(pub(crate) String);
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, PartialEq, PartialOrd, Eq, Hash, Clone, Serialize, Deserialize)]
+pub struct FlowAlias(String);
```
Licenser, line 41: Should the flow alias perhaps include it's "type" (or definition name) so that it's easier to distinguish where it came from? It could render to something like `"my_alias<my_defition>"` or `"my_definition::my_alias"` or a bunch of other things, but having this type would allow to convey both, the definition name and the created alias for it

src/system/flow.rs
```diff
@@ -27,74 +27,110 @@ use async_std::{
...
-    fn from(e: &DeployEndpoint) -> Self {
-        ConnectorAlias(e.alias().to_string())
+impl ConnectorAlias {
+    /// construct a new `ConnectorId` from the id of the containing flow and the connector instance id
+    pub fn new(flow_alias: FlowAlias, connector_alias: impl Into<String>) -> Self {
+        Self {
+            flow_alias,
+            connector_alias: connector_alias.into(),
+        }
     }
```
Licenser, line 88: Changing the new definition would make a lot of the `new` calls easier through the files as they could be passed something like `ConnectorAlias::new("flow", "connector")` in the tests :) 
```suggestion
    pub fn new(flow_alias: T, connector_alias: impl Into<String>) -> Self
    where
      T: Into<FlowAlias> {
        Self {
            flow_alias: flow_alias.into(),
            connector_alias: connector_alias.into(),
        }
    }
```

src/system/flow.rs
```diff
@@ -640,7 +688,7 @@ async fn spawn_task(
                 }
 
                 MsgWrapper::DrainResult(conn_res) => {
-                    info!("[Flow::{}] Connector {} drained.", &alias, &conn_res.alias);
+                    info!("[Flow::{}] Connector {} drained.", &id, &conn_res.alias);
```
Licenser, line 691:
```suggestion
                    info!("[Flow::{id}] Connector {} drained.", &conn_res.alias);
```

src/connectors.rs
mfelsche: This struct is directly serialized as part of our API response. The connectors are always reported in the context of a flow, so i think their aliases should be listed without the flow alias here, hence only a string. I could add the `ConnectorId` here but then would have to introduce a new struct for the API which serializes to the expected format.

src/pipeline.rs
mfelsche: We also have a `ConnectorContext`, which is providing contextual information and some utils to every connector and is passed in to literally all calls. I wanted to have something similar here. This context though is only used for printing the correct prefix for log messages: `[Pipeline::flow::pipeline_alias] ...`

src/connectors.rs
Licenser: Oh I understand the problem, thanks for the explenation. Would [`#[serde(serialize_with = "path")]`](https://serde.rs/field-attrs.html#serialize_with) perhaps work? We could keep it a `ConnectorId` and won't need to introduce a special struct, just a function that serializes it 


```rust
serialize_no_flow<S>(x: &ConnectorId, s: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    s.serialize_str(x.connector_alias)
}
```

src/pipeline.rs
Licenser: Perhaps to make that clearer type wise `PipelineContext` could contain `PipelineId` instead of the `String`? That way it would become obvious that The context is a higher level abstraction that could potentially grow outside of the pure ID but is just that for now.

src/system/flow.rs
mfelsche: done that with an `impl Into<FlowAlias>`

src/system/flow.rs
mfelsche: This is a good idea. this will make tracking instances from metrics back to their definition much easier. We already use the `::` for separating the hierarchical parts of flow and pipeline/connector `flow_alias::pipeline_alias`, so this should be printed differently, maybe: `flow_alias::pipeline_alias from pipeline_definition_name` ? Or `pipeline_definition_name<flow_alias::pipeline_alias>`?

src/connectors.rs
mfelsche: Then i'd rather provider another struct just for the API.

src/system/flow.rs
Licenser: Ja, you've got a point `::` is a bad choice, while `<>` reads kind of "right" from the rust perspective I honestly like your suggestion of `from` as it mirrors the way we define things :+1: for consistency :)

src/system/flow.rs
Licenser: Sorry, this was about something different, not that `From<>` etc exists  but that the `new` function tales a `Into<Flow Alias>` as as first argument.

It's also worth noting that there is a difference between a the generic version `T: Iinto...` and `impl Into<>` as one is compile time resolved the other at runtime, but I'm not sure it's really worth stressing out about perf in this instance OTOH it might be good not getting into the habit of `impl ...` if we don't have to

mfelsche: as far as i know the `impl` in argument position is just syntactic sugar for a generic type param.
The reference says it:

> That is, impl Trait in argument position is syntactic sugar for a generic type parameter like <T: Trait>, except that the type is anonymous and doesn't appear in the [GenericParams](https://doc.rust-lang.org/reference/items/generics.html) list.

So I think it is not really handled at runtime, but the specialization during compile time is happening the same, isn't it?

mfelsche: Also i put an `impl Into<FlowAlias>` into the first argument.

src/system/flow.rs
mfelsche: I tried it and it gets really too long for logs to remain readable. I would add this as another issue and discuss it there?

PR #1837, opened 2022-07-15
Kafka - Initial attempt at doing less calls to the group coordinator.

mfelsche:
# Pull request

## Description
Avoid commit calls to the group-coordinator for each event. Instead make use of the client local offset store as the only way to handle commits, so the traffic from them is bounded.

Try to avoid unnecessary calls for the current assignment, by caching it.

## Checklist

* [ ] The code is tested

### DIFF
--- a/src/connectors/impls/kafka/consumer.rs
+++ b/src/connectors/impls/kafka/consumer.rs
@@ -144,8 +144,10 @@ impl Mode {
                 let interval_ms = format!("{}", Duration::from_nanos(*commit_interval).as_millis());
                 client_config
                     .set("enable.auto.commit", "true")
-                    .set("enable.auto.offset.store", "false")
-                    .set("auto.commit.interval.ms", interval_ms.as_str());
+                    .set("enable.auto.offset.store", "false");
+                if *commit_interval > 0 {
+                    client_config.set("auto.commit.interval.ms", interval_ms.as_str());
+                }
             }
             Mode::Custom {
                 rdkafka_options, ..
@@ -233,8 +235,7 @@ impl ConnectorBuilder for Builder {
             path: vec![],
         };
 
-        let tid = task::current().id();
-        let client_id = format!("tremor-{}-{}-{:?}", hostname(), alias, tid);
+        let client_id = format!("tremor-{}-{}", hostname(), alias);
         let mut client_config = config.mode.to_config();
 
         // we do overwrite the rdkafka options to ensure a sane config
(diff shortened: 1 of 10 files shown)


src/connectors/impls/kafka/consumer.rs
```diff
@@ -37,18 +39,30 @@ const KAFKA_CONSUMER_META_KEY: &str = "kafka_consumer";
...
-    pub brokers: Vec<String>,
+    brokers: Vec<String>,
+
+    /// Whether or not to commit offsets of received messages immediately
+    /// or only upon received ack messages.
+    ///
+    /// If `auto_commit` is set to `false`, events emitted from this consumer
+    /// are considered transactional, soliciting an ack/fail message upon successful or failed handling downstream.
+    #[serde(default = "default_true")]
+    auto_commit: bool,
```
Licenser, line 56: I'm not sure it's a good idea to pull out more variables form `config` in here and have them definable in two places, that could quickly lead to confusion.

It also means that our defaults might not be in sync with the defaults from librdkafka leading to unexpected results - unless we make sure that we're always in sync which we will not.

src/connectors/impls/kafka/consumer.rs
```diff
@@ -329,14 +386,48 @@ impl KafkaConsumerSource {
...
             source_rx,
             consumer: None,
             consumer_task: None,
             metrics_rx: None,
+            last_rebalance_ts: Arc::new(AtomicU64::new(0)),
+            cached_assignment: None,
+        }
+    }
+
+    /// gets the current assignment from the cache or fetches it from the group coordinator
```
Licenser, line 402: I might not have the full context here, but from what I understand, this really sits wrong with me, it feels like we're trying to outsmart kafka/librkafka and make it do something that it's not meant to do.

The problem this is trying to fix, from what I can tell, is a quick succession of CB-open CB-close events. Would be changing the CB system to prevent these floppy changes a option, it seems like behaviour like that is potentially harmful to all connectors and not just this one.

darach: Yep - we could separate this out so that any connector can use a configurable failure detector which would make the fix generally available to all connectors.

src/connectors/impls/kafka/consumer.rs
mfelsche: We overwrite user config with the settings coming from this and if we overwrite user setting, we issue a warning. We ensure the client working sensibly, by enforcing `enable.auto.commit` no matter what. It is meant to prevent people shooting themselves in the foot and allowing them to setup their kafka connector without the need to dive into librdkafka configs.

We might want to add a third mode where we do the current behaviour and send a commit for every single message to the group coordinator.

Another way we could work with the config is to simply look if people have enabled auto commit in their kafka config and store the offsets instead of sending them to the coordinator.

Licenser:
Since this will break stuff lets point it to 0.13-staging?

PR #1862, opened 2022-07-29
Error improvements to pipelines

carolgeng:
# Pull request

## Description

Added clear messages for errors in pipelines when the input `port` doesn't exist, added `port` to `ConnectInput` and weaved them together with existing code

## Checklist

* [ ] The code is tested


src/pipeline.rs
```diff
@@ -572,10 +574,15 @@ pub(crate) async fn pipeline_task(
...
             AnyMsg::Mgmt(MgmtMsg::ConnectInput {
                 endpoint,
+                port,
                 target,
                 is_transactional,
             }) => {
-                info!("{ctx} Connecting {endpoint} to port 'in'");
+                info!("{ctx} Connecting '{endpoint}' to port '{port}'");
+                if !input_does_exist(&port, &pipeline) {
+                    error!("{ctx} Error connecting input pipeline '{port}' as it does not exist",);
```
mfelsche, line 583: If we add a sender to the `ConnectInput` for sending back a result, like we do in `Inspect` below, we could completely error the startup and not just emit a log message. The code sending that message would need to create a channel and wait for a response then.

carolgeng: Thank you for the suggestion Mattias! I have added a transmitter and receiver in `ConnectInput`

Licenser (dismissed):

CHANGELOG.md
```diff
@@ -32,6 +38,8 @@
 - Fix off-by-one error in `kafka_consumer` committing offsets, thus replaying the last committed event.
 - Allow `kafka_consumer` connector to reconnect upon more error conditions and avoid stalls.
 - Include flow alias in pipeline and connector aliases reported via metrics events and logging in order to deduplicate entries
+- Add clear error messages for pipelines for console in, pipeline out, pipeline in
```
mfelsche, line 41: should this be in the `0.13.0-rc.1` section? Or should the above be down here as well?

src/pipeline.rs
```diff
@@ -230,8 +230,12 @@ pub(crate) fn spawn(
 pub(crate) enum MgmtMsg {
     /// input can only ever be connected to the `in` port, so no need to include it here
```
mfelsche, line 231: this comment is not true anymore, is it?

src/connectors.rs
```diff
@@ -632,6 +632,7 @@ async fn connector_task(
                                 .addr
                                 .send(SourceMsg::Link {
                                     port,
+                                    tx: result_tx.clone(),
```
mfelsche, line 635: you need to be careful here. Below in line 657 the result of the `send` operation of the `SourceMsg::Link` is sent to the `result_tx`. The result of the `Link` operation is not waited for, if done like this.
I'd suggest to only send a result to `result_tx` if we get into an errorcondition here. And in the case where the `Link` was correctly sent, do nothing, the one who created the `LinkOutput` msg will wait on its receiving side of the `result_tx` for the one message saying it is all done.

src/connectors/source.rs
```diff
@@ -834,21 +840,24 @@ where
...
-    ) -> Control {
+    ) -> Result<Control> {
         let pipes = if port.eq_ignore_ascii_case(OUT.as_ref()) {
             &mut self.pipelines_out
         } else if port.eq_ignore_ascii_case(ERR.as_ref()) {
             &mut self.pipelines_err
         } else {
             error!("{} Tried to connect to invalid port: {}", &self.ctx, &port);
-            return Control::Continue;
+            return Ok(Control::Continue);
```
mfelsche, line 852: make sure that in every case, be it error or success, we send a `Result<()>` to the `tx`. One idea to simplify this would be to have this function return a `Result<()>` that can then be sent via the `tx` and it doesn't have to be passed in here. Just call the function above, await the result and send it away. This way one response is guaranteed.

src/connectors/source.rs
```diff
@@ -834,21 +840,24 @@ where
...
         };
         // We can not move this to the system flow since we need to know about transactionality
         for (pipeline_url, p) in &pipelines {
             self.ctx.swallow_err(
                 p.send_mgmt(pipeline::MgmtMsg::ConnectInput {
                     endpoint: DeployEndpoint::new(&self.ctx.alias, &port, pipeline_url.meta()),
+                    port: Cow::from(pipeline_url.port().to_string()),
+                    tx: tx.clone(),
```
mfelsche, line 860: Each `ConnectInput` will send a response to this channel, which expects only 1 message when the whole outer operation is done. This will lead to the link reported as ready when in fact it might not be ready yet.

I'd suggest to create a `bounded(1)` channel here and wait for as many messages as we have pipelines here.
If a send failed, we might want to simply return the `Err`.

src/pipeline.rs
```diff
@@ -572,45 +577,85 @@ pub(crate) async fn pipeline_task(
...
-                                ),
-                                target: InputTarget::Pipeline(Box::new(addr.clone())),
-                                is_transactional: true,
-                            })
-                            .await
-                        {
-                            error!("{ctx} Error connecting input pipeline {endpoint}: {e}",);
-                        }
-                    }
+                // if let OutputTarget::Pipeline(pipe) = &target {
```
mfelsche, line 623: why is this commented out?

mfelsche (commented):
I like where this is going.
But we need to take care that `result_tx` we pass through really only send 1 result back when the whole operation is done.

src/pipeline.rs
mfelsche: Then we can scratch this variant, it seems it doesnt do anything, does it?

Licenser: Ja the entire if block that's commented out can go 👍 we'll clean it up

Licenser (dismissed):

Licenser (approved):

PR #1869, opened 2022-08-01
DRAFT: Initial attempt at concurrent gRPC for GCL writer

darach:

# Pull request

Attempt to add bounded concurrent in flight requests to GCL writer which is the first gRPC / tonic
based connector to require it. As the `ConcurrencyCap` method does not work here as tonic clients
are not cloneable - we attempt to use a preallocated set of tasks. However, a preallocated set of
tasks may introduce blocking further downstream and complicates ( makes practically impossible )
unit testing 😬 

## Related

## Checklist

* [ ] The code is tested

### DIFF
--- a/src/connectors/impls/gcl/writer/sink.rs
+++ b/src/connectors/impls/gcl/writer/sink.rs
@@ -12,25 +12,222 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::time::Duration;
+
 use super::meta;
 use crate::connectors::google::AuthInterceptor;
 use crate::connectors::impls::gcl::writer::Config;
 use crate::connectors::prelude::*;
 use crate::connectors::utils::pb;
+use async_std::channel::bounded;
+use async_std::channel::Sender;
 use async_std::prelude::FutureExt;
+use async_std::task::JoinHandle;
 use googapis::google::logging::v2::log_entry::Payload;
 use googapis::google::logging::v2::logging_service_v2_client::LoggingServiceV2Client;
 use googapis::google::logging::v2::{LogEntry, WriteLogEntriesRequest};
 use gouth::Token;
 use prost_types::Timestamp;
-use std::time::Duration;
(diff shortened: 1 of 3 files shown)


src/connectors/impls/gcl/writer/sink.rs
```diff
@@ -12,25 +12,222 @@
...
 use prost_types::Timestamp;
-use std::time::Duration;
 use tonic::codegen::InterceptedService;
 use tonic::transport::{Certificate, Channel, ClientTlsConfig};
 use tonic::{Code, Status};
+use tremor_common::time::nanotime;
+
+#[allow(clippy::large_enum_variant)]
+pub(crate) enum GclEvent {
+    #[allow(dead_code)] // NOTE This is used by the tests and not intended for use in production
```
mfelsche, line 38: would a `#[cfg(test)]` work here? I think i did this in some other place, but i am not 100% sure.

src/connectors/impls/gcl/writer/sink.rs
```diff
@@ -12,25 +12,222 @@
...
+        let channel = Channel::from_static("https://logging.googleapis.com")
+            .connect_timeout(Duration::from_nanos(config.connect_timeout))
+            .tls_config(tls_config)?
+            .keep_alive_while_idle(true)
+            .tcp_keepalive(Some(Duration::from_secs(10)))
+            .http2_keep_alive_interval(Duration::from_secs(10))
+            .connect()
+            .await?;
+
+        let token_header_value = token.header_value();
```
mfelsche, line 98: this function needs to be called in the interceptor function, as this is the place where the token is renewed.

src/connectors/impls/gcl/writer/sink.rs
Licenser: Also probably good not to allow large enum variants and Box the large parts instead, `allow` is usually a sign of something being wrong :)

darach:
Superseded by #1924

PR #1883, opened 2022-08-08
Issue less coordinator requests to kafka group-coordinator

mfelsche:
# Pull request

## Description

Currently for each event we commit the offset directly to the group-coordinator, doing a request to the broker.
During pause/resume we are also doing 1 request for getting the current assignment.

This PR fixes this by caching the current assignment and invalidating it only upon a rebalance and making use of `enable.auto.commit` and the local offset store, thus sending commits to the group coordinator in a fixed interval.

It is still possible to commit directly if this behaviour is desired.

Also the configuration interface has changed in a breaking way, by introducing a `mode` to simplify the common cases and making it possible to use them without detailed `librdkafka` knowledge. We still allow custom configurations with a `custom` mode.

This is a successor to #1837 

## Related

This also contains a small fix to elastic that will be ported to 0.12 branch as well.

## Checklist

* [x] The code is tested

### DIFF
--- a/src/connectors/impls/kafka/consumer.rs
+++ b/src/connectors/impls/kafka/consumer.rs
@@ -144,8 +144,10 @@ impl Mode {
                 let interval_ms = format!("{}", Duration::from_nanos(*commit_interval).as_millis());
                 client_config
                     .set("enable.auto.commit", "true")
-                    .set("enable.auto.offset.store", "false")
-                    .set("auto.commit.interval.ms", interval_ms.as_str());
+                    .set("enable.auto.offset.store", "false");
+                if *commit_interval > 0 {
+                    client_config.set("auto.commit.interval.ms", interval_ms.as_str());
+                }
             }
             Mode::Custom {
                 rdkafka_options, ..
@@ -233,8 +235,7 @@ impl ConnectorBuilder for Builder {
             path: vec![],
         };
 
-        let tid = task::current().id();
-        let client_id = format!("tremor-{}-{}-{:?}", hostname(), alias, tid);
+        let client_id = format!("tremor-{}-{}", hostname(), alias);
         let mut client_config = config.mode.to_config();
 
         // we do overwrite the rdkafka options to ensure a sane config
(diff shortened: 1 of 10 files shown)


src/connectors/impls/kafka/consumer.rs
```diff
@@ -28,43 +30,184 @@ use indexmap::IndexMap;
...
+    ///         "commit_interval": nanos::from_seconds(5)    
+    ///     }
+    ///   ```
+    ///
+    /// - `custom`: Configure the connector yourself as you like, by providing your own set of `rdkafka_options`. You should know what you are doing when using this.
+    ///
+    ///   Example: ```
+    ///   "custom": {
+    ///     "rdkafka_options": {
+    ///         "enable.auto.commit": false
```
Licenser, line 197:
```suggestion
    ///         "enable.auto.commit": "false"
```

src/connectors/impls/kafka/consumer.rs
```diff
@@ -28,43 +30,184 @@ use indexmap::IndexMap;
...
+        /// if `enable.auto.commit` is set to `false`:
+        ///
+        /// if set to `true` this source will reset the consumer offset to a
+        /// failed message, so it will effectively retry those messages.
+        ///
+        /// If set to `false` this source will only commit the consumer offset
+        /// if the message has been successfully acknowledged.
+        ///
+        /// This might lead to events being sent multiple times.
+        /// This should not be used when persistent errors are expected (e.g. if the message content is malformed and will lead to repeated errors)
```
Licenser, line 72:
```suggestion
        /// This config determines the behaviour of this source.
        ///
        /// If `enable.auto.commit` is set to `false`:
        ///
        /// If `retry_failed_events` is set to `true` this source will reset the consumer offset to a
        /// failed message, so it will effectively retry those messages.
        /// DANGER: this might lead to a lot of traffic to the group-coordinator!
        ///
        /// If `retry_failed_events` is set to `false` this source will only commit the consumer offset
        /// if the message has been successfully acknowledged.
        ///
        /// This might lead to events being sent multiple times.
        /// This should not be used when persistent errors are expected (e.g. if the message content is malformed and will lead to repeated errors)
```

Licenser (approved):
🚀 really excited about this fix :D

PR #1894, opened 2022-08-09
Short circuit boolean expressions

### LINKED ISSUE #1794: Boolean expressions in pipeline select statements don't appear to be shortcircuiting
**Problem**

I was using a boolean `or` expression in a select statement, where the second part is only valid if the first part is not true.

I would expect such a boolean expression to short circuit, and not evaluate the second part if the first part is true.

**Steps**

I was using this select statement in a pipeline. But from the error, it seems that both the statements are getting evaluated.

```
      104 |   select event from in where (absent event.exit_now) or (event.exit_now != true) into out;
          |                                                                ^^^^^^^^ Trying to access a non existing event key `exit_now`
```

**Possible Solution(s)**

Match expressions can do this, but this would require me to write a script just to be able to evaluate this expression.

**Notes**

Output of `rustup --version`:
```
$ rustup --version
rustup 1.24.3 (2021-05-31)
```

Output of `rustup show`:
```
Default host: x86_64-apple-darwin
rustup home:  /Users/sb415l/.rustup

installed toolchains
--------------------

stable-x86_64-apple-darwin (default)
1.52.1-x86_64-apple-darwin
1.61.0-x86_64-apple-darwin

active toolchain
----------------

stable-x86_64-apple-darwin (default)
rustc 1.60.0 (7737e0b5c 2022-04-04)
```
Output of `tremor --version`: `tremor 0.12.4`

Agares:

# Pull request

## Related

* Related Issues: fixes #1794

## Checklist

* [x] The code is tested

### DIFF
--- a/tremor-script/src/interpreter/imut_expr.rs
+++ b/tremor-script/src/interpreter/imut_expr.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::ast::{BooleanBinExpr, BooleanBinOpKind};
+use crate::static_bool;
 use crate::{
     ast::{
         base_expr::Ranged, binary::extend_bytes_from_value, BaseExpr, BinExpr, Comprehension,
@@ -37,7 +38,6 @@ use std::{
     borrow::{Borrow, Cow},
     iter, mem,
 };
-use value_trait::StaticNode;
 
 fn owned_val<'val, T>(v: T) -> Cow<'val, Value<'val>>
 where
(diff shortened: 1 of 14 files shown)


tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -455,6 +460,79 @@ impl<'script> ImutExpr<'script> {
...
+        meta: &'run Value<'event>,
+        local: &'run LocalStack<'event>,
+        expr: &'run BooleanBinExpr<'event>,
+    ) -> Result<Cow<'run, Value<'event>>>
+    where
+        'script: 'event,
+    {
+        let lhs = stry!(expr.lhs.run(opts, env, event, state, meta, local));
+
+        match (expr.kind, lhs.as_ref()) {
```
Licenser, line 477: `as_bool` or `try_as_bool` will be better then matching against `Value::Static(StaticNode::Bool(lval))` the enum varriants are not stable part of the API

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -455,6 +460,79 @@ impl<'script> ImutExpr<'script> {
...
+                }
+
+                let rhs = stry!(expr.rhs.run(opts, env, event, state, meta, local));
+
+                if let Some(rval) = rhs.as_bool() {
+                    return if *lval || rval {
+                        Ok(Cow::Borrowed(&TRUE))
+                    } else {
+                        Ok(Cow::Borrowed(&FALSE))
+                    };
```
Licenser, line 490: we have a macro for this:

```suggestion
static_bool!(*lval || rval)
```

Licenser (approved):
nice 🚀

PR #1897, opened 2022-08-10
Fix invalid scope handling for flow definition default args

mfelsche:
# Pull request

## Description

Default arguments expressions have been wrongfully evaluated in the inner scope of the definition. This PR correctly evaluates them on the outer scope.

## Checklist

* [x] The code is tested

## Performance
Doesn't apply, this is only a compile time change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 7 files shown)


Licenser (approved):
:rocket: good catch!

PR #1899, opened 2022-08-10
Allow adding arrays with the plus operator

Agares:

# Pull request

## Related

## Checklist

* [ ] The code is tested


Licenser:
It would be nice to check if the RHS of the `+` is just created for addition and instead of of going over that doing a pre-allocation with a number of pushes.

```
event = event + ["const", function(), $something]
```

would right now create an array and then clone the values out of that for every invocation turning that into (not valid tremor code!)

```
event = {
 let new = event.clone().allocate(3)
 new.push("const");
 new.push(function())
 new.push(#something);
 new
```

would make that kind of construct reasonably faster

Licenser (commented):
That seems to do the trick 👍 but I think we can do a lot better. We know when we can do this optimisation during compile time, the interpreter doesn't need to do the decision. We add a intrinsic / interpreter function for `Append to Array` that replaces `BinOp(a, [ ... ])` with  `Append(a, ...)`.

That way we avoid a (or 3) branches during execution because we already know the RHS is a List, and already know we're doing an array add. And we'd avoid two of them for every binary operation.

We have a number of code re-writes like that usually implemented as walkers (thanks Mattias, he implemented them first and they work wonderfully <3).

tremor-script/src/interpreter/imut_expr.rs
```diff
@@ -257,6 +257,23 @@ impl<'script> ImutExpr<'script> {
             ImutExpr::Comprehension(ref expr) => {
                 self.comprehension(opts, env, event, state, meta, local, expr)
             }
+            ImutExpr::ArrayAppend { left, right, mid } => {
+                let left = stry!(left.run(opts, env, event, state, meta, local));
+
+                if let Some(left_val) = left.as_array() {
+                    let mut new_left = left_val.clone();
```
Licenser, line 264:
```suggestion
                 let mut new_left = left.into_owned().try_into_array().map_err(|e| todo!("Handle type mismatch!"))?;

```

tremor-script/src/ast/optimizer.rs
```diff
@@ -0,0 +1,80 @@
...
+
+            ConstFolder::new(self.helper).$walker_name(walkable)?;
+            ArrayAdditionOptimizer{}.$walker_name(walkable)?;
+
+            Ok(())
+        }
+    }
+);
+
+impl<'run, 'script, 'registry> Optimizer<'run, 'script, 'registry>
```
Licenser, line 37: I think a nicer approach would be something like:
```rust
 CombinedVisitor<First, Second>
  {
  first: First,
  second: Second
}

impl<First, Second> ImutExprWalker for CombinedVisitor
  where
  First: ImutExprVisitor,
  Second: ImutExprVisitor,
{}
impl<First, Second> ImutExprVisitor for CombinedVisitor
  where
  First: ImutExprVisitor,
  Second: ImutExprVisitor,
{
/// ... all the impls ... his is where a macro could go :P
    fn leave_expr(&mut self, e: &mut ImutExpr<'script>) -> Result<()> {
      first.leave_expr(e)?;
      second.leave_expr(e)?;
    }
/// ... all the impls ...
}
/// same for Deploy, Expr, Query


pub struct Optimizer<'run, 'script, 'registry>
where
    'script: 'run,
{
    visitor: CombinedVisitor<ConstFolder<'run, 'script, 'registry>, ArrayAdditionOptimizer>
}

 impl<'run, 'script, 'registry> Optimizer<'run, 'script, 'registry>{
    pub fn new(helper: &'run Helper<'script, 'registry>) -> Result<Self> {
        Ok(Self { visitor: CombinedVisitor::new(ConstFolder::new(self.helper).$walker_name(walkable)?, ArrayAdditionOptimizer{})}
    })
}

impl<'run, 'script, 'registry> AsRef<CombinedVisitor<ConstFolder<'run, 'script, 'registry>, ArrayAdditionOptimizer>
}> for Optimizer<'run, 'script, 'registry>{
fn as_ref(&self) -> &CombinedVisitor<ConstFolder<'run, 'script, 'registry>, ArrayAdditionOptimizer>
&self.visitor
}
}
}
```

Agares: How would that approach work when more optimizers get added? Adding more generic arguments seems very tedious.

Licenser: Absolutely, two are enough:) they can be chained ;). As the combiner also is a visitor it can combine itself. 

One advantage is that it only needs to walk once no matter how many visitors are present

Licenser (approved):
🚀

PR #1904, opened 2022-08-11
Rust 1.63.0

### LINKED ISSUE #1900: Set `rust-version` to MSRV in `Cargo.toml`

It seems the newest version of Tremor doesn't build with the Rust from nixpkgs, while attempting to sort out why, I noticed that `rust-version` was not set to the MSRV in `Cargo.toml`. That can be set according to the [docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)

I would like a `rust-version` set to the appropriate MSRV in `Cargo.toml` according to the [documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field).

Licenser:
# Pull request

## Description

Rust 1.63.0 update

Fixes #1900


Cargo.toml
```diff
@@ -10,6 +10,7 @@ name = "tremor-runtime"
 readme = "README.md"
 repository = "https://github.com/tremor-rs/tremor-runtime"
 version = "0.12.4"
+rust-version = "1.62"
```
mfelsche, line 13:
```suggestion
rust-version = "1.63"
```

Licenser: It's on purpose 1.63.

`rust-version` notates the minimal working version which is 1.62. It introduces the `Default` for `enum`, which we use in Kafka.

`rust-toolchain` defines the default toolchain for the project to use (and in our case compile production release too).

Hope that explains the difference?

Agares (approved):

PR #1912, opened 2022-08-16
Fix runtime swallowing acks and fails for some streams for batched events

mfelsche:
# Pull request

## Description

The runtime was swallowing all acks and fails but the ones from the highest stream. (Also all fails from the lowest stream). This is only a problem with batched events or events emitted from a window whose input events come from multiple streams.

This can lead to "interesting" effects e.g. when using the `kafka_consumer`, leaving quite big consumer lags for some partitions (not all).

With this PR, each all acks and fails from a batched event from all streams from the current source connector are forwarded and not swallowed anymore.

This PR also adapts the `EventId` a little bit, to make retrieval of those streams easier while more correct. There should be no notable difference in behaviour.

Last but not least, this PR (like a box of pralines) also contains a fix for `kafka_consumer` and `kafka_producer` metrics swallowing a lot of custom metrics due to a misuse of the `async_broadcast::Channel` API. The `stdio` and the `metrics` connector are possibly also affected and have been fixed as well.

In order to test the improved ack/fail behaviour, the `cb` connector has been changed to take `paths` instead of `path` now, in order to emit multiple streams from different files.
 
## Related

* https://github.com/tremor-rs/tremor-www/pull/235

## Checklist

* [x] The code is tested


tremor-pipeline/src/lib.rs
```diff
@@ -417,12 +418,20 @@ impl EventId {
...
-            other.stream_id,
-            other.pull_id,
-            other.pull_id,
-        );
+        // if self.source_id and self.stream_id are the same, calculate the max from the given pull_id and the one in other
+        // and insert it with binary search
+        // this makes searching the tracked id easier
+        // as we only need to look at the tracked_pull_ids then
+        let (min, max) =
+            if self.source_id() == other.source_id() && self.stream_id() == other.stream_id() {
```
Licenser, line 426: since this is happening a few times, would it make sense to have a `self.same_stream(other: &Self)` function?

tremor-pipeline/src/lib.rs
```diff
@@ -434,13 +443,14 @@ impl EventId {
                     unsafe { self.tracked_pull_ids.get_unchecked_mut(idx) }
                         .track_ids(other_tracked.min_pull_id, other_tracked.max_pull_id);
                 }
-                Err(idx) => self.tracked_pull_ids.insert(idx, other_tracked.clone()),
+                Err(idx) => self.tracked_pull_ids.insert(idx, *other_tracked),
             }
         }
     }
 
+    #[cfg(test)]
```
Licenser, line 451: <3 less `pub` less code to compile for normal builds!

Licenser (approved):
👍

tremor-pipeline/src/lib.rs
mfelsche: good point! :)

Licenser (dismissed):

Licenser (approved):

PR #1924, opened 2022-08-23
GCL - concurrent requests

Agares:

# Pull request

## Description

Adds support for in-flight concurrency / multiple simultaneous requests.

## Related

Not applicable.

## Checklist

* [x] The code is tested

## Performance

None at this time. Pending feedback production use cases who requested the feature.

### DIFF
--- a/src/connectors/google.rs
+++ b/src/connectors/google.rs
@@ -18,39 +18,8 @@ use tonic::metadata::MetadataValue;
 use tonic::service::Interceptor;
 use tonic::{Request, Status};
 
-pub trait TokenProvider: Clone {
+pub trait TokenProvider: Clone + Default + Send {
     fn get_token(&mut self) -> ::std::result::Result<Arc<String>, Status>;
-}
-
-#[cfg(not(test))]
-pub type DefaultTokenProvider = GouthTokenProvider;
-#[cfg(test)]
-pub type DefaultTokenProvider = TestTokenProvider;
-
-#[derive(Clone)]
-#[cfg(test)]
-pub struct TestTokenProvider {
-    token: Arc<String>,
-}
-
-#[cfg(test)]
-impl TestTokenProvider {
-    pub fn new() -> Self {
-        Self {
-            token: Arc::new(String::new()),
-        }
-    }
-
-    pub fn new_with_token(token: Arc<String>) -> Self {
-        Self { token }
-    }
-}
-
-#[cfg(test)]
-impl TokenProvider for TestTokenProvider {
-    fn get_token(&mut self) -> ::std::result::Result<Arc<String>, Status> {
-        Ok(self.token.clone())
-    }
 }
 
 pub struct GouthTokenProvider {
@@ -63,7 +32,12 @@ impl Clone for GouthTokenProvider {
     }
 }
 
-#[cfg(not(test))]
+impl Default for GouthTokenProvider {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl GouthTokenProvider {
     pub fn new() -> Self {
         GouthTokenProvider { gouth_token: None }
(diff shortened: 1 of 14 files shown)


src/connectors/google.rs
```diff
@@ -12,18 +12,89 @@
...
 
+use gouth::Token;
 use std::sync::Arc;
 use tonic::metadata::MetadataValue;
 use tonic::service::Interceptor;
 use tonic::{Request, Status};
 
-pub(crate) struct AuthInterceptor {
-    pub token: Box<dyn Fn() -> ::std::result::Result<Arc<String>, Status> + Send>,
+pub trait TokenProvider: Clone {
```
Licenser, line 21: let's avoid pubing things if we can :)
```suggestion
pub(crate) trait TokenProvider: Clone {
```

src/connectors/impls/gbq/writer/sink.rs
```diff
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::connectors::google::AuthInterceptor;
+#[cfg(not(test))]
```
Licenser, line 15: I think we can avoid all the cfg test in this file if we generalize `GbqSink` and `AuthInterceptor ` over `TokenProvider` and implement a default method, then during normal operations we can use `GbqSink<GouthTokenProvider>` and during tests `GbqSink<TestTokenProvider>`

src/connectors/impls/gcl/writer/sink.rs
```diff
@@ -13,24 +13,36 @@
 // limitations under the License.
 
 use super::meta;
-use crate::connectors::google::AuthInterceptor;
+#[cfg(not(test))]
```
Licenser, line 16: Same as for the gbq writer

src/connectors/impls/gpubsub/consumer.rs
```diff
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::connectors::google::AuthInterceptor;
+#[cfg(not(test))]
```
Licenser, line 15: same :D

src/connectors/impls/gpubsub/producer.rs
```diff
@@ -12,7 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::connectors::google::AuthInterceptor;
+#[cfg(not(test))]
```
Licenser, line 15: Same again ;)

Licenser (commented):
👍 not much to improve I would only to to get rid of the code that is `cfg`'ed for tests

src/connectors/impls/gbq/writer/sink.rs
Agares: I did that without generalizing, just by using a type alias, I think that's a simpler solution, let me know what you think :)

Licenser: Not really to be honest, there are still 10 compile-time conditionals in the code which makes the whole thing quite hard to read.

Agares: Hmm... I only see a few right now, almost all of them making the TestTokenProvider only compile in tests 🤔. Can you point at the specific problematic lines? Maybe I'm missing something

Licenser: Sure,

25, 27, 31, 36, 49, 66 in google.rs

all of those can be avoided, the Test porovider could move into `mod test` entirely.

I'm not a fan of the type alias either, as compile time switches are, when avoidable, always an addition in complexity that makes code harder to reason about, read, and maintain. We are not writing C where we don't have any other options.

Agares: donezo, unless I missed anything, this should be done now :)

Licenser (approved):
:rocket:!

Licenser (approved):
🚀

darach:
FYI. Had to fix DSO back to HEAD~15 for CI so changes are mostly @Agares ( on PTO! ) with bits of myself

darach (dismissed):
LGTM 👍

mfelsche (dismissed):

darach (approved):
LGTM. Flakiness in code coverage in patch mode 😬

mfelsche (approved):

mfelsche (approved):

mfelsche (approved):

darach (approved):
Go. Go. Gadget. Armpits! 😬

mfelsche (approved):

PR #1928, opened 2022-08-29
Fix server connectors to properly close sockets when client disconnects.

mfelsche:
# Pull request

## Description

Fix server connectors to properly close sockets when client disconnects.
Also fixing those server connectors when no pipeline is connected to the sink part.

## Checklist

* [x] The code is tested


src/connectors/impls/ws/server.rs
```diff
@@ -209,19 +213,24 @@ impl Connector for WsServer {
...
+                                        stream_id,
+                                        Some(connection_meta.clone()),
+                                        &ctx,
+                                        ws_writer,
+                                    )
+                                    .await;
+                                Some(sink_runtime.clone())
+                            } else {
+                                None
+                            };
```
Licenser, line 229: I notice this code in quite a few okaces, is there a chance to generalize it? They sem slightly different but somewhat the sasme.

mfelsche: it comes up in exactly  places, and the implementations are completely different each time. Don't know if it is worth the generics-juggling.

Licenser (approved):
🚀 nice!

PR #1971, opened 2022-09-20
Fix pipeline creation when pipelines has unused nodes

mfelsche:
# Pull request

## Description

An example pipeline like this would result in a pipeline graph without any edges:

```
define pipeline foo
pipeline
  create stream unused;
end;
```

The root cause is that somehow the node has been detected as unused twice and was removed twice. Due to petgraph (our DAG library) reusing the idx of a removed node as end-node, removing then ode twice did remove the final output node for some reason.
This PR takes care that each idx is only removed once by using a HashSet instead of a Vec for storing the nodes detected as unused.

## Checklist

* [x] The code is tested

## Performance

This doesn't impact pipeline DAG structure beyond fixing broken pipelines without edges.


darach (dismissed):
LGTM 👍

darach (approved):

PR #1972, opened 2022-09-20
Fix possible hangs for multiple connectors when no pipeline is connected to their out port

mfelsche:
# Pull request

## Description

When no source was consuming from these connectors the internal queue grew full and the sender on that queue was hanging infinitely. Now we only send to that queue if we know that there is a queue consuming from it.

Affected connectors: http_client, kv, dns.

## Checklist

* [x] The code is tested

## Performance

We might get a little performance drop because now we check an atomic bool on each event.


src/connectors/impls/cb.rs
```diff
@@ -273,58 +273,63 @@ impl CbSource {
...
-        let idx: usize = self.num_sent % self.files.len();
-        // ALLOW: the modulo makes this safe
-        if let Some(file) = self.files.get_mut(idx) {
-            if let Some(line) = file.next().await {
+    async fn pull_data(&mut self, pull_id: &mut u64, ctx: &SourceContext) -> Result<SourceReply> {
+        if !self.files.is_empty() {
+            let idx: usize = self.num_sent % self.files.len();
+            let file = self
+                .files
+                .get_mut(idx)
```
darach, line 281: Perhaps add the modulo comment safety comment back here as it still applies?

darach (approved):
LGTM 👍

PR #2003, opened 2022-10-12
Select window timeout

mfelsche:
# Pull request

## Description

Fix windows based on `interval` that time out (interval elapsed) not sending the correct event id.
In effect the events that are accumulated in the window are not tracked and cannot be acked or failed.
This might show up as consumer lag in kafka if `kafka_consumer` is used as an upstream connector.

## Checklist

* [x] The code is tested

## Performance
No significant diff between main and this branch on the benchmarks we have.

### DIFF
--- a/tremor-pipeline/src/op/trickle/select.rs
+++ b/tremor-pipeline/src/op/trickle/select.rs
@@ -17,13 +17,11 @@
 #[cfg(test)]
 mod test;
 
-use std::mem;
-
 use super::window::{self, Group, Window};
 use crate::op::prelude::trickle::window::{GroupWindow, SelectCtx, Trait};
+use crate::op::prelude::*;
 use crate::{errors::Result, SignalKind};
-use crate::{op::prelude::*, EventIdGenerator};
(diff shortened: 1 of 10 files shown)


tremor-pipeline/src/op/trickle/select.rs
```diff
@@ -61,12 +58,11 @@ impl Select {
                 window_impl,
             })
             .collect();
-        let event_id_gen = EventIdGenerator::for_operator(operator_uid);
         let SelectStmt { aggregates, .. } = select;
-        let windows_itr = windows.iter();
+        let windows_itr = (0_u64..).zip(windows.iter());
```
Licenser, line 62: would `.enumerate()` be better here?

Licenser (commented):
👍

Licenser (approved):

tremor-pipeline/src/op/trickle/select.rs
Licenser: ah sneaky!

PR #2008, opened 2022-10-17
Sink traits prototype

Agares:

# Pull request

## Related

## Checklist

* [ ] The code is tested


Cargo.toml
```diff
@@ -22,6 +22,7 @@ members = [
   "tremor-pipeline",
   "tremor-script",
   "tremor-value",
+  "tremor-runtime-derive"
```
Licenser, line 25: Perhaps we can pick a better name for this? it's less about the runtimes than the connector traits.

src/connectors/sink.rs
```diff
@@ -751,16 +762,19 @@ where
...
-                                .sink
-                                .on_event(
-                                    port.borrow(),
-                                    event,
-                                    &self.ctx,
-                                    &mut self.serializer,
-                                    start,
-                                )
-                                .await;
+                            let res = if port == "control" {
```
Licenser, line 765: would be nice to `const` this out like `IN`, `OUT`, `ERR` so if we ever change it we can do it easily :) also saves me from typos

src/connectors/traits.rs
```diff
@@ -0,0 +1,33 @@
+use crate::connectors::sink::Sink;
+use crate::errors::Result;
+
```
Licenser, line 3: Would be nice to have a `prelude` module that exports all the traits so they can be imported via `*`

Licenser (commented):
Looks good so far, would be nice to have `file` and perhaps `gcs` implement the FileIO trait to see how it looks for connectors that implement a given trait

Cargo.toml
Agares: `tremor-runtime-connector-traits-derive`? Generally, the convention seems to be to keep proc macros in a crate with the same name as the main crate + `-derive`, but I agree this is a bit unspecific in this case.

Licenser: How about `tremor-connector-derive`. The convention makes sense if the derive is for what is in the trait (serde-derive, simd-json-derive etc), but that's not the case for us it's explicitly about connectors in our case, and down the road when the PDK comes in they'll probably be split out in an own crate.

src/connectors/impls/oneshot.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+pub(crate) struct OneShotSource {
+    data: Option<Value<'static>>,
+}
+
+#[async_trait::async_trait]
+impl Source for OneShotSource {
+    async fn pull_data(&mut self, _pull_id: &mut u64, _ctx: &SourceContext) -> Result<SourceReply> {
+        if let Some(data) = self.data.take() {
```
Licenser, line 82: I love the use of `take` here 👍 really nice

.github/checks/safety.sh
```diff
@@ -138,6 +138,10 @@ while getopts hamuiprebldxcft opt; do
         p)
             for file in $files
             do
+                if [[ $file == './tremor-connector-derive/'* ]]; then
```
Licenser, line 141: we have an .checkignore file that could house this instead of adjusting the script :)

Agares: AFAIK `.checkignore` stops all the checks from happening, but I think for the derives, all the others make sense, it just needs to be able to panic.

Licenser: That makes sense, would a `// ALLOW: this is a derive macro so panicking is the way to error` in the derive be an alternative? Hardcoding the file path in the check script seems wrong to me o.O especially when we have alternative ways to deal with it already.

src/connectors/impls/file.rs
```diff
@@ -41,45 +45,24 @@ pub(crate) enum Mode {
...
 pub(crate) struct Config {
-    /// path to the file
-    pub(crate) path: PathBuf,
-    /// how to interface with the file
-    pub(crate) mode: Mode, // whether we read or write (in various forms)
     /// chunk_size to read from the file
     #[serde(default = "default_buf_size")]
     pub(crate) chunk_size: usize,
+
+    pub(crate) initial_command: Option<crate::connectors::traits::Command>,
```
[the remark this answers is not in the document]
Agares: I don't know, standarisation is definitely good, OTOH not all connectors will be command driven. It would be nice to be able to make that an option for all the command driven connectors, and only those, but I don't think that's easily doable with the current abstractions that we have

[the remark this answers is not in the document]
Licenser: Verify means that also other errors could be caught at "compile time" which would be really nice to be hoenst. Things like malformated commands, or commands that are not supported

src/config.rs
```diff
@@ -149,6 +150,8 @@ pub(crate) struct Connector {
 
     //pub(crate) on_pause: PauseBehaviour,
     pub(crate) metrics_interval_s: Option<u64>,
+
+    pub(crate) initial_command: Option<Command>,
```
Licenser, line 154: could we make this `commands: Vec<Command>` instead of an option. This would help in, for example a UDP socket where you want to open and set a few flags, or a file connector that wants to open multiple files etc.

src/connectors.rs
```diff
@@ -1100,6 +1101,14 @@ pub(crate) trait Connector: Send {
 
     /// Returns the codec requirements for the connector
     fn codec_requirements(&self) -> CodecReq;
+
+    fn validate(&self, config: &ConnectorConfig) -> Result<()> {
```
Licenser, line 1105: instead of a "supports" / "not supports" could we use the same tactic we use for dispatching a command and instead of executing it returning valid / invalid? So parse the command into a struct, check if the struct is valid, check if the connector in question support this type of command.

Not sure on the implementation cost of this, but it's worth a thought as it'd be more helpful then a global supported / not supported.

So basically a hygenic error that validates the command during compile time to the limits that are possible (as in not checking of the file is readable, but checking if the connector supports file IO commands and possibly letting it check if the connector supports this specific command.)

src/connectors/impls/file.rs
```diff
@@ -250,15 +260,19 @@ struct FileSink {
 }
 
 impl FileSink {
-    async fn new(config: &Config, files_to_read: Sender<FileDescriptor>) -> Result<Self> {
+    async fn new(
+        config: &Config,
+        initial_command: &Option<Command>,
```
Licenser, line 265: I think it would be nicer to pull this out into the wrapping connector abstraction itself, that way we don't need to implement it for each source/sink repeatedly. Instead the wrapper could send the initial commands as messages to control upon startup making the whole initial commands logic opaque to the connector implementer.

src/connectors.rs
Agares: I think we can do it with the current setup - the command can be matched upon in the validate method, and then the method can decide if it passes

Agares: (it is passed as an unserialized enum to ConnectorConfig)

tremor-cli/tests/integration/udp-postprocessors/config.troy
```diff
@@ -9,9 +9,8 @@ flow
   with
     codec = "json",
     preprocessors = ["separate"],
+    initial_commands = [{"FileIo": {"Open": {"path": "in.json", "handle": "input", "mode": "read"}}}],
```
Licenser, line 12: some tremor script helper functions would be super nice for those :)

tremor-cli/tests/integration/tcp-reuseport/before/config.troy
```diff
@@ -10,11 +10,13 @@ flow
...
     ],
     config = {
-      "url": "127.0.0.1:22224",
       "socket_options": {
         "SO_REUSEPORT": true
       }
-    }
+    },
+    initial_commands = [
+      {"SocketServer": {"Listen": {"address": "tcp://127.0.0.1:22224", "handle": "a"}}}
```
Licenser, line 18: leaving this in here for later so we don't forget: we should follow the same convention for naming as every where else `SocketServer` => `socket_server`

tremor-cli/tests/integration/tcp-reuseport/config.troy
```diff
@@ -10,7 +10,7 @@ flow
       "separate"
     ],
     config = {
-      "url": "127.0.0.1:22224",
+      "default_handle": "a",
```
Licenser, line 13: anotehr not for later, we should see if we can avoid a default handle config it seems to make things complicated and error prone. I'm not sure what a better approach is would it's something to think about as this seems wrong. Perhaps the "first": handle can be the default?

tremor-cli/tests/integration/tcp/config.troy
```diff
@@ -23,7 +25,21 @@ flow
   create connector stdio from connectors::console;
   create connector tcp_server;
 
-  create pipeline server_side from pipelines::passthrough;
+  define pipeline server_side
+  pipeline
+    define script scr
```
Licenser, line 30: note for later:

having scripts top set the different connector metadata in a library would be nice in addition to the functions

i.e.

```
use tremor::traints::socket_server;

create script socket_server::set_handle with handle = "output" end;
```

PR #2017, opened 2022-10-20
script enhancements

Licenser:
# Pull request

## Description

This pull request implements script enhancements from the [script operator enhancements RFC](https://www.tremor.rs/rfc/accepted/script-operator-enhacements).

A derivation from he original RFC is that `script from <port>` takes an ident for port instead of a string, as in when in both the pipeline as well as in the deploy level we use idents.

It introduced the `script from <port>`  and the `state` section,

## Related

* [RFC](https://www.tremor.rs/rfc/accepted/script-operator-enhacements)
* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/260)

## Checklist

* [x] The code is tested

## Performance

-


tremor-script/src/script.rs
```diff
@@ -58,6 +59,8 @@ pub enum Return<'event> {
 pub struct Script {
     /// Rental for the runnable script
     pub script: crate::ast::Script<'static>,
+    /// Rental for the runnable script
+    pub named: HashMap<String, crate::ast::Script<'static>>,
```
mfelsche, line 63: could we consolidate all the scripts into this map? the `script` above is also just listening on the `in` port, isn't it? And a `script from in` would overwrite this script, no?

tremor-script/src/ast/query/raw.rs
```diff
@@ -377,19 +379,27 @@ impl<'script> Upable<'script> for ScriptDefinitionRaw<'script> {
...
         helper.enter_scope();
-        let script = self.script.up_script(helper)?;
+        let mut script = self.script.up_script(helper)?;
+        script.state = state;
         let mid = self.mid.box_with_name(&self.id);
         helper.leave_scope()?;
         // Handle the params in the outside module
         let params = self.params;
         let params = params.up(helper)?;
+        let named = self
```
mfelsche, line 391: we need to error here, if the `in` port is part of the named scripts and we have, as it would never be executed in the current logic.

tremor-script/src/script.rs
Licenser: we coud but I don't think we should. the `in` port is the 90% case and avoiding a hashmap lookup is nice for that.

tremor-script/src/ast/query/raw.rs
Licenser: good call, also prably errors for re-defining the same from section 👍

tremor-script/src/script.rs
mfelsche: True dat

tremor-script/src/ast/query/raw.rs
Licenser: fixed

tests/query_errors/script_in_port/error.txt
```diff
@@ -0,0 +1,8 @@
+Error: 
+    1 | define script in_port
+    2 | script from in
+      |             ^^ script port `in` is reserved
```
mfelsche, line 4: this err message could contain a hint to use a "plain" `script` instead.

tests/query_errors/state_not_static/error.txt
```diff
@@ -0,0 +1,8 @@
+Error: 
+    3 | define script bad_state
+    4 | state
+    5 |   system::nanotime()
```
mfelsche, line 4: i find it sad, that we cannot use the timestamp of initialization for the initial state. It rules out so many cool possible time-based algos we could pull with scripts. With this we need workarounds to express this. :(

tremor-script/src/script.rs
```diff
@@ -58,6 +59,8 @@ pub enum Return<'event> {
 pub struct Script {
     /// Rental for the runnable script
     pub script: crate::ast::Script<'static>,
+    /// Rental for the runnable script
```
mfelsche, line 62:
```suggestion
    /// A map from input port to runnable script
```

tests/query_errors/state_not_static/error.txt
Licenser: I don't think there is, and I don't think there should be a workaround. the timestamp it initializes with becomes pointless as there is no time guarantee as to when the compilation of the code happens it'd be a arbritary.

tremor-script/src/script.rs
```diff
@@ -58,6 +59,8 @@ pub enum Return<'event> {
 pub struct Script {
     /// Rental for the runnable script
```
mfelsche, line 60:
```suggestion
    /// Script to be executed upon events entering via `in` port
```

mfelsche (commented):
another test verifying the behaviour upon sending an event to an undefined port would be nice.

and one with a script with multiple defined ports, verifying that the routing works well.

tests/query_errors/state_not_static/error.txt
mfelsche: So, should we maybe defer evaluation of the state expression to the moment when the first event arrives and the state is not yet initialized?

Licenser:
> another test verifying the behaviour upon sending an event to an undefined port would be nice.
👍 

> and one with a script with multiple defined ports, verifying that the routing works well.
`script_ports` verifies does that

tests/query_errors/state_not_static/error.txt
Licenser: Then we should do the same for args and with to keep it consistent; it means our pipelines are no longer deterministic and replay testing isn't possible any more. With clustering now the same operator will initialize differently based on the system it's one.

I still think that's a bad place to be in.

mfelsche: > With clustering now the same operator will initialize differently based on the system it's one.

I honestly don't see a problem with this.

mfelsche: But it is fine, lets keep it a compile time constant for now and see if the need arises to wire this up differently.

Licenser: Ja it's always easier to make things more relaxed then harder, I just looked into how to do a init  on first event and it's a code wise PITA too, our `run` function is structured to change the event as a return value so we'd either need to rewrite that whole logic or clone the event (probably cloning wouldn't be soooo terrible for a init but not pretty either)

PR #2033, opened 2022-11-01
Dogstatsd perf

Licenser:
# Pull request

## Description

Improve dogstats performance

## Checklist

* [x] The code is tested

## Performance

10-20% perf improvement on parsing/rendering dogstatsd data.


mfelsche (approved):
Great stuff!
I think we could even get some more performance improvements if we use the `memchr` crate (already in dependencies) instead of `str::split_once`. We could even iterate over the `:` and `|` splits.

Also the same improvements can be applied to the `statsd` codec. As dogstatsd seems to be just an addition to plain `statsd` we could also merge both into one.

Licenser:
Uh that's a good idea, I should give that a try. I admit I was tempted to throw in the separator detection from simd-json, but I'm not sure it's worth the effort.

Licenser:
I am not sure how backwards compatible dogstatsd is, I think they've slightly different types and structures but on the other hand having only a single codec would be nicer

darach (approved):
LGTM 👍

mfelsche (approved):

PR #2039, opened 2022-11-03
Improve performance of array::flatten

mfelsche:
# Pull request

## Description

Improve performance of `array::flatten` by using iterators for recursing. This gives a performance bump of roughly 30% (depending on the array length).

## Checklist

* [x] The code is tested

## Performance
Running the benchmark against main and this branch showed a rough 30% performance increase on my machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ## New Features
+
 * implement script enhancement RFC (initial state and port based scripts)
 * Improve dogstatsd performance
 
--- a/tremor-value/src/value/iter.rs
+++ b/tremor-value/src/value/iter.rs
@@ -1,0 +1,194 @@
+// Copyright 2022, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ## New Features
+
```
Licenser, line 6: can we have a changelog entry for the improvement? :)

tremor-value/src/value/iter.rs
```diff
@@ -0,0 +1,191 @@
+// Copyright 2022, The Tremor Team
```
Licenser, line 1: Could we move that to a contribution to simd-json's value_trait it seems generally useful

mfelsche: Do you think the logic employed for `IntoIterator` and the `.array_iter()` function make sense as they are?

CHANGELOG.md
mfelsche: It doesn't really fit into any of the existing categories... should we open a new one for: `Improvements` or `Performance`?

Licenser: I just threw the statsd one into the new features but yes it's not perfect, Improvements, works Performance works too, Fixes would work (if we consider low performance a bug but it really isn't fitting perfect either), new feature works (if we consider fast to be a feature, but not really either :P:)

tremor-value/src/value/iter.rs
Licenser: `array_iter` is 👍 into iter from the general idea too, I'm just a bit thinking if the trait makes sense as it would equally make sense to have `IntoIter` for a object and not an array but I'm not sure how to achieve that nicely to be honest

mfelsche: I am also having a hard time coming up with a good iterator interface. I think I am going to ditch the whole `iter` stuff for now. array::flatten also works without this.

Licenser (approved):

PR #2065, opened 2022-11-10
implement new `std::datetime` library

mfelsche:
# Pull request

## Description

Implement the new `std::datetime` library.

And add `truncate_*` functions to `std::time::nanos`.

## Related

* Related [TBD](https://github.com/tremor-rs/tremor-www-docs/pull/000)

## Checklist

* [x] The code is tested

## Performance

Doesn't really apply as those changes are net new. (A benchmark for some function would still be nice though.)

### DIFF
--- a/tremor-script/Cargo.toml
+++ b/tremor-script/Cargo.toml
@@ -26,6 +26,7 @@ base64 = "0.13"
 beef = { version = "0.5", features = ["impl_serde"] }
 byteorder = "1.4"
 chrono = "0.4"
+chrono-tz = "0.8"
 cidr-utils = "0.5"
 codespan = "0.11"
 dissect = "0.5"
@@ -64,6 +65,7 @@ xz2 = "0.1"
 
 [build-dependencies]
 lalrpop = "0.19"
+chrono-tz = "0.8"
 
 [dev-dependencies]
 criterion = "0.4"
@@ -72,6 +74,7 @@ matches = "0.1"
 pretty_assertions = "1.3.0"
 proptest = "1.0"
 tempfile = "3"
+test-case = "2.2"
 
 [features]
 # this is required for the erlang EQC tests of the language
--- a/tremor-script/lib/std/datetime.tremor
+++ b/tremor-script/lib/std/datetime.tremor
@@ -1,0 +1,172 @@
+### Support for parsing date strings to nanosecond timestamps and formatting nanosecond timestamps into strings and more.
+### 
+### The submodule [timezones](datetime/timezones.md) contains constants for each existing timezone in the IANA timezone database.
+### The submodule [formats](datetime/formats.md) contains constants commonly used formats.
+###
+### In this module we introduce a type called `datetime`.
+###
+### It is either:
+###  * a simple integer timestamp which is interpreted as *UTC nanosecond precision timestamp since epoch* or
+###  * a record with the following fields:
+###    * `timestamp`: a *UTC nanosecond precision timestamp since epoch*
+###    * `tz`: a timezone identifier from the [`std::datetime::timezones`](datetime/timezones.md) module or a string with a timezone name or a fixed offset
+###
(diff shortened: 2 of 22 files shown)


darach (approved):
Love it ❤️ ( once we obey clippy, etc.. )

tremor-script/Cargo.toml
```diff
@@ -64,6 +65,7 @@ xz2 = "0.1"
 
 [build-dependencies]
 lalrpop = "0.19"
+chrono-tz = "0.8"
```
Licenser, line 68: do we need this twice as it's already in the deps?

tremor-script/lib/std/datetime.tremor
```diff
@@ -0,0 +1,172 @@
...
+use std::datetime::timezones;
+use std::datetime::formats;
+
+## Formats the given datetime according to [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339#section-5.6)
+##
+## Returns a string.
+##
+## > ```tremor
+## > use std::datetime;
+## > datetime::rfc3339(1667997961)
```
Licenser, line 25: this is missing a
```
## > ```
```

tremor-script/lib/std/datetime.tremor
```diff
@@ -0,0 +1,172 @@
...
+## > datetime::day_of_month(0) == 1
+## > ```
+intrinsic fn day_of_month(datetime) as datetime::day_of_month;
+## Extract the day of the year starting from 1
+##
+## > ```tremor
+## > use std::datetime;
+## > datetime::day_of_year(0) == 1
+## > ```
+intrinsic fn day_of_year(datetime) as datetime::day_of_year;
```
Licenser, line 106: should we add `week_of_year` too?

Licenser (commented):
niiiice!

tremor-script/Cargo.toml
mfelsche: yeah, we need it at build time. Normal dependencies aren't available at build time (in build.rs).

tremor-script/lib/std/datetime.tremor
mfelsche: this is what `iso_week` does

Licenser: ah darn I missed that

tremor-script/Cargo.toml
Licenser: oh darn, thanks I didn't knew that :)

tremor-script/src/ast/helper.rs
```diff
@@ -61,7 +61,7 @@ impl Warning {
 #[derive(Default, Debug, Clone, Serialize, PartialEq)]
 pub struct Scope<'script> {
     /// Module of the scope
-    pub(crate) modules: std::collections::HashMap<String, module::Index>,
+    pub(crate) modules: std::collections::BTreeMap<String, module::Index>,
```
Licenser, line 64: No objection to the BTreeMap, just curious for a learning perspective why choose the btree map over the hashmap?

Licenser (dismissed):
Yaaaaas!

tremor-script/src/ast/helper.rs
mfelsche: The reason for this is that iterating over it will do it in alphabetical order. This leads to alphabetically ordered docs.

mfelsche: I would prefer insertion order, but this was the quickest i did come up with.

Licenser: Nice insight <3!

Licenser: > I would prefer insertion order, but this was the quickest i did come up with.

could this work?
```rust
Vec<(String, module::Index)>
```

Licenser (dismissed):

Licenser (approved):

PR #2070, opened 2022-11-14
Adding the chunk postprocessor

mfelsche:
# Pull request

## Related

* [docs PR](https://github.com/tremor-rs/tremor-www/pull/265)

## Checklist

* [x] The code is tested

### DIFF
--- a/src/postprocessor.rs
+++ b/src/postprocessor.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+mod chunk;
 mod compress;
 mod gelf;
 pub(crate) mod separate;
@@ -59,6 +60,7 @@ pub trait Postprocessor: Send + Sync {
 
 pub fn lookup_with_config(config: &PostprocessorConfig) -> Result<Box<dyn Postprocessor>> {
     match config.name.as_str() {
+        "chunk" | "chonk" => Ok(Box::new(chunk::Chunk::from_config(config.config.as_ref())?)),
         "compress" => Ok(Box::new(Compress::from_config(config.config.as_ref())?)),
         "separate" => Ok(Box::new(separate::Separate::from_config(&config.config)?)),
         "base64" => Ok(Box::new(Base64::default())),
--- a/src/postprocessor.rs
+++ b/src/postprocessor.rs
@@ -60,7 +60,7 @@ pub trait Postprocessor: Send + Sync {
 
 pub fn lookup_with_config(config: &PostprocessorConfig) -> Result<Box<dyn Postprocessor>> {
     match config.name.as_str() {
-        "chunk" | "chonk" => Ok(Box::new(chunk::Chunk::from_config(config.config.as_ref())?)),
+        "chunk" => Ok(Box::new(chunk::Chunk::from_config(config.config.as_ref())?)),
         "compress" => Ok(Box::new(Compress::from_config(config.config.as_ref())?)),
         "separate" => Ok(Box::new(separate::Separate::from_config(&config.config)?)),
         "base64" => Ok(Box::new(Base64::default())),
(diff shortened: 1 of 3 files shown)


darach (approved):
Chonky 👍

src/postprocessor.rs
```diff
@@ -59,6 +60,7 @@ pub trait Postprocessor: Send + Sync {
 
 pub fn lookup_with_config(config: &PostprocessorConfig) -> Result<Box<dyn Postprocessor>> {
     match config.name.as_str() {
+        "chunk" | "chonk" => Ok(Box::new(chunk::Chunk::from_config(config.config.as_ref())?)),
```
Agares, line 63: As much as I love the `chonk` name, I think having two might lead to unnecessary confusion and less grepability, so I'd stick to just `chunk`

Agares (approved):

Agares (approved):

src/postprocessor/chunk.rs
```diff
@@ -0,0 +1,192 @@
...
+                let mut output = Vec::with_capacity(self.max_bytes);
+                // append the new data to the buffer we do keep
+                output.extend_from_slice(data);
+                std::mem::swap(&mut output, &mut self.chunk);
+                Ok(vec![output])
+            }
+        } else {
+            self.chunk.extend_from_slice(data);
+            Ok(vec![])
+        }
```
Licenser, line 53: This would make the error condition a bit clearer and de-nest the expression
```suggestion
        if data.len() > self.max_bytes {
            self.warn(data.len());
            Ok(vec![])
        } else if (self.chunk.len() + data.len()) > self.max_bytes {
            let mut output = Vec::with_capacity(self.max_bytes);
            // append the new data to the buffer we do keep
            output.extend_from_slice(data);
            std::mem::swap(&mut output, &mut self.chunk);
            Ok(vec![output])
        } else {
            self.chunk.extend_from_slice(data);
            Ok(vec![])
        }
```

src/postprocessor/chunk.rs
```diff
@@ -0,0 +1,192 @@
...
+                if data.len() > self.max_bytes {
+                    self.warn(data.len());
+                    vec![output]
+                } else {
+                    vec![output, data.to_vec()]
+                }
+            } else {
+                output.extend_from_slice(data);
+                vec![output]
+            }
```
Licenser, line 71:
```suggestion
            if data.len() > self.max_bytes {
                self.warn(data.len());
                vec![output]
            } else if data.len() + output.len() > self.max_bytes {
                vec![output, data.to_vec()]
            } else {
                output.extend_from_slice(data);
                vec![output]
            }
```

Licenser (dismissed):
🚀

src/postprocessor/chunk.rs
```diff
@@ -0,0 +1,188 @@
...
+                vec![output]
+            } else if data.len() + output.len() > self.max_bytes {
+                vec![output, data.to_vec()]
+            } else {
+                output.extend_from_slice(data);
+                vec![output]
+            }
+        } else {
+            vec![output]
+        };
```
Licenser, line 70: Sorry I missed this before. I think there is an edge case when `output` is empty it could create a "fake" event - stream part.

I'm not sure if this is an issue with all connectors, but I could see some sending an empty message that they shouldn't.

```suggestion
        } else if output.is_empty() {
            vec![]
        } else {
            vec![output]
        };
```

mfelsche: Good catch!

mfelsche: I wonder why the proptest didnt catch it.

Licenser: I think because the minimal length of the tests is 1, 0 would probably have surfaced it. If we have no zero length messages we currently delay the sending of a "full" chunk to finalize (might be a 2nd thing?)

or in other words `data + current == max_buffer` it'd send in the next event not current perhaps that should/could be optimized?

Licenser (dismissed):
🚀 I love this :D

Licenser (approved):

PR #2075, opened 2022-11-15
Improved shrinking for expressions

ThomasArts:
# Pull request

## Description

This PR simplifies outcome of QuickCheck tests, i.e., when a test fails, with these enhancements the detected failure is smaller and therefore easier to understand.

## Related

No related issues

## Checklist

This code only affects test code

* [ *] The code is tested (ran 10k tests for both affected properties

## Performance

This change only affects testing


Licenser:
This is really helpful, thank you! Sadly before we can merge it you have to go through the DCO dance. It's a git feature about signing the commits, it can be done at the time with the -s flag or after the fact, the DCO bot that's complaining above has some easy and copy&pastable commands for that

PR #2077, opened 2022-11-16
Errors on overflows

Licenser:
# Pull request

## Description

Add checks for overflows in arithmetic and error accordingly instead of silently rolling over

## Checklist

* [x] The code is tested

## Performance

in the new arithmetics benchmark that effectively is a passthrough with a subtraction the overhead is within the margin of error:

`578.2k events/s` vs. `575.5k events/s`

or close numbers on repeated runs


tremor-script/src/interpreter.rs
```diff
@@ -248,8 +248,28 @@ where
     }
 }
 
+fn try_math<'run, 'event, T, O, I>(
```
mfelsche, line 251: would this make sense here (i.e. can this work given the generics?):
```suggestion
#[inline]
fn try_math<'run, 'event, T, O, I>(
```

mfelsche (dismissed):

tremor-script/src/interpreter.rs
Licenser: given there is no measurable impact of the change I'll trust rust on this one

darach (dismissed):
The `no newline at end of file` warnings are annoying otherwise LGTM 👍

tests/script_errors/sub_overflow/error.txt
```diff
@@ -0,0 +1,3 @@
+Error: 
+    2 | 0 - unsigned::max;
+      | ^^^^^^^^^^^^^^^^^ The binary operation `-` caused an overflow
```
Agares, line 3: should we call it an `underflow` in this case? I think it might make it cleaner in cases where the values are not immediately obvious

tremor-script/src/interpreter.rs
```diff
@@ -313,30 +337,42 @@ where
...
+                .and_then(|r| l.checked_shr(r))
+                .map_or_else(
+                    || error_invalid_bitshift(outer, inner),
+                    |n| Ok(Cow::Owned(Value::from(n))),
+                ),
+            // This is a bit a bit tricky, for unsigned shift right we need to turnj
+            // our left hand side into a u64 so that the shift right operation does no
+            // preserve signness then turn the result back inot a i64 to match the original
+            // type. We can't use try_from for the u64 <-> i64 conversions
             #[allow(
```
Agares, line 357: are the allows still needed here?

Agares (approved):

mfelsche (dismissed):

tests/script_errors/sub_overflow/error.txt
Licenser: I thought about it, we  totally could but that'd introduce a whole new error class and means separating out substractions and additions,not sure if the benefits justify that.

tremor-script/src/interpreter.rs
Licenser: yap since we have to do the i64 -> u64 -> i64 cast

tests/script_errors/sub_overflow/error.txt
Agares: yup, I think it's a good compromise between ease of implementation and ease of use

Agares (approved):

PR #2081, opened 2022-11-17
Quickcheck shrinking

ThomasArts:
# Pull request

## Description

This PR improves shrinking of scripts when a test case fails. It does so by actually evaluating part of the expression in which there are no variables. It also shrinks command. sequences much better than the previous model.

There are 3 unfinished parts, discussed on discord. These 3 are cleaerly marked in the `skip` function in `test_eqc.erl`. By disabling one of these options, faults will be detected and nicely shrunk. When the specification is settled, the model can be extended to handle those cases.
(negative testing is possible, but clutters the screen so much, that it is skipped as well)

## Related

Discussion on
- division by 0.0 gives infinite in Rust
- patch and merge behaviour w.r.t. undefined keys

## Checklist

* [x ] The code is tested by running 200k tests

## Performance

This PR only affects the tests, not the system

### DIFF
--- a/tremor-script/eqc/gen_script.erl
+++ b/tremor-script/eqc/gen_script.erl
@@ -17,7 +17,7 @@
 
 -include_lib("pbt.hrl").
 
--export([gen/1]).
+-export([pp/1]).
 
 gen_({'+', A, B}) ->
     ["(", gen_(A), " + ", gen_(B), ")"];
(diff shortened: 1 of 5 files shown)


Licenser (approved):
Thank you so much Thomas <3 this is great!

PR #2088, opened 2022-11-23
Revamp hanlding of http client

### LINKED ISSUE #2083: HTTP connector config is unclear about how codecs interact
**Problem**
I’m sure I’ve missed something, but after re-reading several times, I still can’t see it.
I’ve got the following connector
```
  define connector `http_in` from http_client
  with
    preprocessors = ["separate"],
    codec = "csv",  # Defaults to HTTP codec
    config = {
      "url": "https://www.cer-rec.gc.ca/open/imports-exports/crude-oil-exports-by-type-monthly.csv",
      "tls": true,
    }
  end;
```
I’ve verified that the file is actually csv, however whenever I run tremor, I get the following error 
```
"2022-11-20T23:12:20.600612414+00:00 ERROR tremor_runtime::connectors - [Sink::http_in] Error serializing event into request body: The value Object cannot be serialized to CSV. Expected an array.”
```

**Possible Solution(s)**
Here is the full file I am using, just in case
```
define flow canada_crude_exports
flow
  use tremor::connectors;
  use tremor::pipelines;

  define connector every_hour from metronome
  with
     config = {"interval": 1000}
  end;

  define connector `http_in` from http_client
  with
    preprocessors = ["separate"],
    codec = "csv",  # Defaults to HTTP codec
    config = {
      "url": "https://www.cer-rec.gc.ca/open/imports-exports/crude-oil-exports-by-type-monthly.csv",
      "tls": true,
    }
  end;

  define pipeline canada_crude_exports
  pipeline

    define script canada_crude_export
    # sample
    # 06/01/1985,1985,June,Light,28019.093333,176320.147616320000
    script
      use std::array;
      use std::type;
      use std::re;
      match event of
        case %["Period", _, _, _, _, _] => drop
        case %[start_date, _, _, type, m3_day, bbl_day] =>
          match type::is_null(state) of
            case true =>
              let state = [[start_date, type]]
            default =>
              match array::contains(state, [start_date, type]) of
                case true => drop
                default => array::push(state, [start_date, type])
              end
          end;
          let date_array = re::split("/", start_date);
          let day = date_array[0];
          let month = date_array[1];
          let year = date_array[2];
          emit {
            "from": "#{year}-#{month}-#{day}T00:00:00Z",
            "frequency": "monthly",
            "type": type,
            "volume": {
              "m3_day": m3_day,
              "bbl_day": bbl_day
            }
          }
        default => drop
      end;
    end;

    create script process from canada_crude_export;
  
    select event from in into process;
    select event from process into out;
  end;

  define connector influx_out from http_client
  with
    codec = "influx",
    # configure the endpoint we're writing to
    config = {
      "url": "http://127.0.0.1:8086/write?db=test",
      # We use a custom header to identify that we're tremor
      "headers": {"Client": ["Tremor"]}
    }
  end;

  create connector every_hour from every_hour;
  create connector http_in from http_in;
  create connector influx_out from influx_out;
  create pipeline passthrough from pipelines::passthrough;
  create pipeline process from canada_crude_exports;

  connect /connector/every_hour to /pipeline/passthrough;
  connect /pipeline/passthrough to /connector/http_in;
  connect /connector/http_in to /pipeline/process;
  connect /pipeline/process to /connector/influx_out;
  # connect /pipeline/process to /connector/surreal_out;
end;
```

**Notes**

Output of `rustup --version`:
Output of `rustup show`:
Output of `tremor --version`:

Licenser:
# Pull request

## Description

Update the HTTP connectir based on feedback in #2083 and the following discussion.

the HTTP connector no longer takes a codec. Instead, it is driven entirely by the content type and the `mime_mapping` config option. It introduces `*/*` as a mime mapping that acts as "default" and will be used when no other codecs can be found.

At this point, the server side isn't done yet.

## Related

* Related Issues: fixes #2083 
* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/272)

## Checklist

* [x] The code is tested

## Performance

-/-

### DIFF
--- a/src/connectors/impls/http/server.rs
+++ b/src/connectors/impls/http/server.rs
@@ -48,7 +48,8 @@ pub(crate) struct Config {
     tls: Option<TLSServerConfig>,
     /// custom codecs mapping from mime_type to custom codec name
     /// e.g. for handling `application/json` with the `binary` codec, if desired
-    custom_codecs: Option<HashMap<String, String>>,
+    /// the mime type of `*/*` serves as a default / fallback
+    mime_mapping: Option<HashMap<String, String>>,
 }
 
 impl ConfigImpl for Config {}
@@ -92,7 +93,7 @@ impl ConnectorBuilder for Builder {
             .as_ref()
             .map_or_else(|| HttpServer::DEFAULT_CODEC.to_string(), |c| c.name.clone());
         let inflight = Arc::default();
-        let codec_map = if let Some(custom_codecs) = config.custom_codecs.clone() {
+        let codec_map = if let Some(custom_codecs) = config.mime_mapping.clone() {
             MimeCodecMap::from_custom(custom_codecs)
         } else {
             MimeCodecMap::new()
--- a/src/connectors/impls/http/server.rs
+++ b/src/connectors/impls/http/server.rs
@@ -48,8 +48,7 @@ pub(crate) struct Config {
     tls: Option<TLSServerConfig>,
     /// custom codecs mapping from mime_type to custom codec name
     /// e.g. for handling `application/json` with the `binary` codec, if desired
-    #[serde(default)]
-    custom_codecs: HashMap<String, String>,
+    custom_codecs: Option<HashMap<String, String>>,
 }
 
 impl ConfigImpl for Config {}
@@ -93,7 +92,11 @@ impl ConnectorBuilder for Builder {
             .as_ref()
             .map_or_else(|| HttpServer::DEFAULT_CODEC.to_string(), |c| c.name.clone());
         let inflight = Arc::default();
-        let codec_map = MimeCodecMap::with_overwrites(&config.custom_codecs);
+        let codec_map = if let Some(custom_codecs) = config.custom_codecs.clone() {
+            MimeCodecMap::from_custom(custom_codecs)
+        } else {
+            MimeCodecMap::new()
+        };
 
         Ok(Box::new(HttpServer {
             config,
--- a/src/connectors/impls/http/server.rs
+++ b/src/connectors/impls/http/server.rs
@@ -12,31 +12,36 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::connectors::{
-    prelude::*,
-    utils::{mime::MimeCodecMap, tls::TLSServerConfig},
+use super::{
+    meta::{consolidate_mime, extract_request_meta, BodyData, HeaderValueValue},
+    utils::{FixedBodyReader, RequestId, StreamingBodyReader},
 };
-use crate::{connectors::spawn_task, errors::err_connector_def};
-use async_std::channel::unbounded;
+use crate::{
+    connectors::{
+        prelude::*,
+        spawn_task,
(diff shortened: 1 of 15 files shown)


src/connectors/impls/http/server.rs
```diff
@@ -48,8 +48,7 @@ pub(crate) struct Config {
     tls: Option<TLSServerConfig>,
     /// custom codecs mapping from mime_type to custom codec name
     /// e.g. for handling `application/json` with the `binary` codec, if desired
-    #[serde(default)]
-    custom_codecs: HashMap<String, String>,
+    custom_codecs: Option<HashMap<String, String>>,
```
mfelsche, line 51:
```suggestion
#[serde(default)]
    mime_mapping: Option<HashMap<String, String>>,
```

Licenser: That's not needed, `Option` can be omitted without the need of default :)

src/connectors/utils/mime.rs
```diff
@@ -63,17 +64,35 @@ impl MimeCodecMap {
...
+        let mut by_codec = CODEC_TO_MIME_TYPES
+            .into_iter()
+            .map(|(k, v)| (k.to_string(), v.to_string()))
+            .collect::<HashMap<String, String>>();
+
+        // After that we overwrite / extend the mapping with the definition given
+        // by the user
+        for (mime, codec_name) in &custom_codecs {
+            if mime != "*/*" {
+                by_codec.insert(mime.to_string(), codec_name.to_string());
```
mfelsche, line 79: this is inserted the wrong way (or into the wrong map).

Licenser: good catch thank you! That was me breaking things in refactoring :(

Licenser: fixed as 👍 
```rust
        for (mime, codec_name) in &custom_codecs {
            if mime != "*/*" {
                by_codec.insert(codec_name.clone(), mime.clone());
            }
        }
```
I'll push it with other feedback / the next round

src/connectors/impls/http/server.rs
mfelsche: Ok, but the renaming is needed to keep it consistent.

Licenser: ohhhh I miseed that part 🤦 thanks!

Licenser: fixed in the current uncommited [art

tremor-script/lib/tremor/mime.tremor
```diff
@@ -0,0 +1,38 @@
+### Mimetypes mappings to codecs
+###
+### Usage:
+###
+### ```tremor
+### use tremor::mime;
+### let codec = mime_to_codec["application/json"];
+### ```
+
```
mfelsche, line 9: may i ask about something like:

```
## constant used for matching any mime type agains a codec
const ANY = "*/*";
```

mfelsche (approved):
Otherwise :+1:

tremor-script/lib/tremor/mime.tremor
Licenser: ja!

PR #2096, opened 2022-11-25
Fix segfault by replacing c NIF with rust

Licenser:
# Pull request

## Description

Remove segfault issue in the c-nif by rewriting it in rust.

## Checklist

* [x] The code is tested

## Performance

nope


tremor-script-nif/build.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020-2022, The Tremor Team
```
mfelsche, line 1:
```suggestion
// Copyright 2022, The Tremor Team
```

Licenser: 👍 fixed locally

tremor-script/rebar.config
```diff
@@ -2,32 +2,35 @@
...
 
-{pre_hooks,
-  [{"(linux|darwin|solaris)", compile, "make -C c_src"},
-   {"(freebsd)", compile, "gmake -C c_src"}]}.
-{post_hooks,
-  [{"(linux|darwin|solaris)", clean, "make -C c_src clean"},
-   {"(freebsd)", clean, "gmake -C c_src clean"}]}.
+{pre_hooks, [
+  {"(linux|darwin|solaris|freebsd)", compile,
+        "sh -c \"cd .. && cargo build -p tremor-script-nif && cp target/debug/libtremor.so tremor-script/priv\""}
```
mfelsche, line 9: the `target/debug/libtremor.so` path is wrong. in the build.rs you build it in `target/<TARGET_TRIPLE>/<PROFILE>/libtremor.{so,dll,dylib}`
You want to use the `libpath` file here like so:

```suggestion
        "sh -c \"cd .. && cargo build -p tremor-script-nif && cp $(cat libpath) tremor-script/priv\""}
```

tremor-script/rebar.config
```diff
@@ -2,32 +2,35 @@
...
+]}.
 
-{port_specs, [{"priv/tremor_script.so",["c_src/*.c"]}]}.
+{port_specs, [{"priv/libtremor.so", ["c_src/*.c"]}]}.
```
mfelsche, line 12: this should also have a platform specific suffix, no?

Licenser: we only compile for unix systems so not needed

tremor-script/rebar.config
Licenser: the path is correct, `libpath` isn't a thing here

tremor-script/rebar.config
mfelsche: but we also compile for darwin, there it needs to be .dylib, no?

tremor-script/rebar.config
mfelsche: according to the `build.rs` above this file is created during compilation.

tremor-script/rebar.config
Licenser: so far this wasn't an issue but also the line is from the old code not the new one and can go away - which solves the problem one way or the other :D

mfelsche: ah yeah, remove it then. :D

tremor-script/rebar.config
mfelsche: yeah, that one. :D Sorry, i have not been very specific

Licenser: hah no worries I just didn't make the mental connection :) the content of the file was also wrong, had to fix it sneaky files...

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Added new config `socket_options` to UDP, TCP and Websocket connectors to enable use of socket option `SO_REUSEPORT`
 * Added the `chunk` postprocessor for creating payloads as close to a configurable `max_bytes` as possible
 * Arithmetic expressions now error on overflows
+* Rust based erlang nif for trmeor-script EQC testing
```
mfelsche, line 10:
```suggestion
* Rust based erlang nif for tremor-script EQC testing
```

mfelsche (dismissed):

darach (approved):
Love the rust nif 👍

tremor-script-nif/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+rustler::atoms! {
+    ok,
+}
+
+#[rustler::nif]
+fn eval<'e>(env: Env<'e>, src: &str) -> NifResult<Term<'e>> {
+    let reg: Registry = registry::registry();
+    let script = Script::parse(src, &reg).map_err(|_| NifError::Atom("compilation"))?;
+
+    let mut event = Value::object();
```
mfelsche, line 47: receiving an erlang term and transforming it into an object would be so cool

mfelsche (approved):

tremor-script-nif/src/lib.rs
Licenser: for v2, but it's now semi possible, rustler has a serde implementation so does `value` muhahah

PR #1261, opened 2021-10-15
State windows

Licenser:
# Pull request

## Description

Adds a new window type, state windows, that allow to define windowing borders based on a script

## Related

* [RFC](https://rfcs.tremor.rs/0000-.../)
* Related Issues: fixes #000, closed #000
* Related [docs PR](https://github.com/tremor-rs/tremor-www-docs/pull/000)

## Checklist

* [ ] The RFC, if required, has been submitted and approved
* [ ] Any user-facing impact of the changes is reflected in docs.tremor.rs
* [ ] The code is tested
* [x] Use of unsafe code is reasoned about in a comment
* [ ] Update CHANGELOG.md appropriately, recording any changes, bug fixes, or other observable changes in behavior
* [ ] The performance impact of the change is measured (see below)

PR #2099, opened 2022-11-28
State windows

Licenser:
# Pull request

This is a update of #1261

## Description

Adds a new window type, state windows, that allow to define windowing borders based on a script.

In introduces a new window that can be used by providing a initial `state` a `script` and a `script from tick`. It allows for completely custom windows that, open close and emit based on operator supplied logic.

It re-uses the syntax of `state` and `script from` introduced with the script enhancements:

```tremor
define window w from tumbling
state
  null                      # initial state
script from tick            # script executed in tick events
  state + 1
  state % 10 == 0           # return a bool if the window should emit
script                      # script executed during normal events
  let state = state + event;
  state % 10 == 0           # return a bool if the window should emit
end;
```

It is a net new window type that doesn't impact existing windows

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/274)

## Checklist

* [x] The code is tested

## Performance

-/- net new addition


tests/queries/window_state_by_two/query.trickle
```diff
@@ -0,0 +1,12 @@
+use tremor::origin;
+define window by_two from tumbling
+state
+  0
+script
```
mfelsche, line 5: should this use the `script from tick`?

tests/query_runtime_errors/bad_state_window_return/error.txt
```diff
@@ -0,0 +1 @@
+Error: Window returned `"snot"` this can't be used as a window reply
```
mfelsche, line 1: The error would be more helpful if it was describing what can be used as a window reply.

```suggestion
Error: Window returned `"snot"` this can't be used as a window reply
```

tests/query_runtime_errors/bad_vec_state_window_return/error.txt
```diff
@@ -0,0 +1 @@
+Error: A array return needs to be a two element array of booleans with the form `[emit, include]`
```
mfelsche, line 1:
```suggestion
Error: An array as return value needs to be a two element array of booleans with the form `[emit, include]`
```

tremor-pipeline/src/op/trickle/window.rs
```diff
@@ -487,6 +497,117 @@ impl Trait for No {
...
+            &context,
+            AggrType::Emit,
+            unwind_event,    // event
+            &mut self.state, // state for the window
+            event_meta,      // $
+        ));
+        match value {
+            Return::Emit { value, .. } => Actions::try_from(&value),
+            Return::EmitEvent { .. } => Actions::try_from(&*unwind_event),
+            Return::Drop { .. } => Err("State based window didn't provide a boolean return".into()),
```
mfelsche, line 579: It would be cool to include the window name here, for finding out which window caused problems.

tremor-pipeline/src/op/trickle/window.rs
```diff
@@ -487,6 +497,117 @@ impl Trait for No {
...
+        let context = EventContext::new(ns, None);
+        let mut unwind_event = Value::const_null();
+        let mut event_meta = Value::const_null();
+        // We protect against altering event by ensuring event and meta
+        // arent accessed during 'compile' time
+        if let Some(tick_script) = &self.tick_script {
+            let value = tick_script.run(
+                &context,
+                AggrType::Emit,
+                &mut unwind_event, // event
```
mfelsche, line 593: would it make sense to have `event` be the tick nanoseconds?

tremor-pipeline/src/op/trickle/window.rs
```diff
@@ -487,6 +497,117 @@ impl Trait for No {
...
+                &mut self.state,   // state for the window
+                &mut event_meta,   // $
+            )?;
+
+            match value {
+                Return::Emit { value, .. } => Actions::try_from(&value),
+                Return::EmitEvent { .. } => Actions::try_from(&unwind_event),
+                Return::Drop { .. } => {
+                    Err("State based window didn't provide a boolean return".into())
```
mfelsche, line 602: again, would be cool to have the window name here.

tremor-script/src/ast/visitors/windows.rs
```diff
@@ -0,0 +1,73 @@
...
+        Ok(VisitRes::Walk)
+    }
+}
+impl<'script> ImutExprVisitor<'script> for OnlyMutState {}
+
+#[derive(Debug, Default, Clone, Copy)]
+pub(crate) struct NoEventAccess {}
+
+impl NoEventAccess {
+    // Validates that only state is every mutated
```
mfelsche, line 52:
```suggestion
    // Validates that only state is every accessed, not `event` or `$`
```

tests/queries/window_state_by_two/query.trickle
Licenser: no `tick` is for the periodic tick we use for timed windows, is is a group by two window that groups two events together it doesn't need ticks

mfelsche: Considering this as an example it could do a `drop` in the `script from tick` branch? We'd also test that the `tick` port handling works.

tremor-pipeline/src/op/trickle/window.rs
Licenser: hm I was thinking about it but decided against it.

The tick script is triggered synthetically and not from any `event`. I found calling the data in here `event` will be misleading as it suggests there is an event triggering this. 

To emphasize this using `event` is completely forbidden in the trick script and will error.

mfelsche: So if i want to have the current time to do some time based window, i need to call `system::nanotime()` everytime?

Licenser: yes, same as you'd do in the `script` section code can be copied between then this way :)

mfelsche: It is sad, as we have the tick info already available. Would it make sense to have a function in the `std::win` module that just gets that number without asking the CPU for a time?

mfelsche:
Can we also add a function for creating the two-element `[emit, include]` array. Maybe constants? So writing windows is easier?

Licenser:
I like that!

tremor-pipeline/src/op/trickle/window.rs
mfelsche: Cool, that needs to be documented!

Licenser: done

tremor-script/lib/tremor/windows.tremor
```diff
@@ -0,0 +1,10 @@
+### Helpers for state based windows
+
+## Emits the window and include the current event into the **emitted** window.
+const EMIT_AND_INCLUDE = [true, ture];
```
mfelsche, line 4:
```suggestion
const EMIT_AND_INCLUDE = [true, true];
```

tremor-script/lib/tremor/system.tremor
```diff
@@ -38,6 +38,11 @@ intrinsic fn instance() as system::instance;
 
 ## Returns the current time in epoch nanoseconds
 ##
+## WARNING: This function is expensive, please
+## avoid using it if you can use `ingest_ns` instead.
```
mfelsche, line 42:
```suggestion
## avoid using it. If you can use `ingest_ns` instead.
```

tremor-script/src/ast/visitors/windows.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020-2021, The Tremor Team
```
mfelsche, line 1:
```suggestion
// Copyright 2022, The Tremor Team
```

mfelsche (dismissed):

mfelsche (approved):

PR #2107, opened 2022-12-01
Fix json codec to deduplicate keys consistently

mfelsche:
# Pull request

## Description

The following payload was preserved when read via the `json` codec:

```json
{
  "key": 1,
  "key": 2
}
```

When using the `json` extractor or assigning the resulting value to `state`, the keys were finally deduplicated. Which is wildly inconsistent behaviour.

The `json` codec was also serializing those duplicate keys, which can lead to problems with JSON parsers rejecting documents with duplicate values.

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/273)

## Checklist

* [x] The code is tested

## Performance

main:

```
Throughput   (data): 339.0 MB/s
Throughput (events): 565.1k events/s
```

this PR:

```
Throughput   (data): 331.4 MB/s
Throughput (events): 552.4k events/s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixes
 
 * Fix `udp_client` connector connecting to IPv6 hosts.
+* Fix `json` codec to deduplicate duplicate keys by using the last key in the JSON record.
 
 ### Breaking Changes
 
--- a/src/codec/json.rs
+++ b/src/codec/json.rs
@@ -173,4 +173,43 @@ mod test {
         );
         Ok(())
     }
+
+    #[test]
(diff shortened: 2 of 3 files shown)


Licenser (approved):
huzza!

Licenser (approved):

mfelsche:
@Licenser you happy with merging it like that? Or do we want to introduce a runtime switch for people that know that there will be no duplicates that do not want to pay the toll on performance?

Licenser:
I've no objections but I'm very unsure about the benchmark results, 30mb/s is super slow I would expect something an order of magnitude larger.

PR #2113, opened 2022-12-02
Fix metronome configs

mfelsche:
They were assuming interval to be milliseconds, although the `interval` is interpreted as nanoseconds.

# Pull request

## Related

* Related [docs PR](https://github.com/tremor-rs/tremor-www/pull/275)

## Checklist

* [x] The code is tested

### DIFF
--- a/packaging/distribution/etc/tremor/config/main.troy.sample
+++ b/packaging/distribution/etc/tremor/config/main.troy.sample
@@ -5,9 +5,11 @@ define flow main
 flow
   use tremor::pipelines;
   use bench;
+  use std::time::nanos;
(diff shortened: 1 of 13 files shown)


Licenser (commented):
Should this be nanoseconds? That is our default time unit everywhere else it is possibly confusing to users that, unlike every other time we have, metronome uses milliseconds instead of nanoseconds.

mfelsche:
metronome is using nanoseconds, the documentation and the examples are out of date. This is what i am fixing here.

Licenser (approved):
Aproooooooove!

PR #2114, opened 2022-12-02
Upper case consts and warning improvements

Licenser:
# Pull request

## Description

This PR renames all constants to the ALL_UPPER_CASE convention. It also improves the warning sub-system to allow for better warnings and introduces a number of new warnings for performance and behaviour.

## Related

* [Docs PR](https://github.com/tremor-rs/tremor-www/pull/276)

## Checklist

* [x] The code is tested

## Performance

-/-


Licenser:
What do you all think about including function warnings in the docs? That could be pretty nice to print them right there.

tests/script_warnings/match_imut_multiple_default/warning.txt
```diff
@@ -1,7 +1,7 @@
...
+Warning(behaviour): 
     1 | merge event of match true of
     2 |   case true => true
-    3 |   default => true
-    4 |   default => false
+    3 |   case _ => true
+    4 |   case _ => false
     5 | end end
-      | ^^^ A match statement with more then one default clause will never reach any but the first default clause.
+      | ^^^ A match statement with more then one default clause will never reach any but the first default clause.
```
mfelsche, line 7: this should be an error about unreachable code.

tremor-script/src/ast/helper.rs
```diff
@@ -28,14 +28,40 @@ use crate::{
...
 use halfbrown::HashMap;
-use std::{collections::BTreeSet, mem};
+use std::{collections::BTreeSet, fmt::Display, mem};
 
 /// ordered collection of warnings
 pub type Warnings = std::collections::BTreeSet<Warning>;
 
+/// Class of warning that gives additional insight into the warning's intent
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Copy)]
+pub enum WarningClass {
```
mfelsche, line 38: what about moving the warning struct etc. to a new module `warning.rs` and naming this only `Category`?

mfelsche: What about adding a `Deprecation` variant?

And should we remove `std::array::concatenate` and `std::record::combine` if we have constructs that work better? Or should we keep them, as they are nice pointers to the cool constructs we have?

mfelsche (commented):
I like.

Printing the warning string in the docs inside a `warn` admonition makes sense to me!

tremor-script/src/ast/helper.rs
Licenser: 1) warning module 👍 good call, will do

2) deprication warning, yes I had it, but we don't have any so it'll yell at us for dead code, we can add them once we get them :)

3) I was thinking about it and didn't since eventually we want higher order functions and function arguments where we'll need to concatenate and combine as functions

tests/script_warnings/match_imut_multiple_default/warning.txt
mfelsche: But we can do this in another PR, as it was there before already

Licenser: nah it's a good point I'll make it nicer!

mfelsche (approved):

PR #2129, opened 2022-12-13
Add tremor codec

Licenser:
# Pull request

## Description

This adds a new `tremor` codec. The codec is a binary representation of the `tremor_value::Value`. It includes versioning for later upgrades.

## Checklist

* [x] The code is tested

## Performance

None but in local tests the codec is about 20-30% faster than `json` for the test data


src/codec/tremor.rs
```diff
@@ -0,0 +1,423 @@
...
+    fn decode<'input>(
+        &mut self,
+        data: &'input mut [u8],
+        _ingest_ns: u64,
+    ) -> Result<Option<Value<'input>>> {
+        Tremor::decode(data).map(Some)
+    }
+
+    fn encode(&self, data: &Value) -> Result<Vec<u8>> {
+        let mut dst = Vec::with_capacity(512);
```
mfelsche, line 260: this could be made configurable, so people could optimize for their expected sizes.

src/codec/tremor.rs
```diff
@@ -0,0 +1,423 @@
...
+    const BOOL_TRUE: u8 = 1;
+    const BOOL_FALSE: u8 = 2;
+    const I64: u8 = 3;
+    const U64: u8 = 4;
+    const F64: u8 = 5;
+    const STRING: u8 = 6;
+    const BYTES: u8 = 7;
+    const ARRAY: u8 = 8;
+    const OBJECT: u8 = 9;
+    const VERSION: u32 = 0;
```
mfelsche, line 61: you can totally make this a `u8`. We likely won't make it past version 255 anytime...

src/codec/tremor.rs
```diff
@@ -0,0 +1,423 @@
...
+            }
+            StaticNode::Bool(true) => w.write_u8(Self::BOOL_TRUE)?,
+            StaticNode::Bool(false) => w.write_u8(Self::BOOL_FALSE)?,
+            StaticNode::Null => w.write_u8(Self::NULL)?,
+        }
+        Ok(())
+    }
+    #[inline]
+    #[allow(clippy::cast_possible_truncation)]
+    fn write_len_type<E: ByteOrder>(t: u8, len: usize, w: &mut impl Write) -> Result<()> {
```
mfelsche, line 189:
```suggestion
    fn write_type_and_len<E: ByteOrder>(t: u8, len: usize, w: &mut impl Write) -> Result<()> {
```

src/codec/tremor.rs
Licenser: We don't have a way to make codecs configurable, I like the idea, I would love to generalize this as we have a few codecs that have the same logic, but I don't think  this is the right PR to "make codecs configurable". I'll ticket it!

src/codec/tremor.rs
Licenser: that's what they said with the Y2K problem :P lets pick the middle and make it 16?

src/codec/tremor.rs
Licenser: fild under #2130

mfelsche: Codecs are configurable. `resolve` takes a `NameWithConfig`.
It is just that none of the existing codecs is doing it as of now.
Yours could be the first.

For inspiration on how to do it look into the preprocessors for example :)

Licenser: Nice, so I did change the whole thing around to avoid the problem and let the codec remember what the sizes are and allocate to fit the largest so it'll armotize against a single allocation w/o user interaction now :D

src/codec/tremor.rs
mfelsche: your call :)

src/codec/tremor.rs
Licenser: We came to a solution on this so I'm closing it :)

darach (approved):
LGTM 👍

PR #2148, opened 2022-12-20
Handle routing information for each bulk item in the elastic connector

mfelsche:
# Pull request

## Description

Previously we didnt handle routing metadata for single bulk items in a batched event, now we do.

## Checklist

* [x] The code is tested


Licenser (dismissed):
🚀

src/connectors/impls/elastic.rs
```diff
@@ -703,101 +704,156 @@ impl<'a, 'value> ESMeta<'a, 'value> {
...
+        Ok(())
+    }
+
+    fn insert_update_op(&self, data: &Value, ops: &mut BulkOperations) -> Result<()> {
+        let mut op = self
+            .get_id()
+            .map(|id| {
+                // we are sure that this BulkUpdateOperation lives only as long as the event
+                // so it is fine to only do a `.clone()` here instead of a `.clone_static()`
+                let src = literal!({ "doc": data.clone() });
```
mfelsche, line 819: this one is indeed serialized as soon as it is pushed below, so this is *totally* safe :P

Licenser (approved):

Licenser (approved):
I think we also need a docs update to indicate that we now have `routing` as an option

mfelsche:
It already says it is there in the docs, thats why this is a bug.

Licenser:
Oh sneaky so the docs PR was done long before the code PR :joy: more of a very long lag between code and docs then :P

PR #2219, opened 2023-02-08
Raft

Licenser:
# Pull request

## Description

Clustering 

## Related

## Checklist

* [ ] The code is tested


.github/workflows/integration.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
 
     env:
       TREMOR_PATH: "${{ github.workspace }}/tremor-script/lib:${{ github.workspace }}/tremor-cli/tests/lib"
-      RUSTFLAGS: -D warnings -C target-feature=${{ matrix.target_feature }} --cfg tokio_unstable
```
darach, line 33: What do we need `tokio_unstable` for?

Licenser: we don't this removes it :)

k8s/README.md
```diff
@@ -0,0 +1,30 @@
+# NO!
+Experimental, for testing, don't use in production or if you do fix all the things.
+
+Based on https://kubernetes.io/docs/tutorials/stateful-application/cassandra/
+
+this are purely notes
```
darach, line 6:
```suggestion
These are purely notes
```

packaging/cross_build.sh
```diff
@@ -113,7 +113,7 @@ echo "Successfully built the binary: ${TARGET_BIN}"
 # linking check
 echo "Printing linking information for the binary..."
 file "$TARGET_BIN"
-ldd "$TARGET_BIN"
+ldd "$TARGET_BIN" || true
```
darach, line 116: Ok, this definitely needs at least a comment!

Licenser: It would, I have no idea why this was there :D the good news it it isn't needed so I'll just remove it

k8s/README.md
Licenser: fixed locally with other changes to reduce commit clutter

Licenser:
We should eventually discuss how to deal with this PR, the connector separation did force a quite different architecture to raft so there are a lot of conflicting parts in this PR

It will at least need a rearchitrecting of how the raft subsystem interacts with connectors - which is probably a significant change

the simplest solution might be to completely kill the idea of a raft based connector or at least the only idea I have right now without leaking raft into every connector which seems icky

PR #2324, opened 2023-05-12
Move google connectors to use  instead of env var

Licenser:
# Pull request

## Description

The old code did only work with a single environment variable this makes it impossible to use tremor with more then a single service account. This change instead allows specifying either the file to read from (equivalent from the env variable) or as a embedded json.

## Related

## Checklist

* [ ] The code is tested

## Performance
-/-


src/connectors/impls/gcs/resumable_upload_client.rs
```diff
@@ -338,11 +339,15 @@ impl<TClient: HttpClientTrait, TBackoffStrategy: BackoffStrategy + Send + Sync>
 impl<TClient: HttpClientTrait, TBackoffStrategy: BackoffStrategy>
     DefaultClient<TClient, TBackoffStrategy>
 {
-    #[allow(clippy::unnecessary_wraps)] // test requerst a result
```
darach, line 341: Did you mean? // test requires a result

src/connectors/impls/gcs/resumable_upload_client.rs
```diff
@@ -338,11 +339,15 @@ impl<TClient: HttpClientTrait, TBackoffStrategy: BackoffStrategy + Send + Sync>
...
-    pub fn new(client: TClient, backoff_strategy: TBackoffStrategy) -> Result<Self> {
+    #[allow(clippy::unnecessary_wraps, unused_variables)] // test requerst a result
```
darach, line 342: Ditto, did you mean? // test requires a result

darach (changes requested):
LGTM - just the 2 typos in the comments

src/connectors/impls/gcs/resumable_upload_client.rs
Licenser: resolving this as it's the delete of the next line but still true :) the message should be better

Agares (approved):

Licenser:
I'll do the docs in #2325 as doing them now would just mean moving them there 5 minutes after

PR #2328, opened 2023-05-25
Add axiom http and syslog endpoint

Licenser:
# Pull request

## Description

Connector definitions to send data to axiom

## Checklist

* [x] The code is tested

-/-

### DIFF
--- a/tremor-script/lib/axiom.tremor
+++ b/tremor-script/lib/axiom.tremor
@@ -1,0 +1,61 @@
+### # Tremor language Axiom integration.
+###
+### Integration with [Axiom](https://axiom.co).
+###
+
+use std::time::nanos;
+
+## Axiom HTTP client to send logs to the axiom http endpoint.
+##
+## arguments:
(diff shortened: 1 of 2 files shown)


tremor-script/lib/axiom.tremor
```diff
@@ -0,0 +1,61 @@
+### # Tremor language Axiom integration.
```
darach, line 1: Optional. Scoping to `integrations::axiom` might make more sense given other integrations are likely and encouraged.

Licenser: I was thinking about that too, then I noticed integrations we have are also top level, so I went with the same model we used there so it's a uniform interface for users.

I think it's also nice, eventually we should have some kind of package manger and doing `targo install axiom` would feel a lot more natural then `targo install integrations::axiom`

darach: Ok 👍 - extra comedy points for `targo install` 🤡

PR #2338, opened 2023-07-21
Add hygienic error for using strings in path, fixes #2337

### LINKED ISSUE #2337: Consider opening up path and record name syntax to wider syntax for better usability

The use of backtick for idents, and double-quote for record field names can be confusing
at times.

```tremor
    1 | let $gcl_writer = {
    2 |   "timestamp": event."@timestamp"
      |                      ^ Found the token `"` but expected `<ident>`
    3 | };
 ```

Given quoted strings are illegal as path segments:
* Allow double-quoted string as ident in path segment?

And as "@timestamp" is a legal field name in tremor perhaps:
* Allow record field names to be idents, or backtick idents

**Notes**

If we made the above changes, the following would be legal tremor, but illegal json:

```tremor
{ foo: "bar" }
```

However, if we made such a change to tremor it should always serialise to json correctly when a json
codec is in force.

As this changes usability and the language, would need an RFC and community discussion.

Licenser:
## Description

Add a hygienic error for using a string as a path element instead of a quoted ident.

## Related

* Related Issues: fixes #2337

## Checklist

* [x] The code is tested

## Performance

-/-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New features
 
+* hygienic errors for strings in path segements
 * axiom stdlib module
 * gcl now supports the `timestamp` metadata overwrite
 
--- a/tests/script_error.rs
+++ b/tests/script_error.rs
@@ -129,6 +129,7 @@ test_cases!(
     pp_cyclic,
     pp_nest_cyclic,
     // INSERT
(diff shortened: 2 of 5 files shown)


darach (approved):
Nice.

PR #2343, opened 2023-08-14
Add avro codec

### LINKED ISSUE #379: Add avro codec

Tremor can’t encode/decode data in [avro](https://avro.apache.org/docs/current/spec.html) format.

Add a [codec](https://docs.tremor.rs/artefacts/codecs/) to support it.

**Notes**

There’s [avro-rs](https://github.com/flavray/avro-rs) out there which should be useful here. It uses serde-json under the hood but we can probably modify it to use simd-json too.

Codecs in tremor so far don’t take any configuration, but in this case, we would need to support schemas, either provided in-line in the config or sourced from external schema registries (eg: like the one from [confuent](https://docs.confluent.io/current/schema-registry/schema_registry_tutorial.html)).

Licenser:
# Pull request

## Description

A simple encoder and decoder of the apache avro binary codec

## Related

* Related Issues: fixes #379

## Checklist

* [x] The code is tested

## Performance

-/-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * hygienic errors for strings in path segements
 * axiom stdlib module
 * gcl now supports the `timestamp` metadata overwrite
+* basic avro codec
 
 ### Breaking Changes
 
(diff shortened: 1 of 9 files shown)


darach (approved):
LGTM. 👍

PR #2349, opened 2023-09-18
Expanded for comprehsions with folding

Licenser:
# Pull request

## Description

This PR expands `for` comprehension to allow for initial values and a fold operator. Both are optional and default, so the current behavior is maintained, and full backward compatibility is given.

The syntax changes from:

```
for <target> of
  <cases>
end;
```

to

```
for <target> of
  <cases>
[into <initial> [use <operator>]] end;
```

where `initial` defaults to `[]` and `operator` defaults to `+`.

Vaid operators are `+`, `-`, `*` and `/`.

The syntax leaves open the opportunity to expand `operator` to have more complex folding using lambdas or function pointers once those are implemented. 

## Checklist

* [x] The code is tested

## Performance

Summary: We gain nearly 20% performance for the record case while maintaining near parity / stay within the noise level for arrays.

Simple "for" loop creating an array

```tremor
    define script runtime
    script
      for event of
        case (k, v) => v
      end
    end;
```

Creating a record:

```tremor
# old
    define script runtime
    script
      use std::record;
      record::from_array(for event of
        case (k, v) => [k, v]
      end)
    end;
# new
    define script runtime
    script
      for event of
        case (k, v) => {"{k}": v}
      into {} end
    end;

```

| target | old | new| 
|-|-|-|
| array | 107.6 MB/s | 108.6 MB/s |
| record | 121.4 MB/s | 102.2 MB/s |


Licenser:
I got to admit, I'm disappointed by the changes in performance, but it does read nicer at least

Licenser:
I am no longer disappointed by the performance reasults

tremor-script/src/grammar.lalrpop
```diff
@@ -1250,7 +1250,34 @@ Merge: MergeRaw<'input> = {
...
+        fold: ComprehensionFoldOpRaw(BinOpKind::Add)
+    },
+    <start:@L> "for" <target:ComplexExprImut> "of" <cases:ForCaseClauses> "into" <initial:ComplexExprImut> "end" <end:@L> => ComprehensionRaw{
+        target,
+        cases,
+        mid: NodeMeta::new_box(start, end),
+        initial,
+        fold: ComprehensionFoldOpRaw(BinOpKind::Add)
+    },
+    <start:@L> "for" <target:ComplexExprImut> "of" <cases:ForCaseClauses> "into" <initial:ComplexExprImut> "use" <fold:FoldOperator> "end" <end:@L> => ComprehensionRaw{
```
mfelsche, line 1267: can't this be solved with 1 rule here where the `use` and `into` clauses are optional? That is, are marked with `?`

static/language/epilog/for.md
```diff
@@ -7,5 +7,39 @@ An illustration of the rule is a set of case statements, followed by an optional
...
+
+```tremor
+  for event of
+    case (k,v) => v
+  into 1
+  use *
+  end
+```
+
+The current supported operators in `use` are: `+`, `-`, `*` and `/`.
```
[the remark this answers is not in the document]
mfelsche: And what if we use `-` or `/` on folding records. Will the correct (helpful) error message appear?

mfelsche (commented):
Looking good overall!

Some minor nitpicks.

mfelsche:
Would this also work as expected on strings with `+`?
I do think this table here: https://www.tremor.rs/docs/edge/language/expressions#operators needs some updating. And it should be referenced from the for-comprehension docs, in order to show what operations are allowed in the context of a for comprehension.

darach:
I do not like this syntax at all and believe we need an RFC as head/tail pattern matching in functions would resolve this far more elegantly.

tremor-script/src/grammar.lalrpop
Licenser: I've reduced it to too, as it simplifies the code in the implementation

static/language/epilog/for.md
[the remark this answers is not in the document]
Licenser: I improved the errors on that :)

tremor-script/src/ast/raw.rs
```diff
@@ -1256,9 +1256,17 @@ where
     fn up<'registry>(self, helper: &mut Helper<'script, 'registry>) -> Result<Self::Target> {
         // We compute the target before shadowing the key and value
 
+        if self.fold.is_some() {
+            helper.warn(
+                self.extent().expand_lines(2),
+                self.extent(),
+                &"The `use <op>` syntax is experimental and may be deprecated or replaced",
```
darach, line 1263: 👍  This is great, keeps the door open for [H|T] functions and functional fold down the road

Licenser:
https://github.com/tremor-rs/tremor-www/pull/302

mfelsche (approved):

PR #2447, opened 2024-02-28
Fix feature = "cargo-clippy" deprecation

rex4539:
https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html


Licenser (changes requested):
The change is nice but it seems to create a lot of clippy which would need to be resolved before we can merge it

Licenser:
Seems like `cargo fmt` is needed then it should be good :)

Licenser (approved):
Awesome, thanks again!

PR #2497, opened 2024-04-18
Try to simplify Tests

Licenser:
# Pull request

## Description

right now our branch rules need to be added for every single matrix combination :sob: that allows us to just use the last element in the chain to enforce it so adding / removing from the matrixes no long requires reconfiguration

## Checklist

* [x] The code is tested

## Performance

-/-


darach (dismissed):
LGTM 👍

darach (approved):
LGTM 👍  except the makefile bits, ha ha!

PR #2583, opened 2024-06-14
Add experimental Microsoft Azure Monitor Log Ingest connector

darach:
# Pull request

## Description

Add experimental Microsoft Azure Monitor Log Ingest connector

## Checklist

* [ x ] The code is tested

## Performance

This is a net new experimental connector and may need tweaks based on live testing which is tbd

### DIFF
--- a/tremor-connectors-azure/Cargo.toml
+++ b/tremor-connectors-azure/Cargo.toml
@@ -28,7 +28,6 @@ async-trait = "0.1.80"
 url = "2.5.0"
 reqwest = { version = "0.12.4", default-features = false, features = ["json"] }
 tokio = "1.37.0"
-azure_identity = "0.20.0"
 log = "0.4.21"
 tremor-script = { version = "0.13.0-rc.25", path = "../tremor-script" }
 bytes = "1.6.0"
--- a/tremor-connectors-azure/Cargo.toml
+++ b/tremor-connectors-azure/Cargo.toml
@@ -11,6 +11,12 @@ homepage = "https://www.tremor.rs"
 license = "Apache-2.0"
 
 [dependencies]
+tremor-connectors = { path = "../tremor-connectors", version = "0.13.0-rc.28", default-features = false }
+tremor-common = { path = "../tremor-common", version = "0.13.0-rc.28" }
+tremor-system = { path = "../tremor-system", version = "0.13.0-rc.28" }
+tremor-value = { path = "../tremor-value", version = "0.13.0-rc.28" }
+tremor-config = { path = "../tremor-config", version = "0.13.0-rc.28" }
+
 azure_core = { version = "0.20.0", default-features = false, features = [
     "enable_reqwest",
     "enable_reqwest_gzip",
@@ -36,26 +42,6 @@ serde_json = "1.0.117"
 beef = "0.5.2"
 simd-json = "0.13.10"
 mockito = "1.4.0"
-
-[dependencies.tremor-config]
-# version = "0.0.1"
-path = "../tremor-config"
-
-[dependencies.tremor-connectors]
-# version = "0.0.1"
-path = "../tremor-connectors"
-
-[dependencies.tremor-common]
-# version = "0.0.1"
-path = "../tremor-common"
-
-[dependencies.tremor-value]
-# version = "0.0.1"
-path = "../tremor-value"
-
-[dependencies.tremor-system]
-# version = "0.0.1"
-path = "../tremor-system"
 
 [features]
 integration-harness = []
--- a/tremor-connectors-azure/Cargo.toml
+++ b/tremor-connectors-azure/Cargo.toml
@@ -1,0 +1,78 @@
+bin = []
+bench = []
+test = []
+
+[[example]]
+path = "examples/test.rs"
+name = "test"
+plugin = false
+proc-macro = false
+edition = "2021"
+required-features = []
+
+[package]
+name = "tremor-connectors-azure"
+edition = "2021"
+version = "0.0.1"
+authors = ["The Tremor Team"]
+homepage = "https://www.tremor.rs"
+license = "Apache-2.0"
+
+[dependencies]
+azure_core = { version = "0.20.0", default-features = false, features = [
+    "enable_reqwest",
+    "enable_reqwest_gzip",
+    "enable_reqwest_rustls",
+] }
+serde = "1.0.202"
+anyhow = "1.0.83"
+async-trait = "0.1.80"
+url = "2.5.0"
+reqwest = { version = "0.12.4", default-features = false, features = ["json"] }
+tokio = "1.37.0"
+azure_identity = "0.20.0"
+log = "0.4.21"
+tremor-script = { version = "0.13.0-rc.25", path = "../tremor-script" }
+bytes = "1.6.0"
+futures = { version = "0.3.30", default-features = false, features = [
+    "alloc",
+    "std",
+    "async-await",
+    "futures-executor",
+] }
+serde_json = "1.0.117"
+beef = "0.5.2"
+simd-json = "0.13.10"
+mockito = "1.4.0"
+
+[dependencies.tremor-config]
+# version = "0.0.1"
+path = "../tremor-config"
+
+[dependencies.tremor-connectors]
+# version = "0.0.1"
+path = "../tremor-connectors"
+
+[dependencies.tremor-common]
+# version = "0.0.1"
+path = "../tremor-common"
+
+[dependencies.tremor-value]
+# version = "0.0.1"
+path = "../tremor-value"
+
+[dependencies.tremor-system]
+# version = "0.0.1"
+path = "../tremor-system"
+
+[features]
+integration-harness = []
+
+[lib]
+path = "src/lib.rs"
+name = "tremor_connectors_azure"
+plugin = false
+proc-macro = false
+edition = "2021"
+required-features = []
+crate-type = ["rlib"]
--- a/tremor-connectors-azure/Cargo.toml
+++ b/tremor-connectors-azure/Cargo.toml
@@ -5,7 +5,7 @@ test = []
 [package]
 name = "tremor-connectors-azure"
 edition = "2021"
-version = "0.0.1"
+version = "0.13.0-rc.28"
 authors = ["The Tremor Team"]
 homepage = "https://www.tremor.rs"
 license = "Apache-2.0"
--- a/tremor-connectors-azure/Cargo.toml
+++ b/tremor-connectors-azure/Cargo.toml
@@ -2,14 +2,6 @@ bin = []
 bench = []
 test = []
 
-[[example]]
-path = "examples/test.rs"
-name = "test"
-plugin = false
-proc-macro = false
-edition = "2021"
-required-features = []
-
 [package]
 name = "tremor-connectors-azure"
 edition = "2021"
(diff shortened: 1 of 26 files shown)


tremor-connectors-azure/src/monitor/ingest/writer.rs
```diff
@@ -0,0 +1,428 @@
...
+        serializer: &mut EventSerializer,
+    ) -> anyhow::Result<()> {
+        let chunks = serializer
+            .serialize_for_stream_with_codec(value, meta, ingest_ns, self.id.get(), None)
+            .await?;
+        self.append_data(chunks);
+        Ok(())
+    }
+
+    fn append_data(&mut self, chunks: Vec<Vec<u8>>) {
```
Licenser, line 137: This is probably fine for a experimental connector but to release it we'll want a streaming sender. Let's add a TODO so we remember?

tremor-connectors-azure/src/monitor/ingest/writer/sink.rs
```diff
@@ -0,0 +1,296 @@
...
+                path: vec![],
+            },
+            concurrency_cap,
+            source_is_connected,
+            serializer: json_serializer(),
+        }
+    }
+}
+
+#[allow(clippy::unwrap_used)] // NOTE we know that the codec is always present/correct
```
Licenser, line 88: let's still not do that, it introduces headache later on if we ever extend the function. We an bubble up the error

tremor-connectors-azure/src/monitor/ingest/writer/sink.rs
```diff
@@ -0,0 +1,296 @@
...
+                    }
+                    Some(builder) => {
+                        builder
+                            .append(value, meta, start, &mut self.serializer)
+                            .await?;
+                    }
+                }
+            }
+
+            assert!(builder.is_some(), "No builder");
```
Licenser, line 172: This will panic the runtime, we want to use an error instead, we can completely remove this since builder.some is checked in line 179 and will error corretly

tremor-connectors-azure/src/monitor/ingest/writer/sink.rs
```diff
@@ -0,0 +1,296 @@
...
+                match timeout(t, response).await {
+                    Ok(Ok(response)) => {
+                        let response_meta = extract_response_meta(response).await?;
+                        // let headers = response.headers();
+
+                        if let Some(response_tx) = response_tx {
+                            let mut meta = task_ctx.meta(literal!({
+                                "request": std::convert::Into::<Value>::into(req_meta),
+                                "request_id": request_id.get(),
+                                // "response": std::convert::Into::<Value>::into(response_meta),
```
Licenser, line 211: the response seems interesting? Any reason why we exclude it?

tremor-connectors-azure/src/rest.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl std::fmt::Display for RequestId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+pub(crate) struct RequestMeta {
+    method: String,
```
Licenser, line 44: We have enums for that flying around somewhere :) we can avoid stringly typing the method

tremor-connectors-azure/src/rest.rs
```diff
@@ -0,0 +1,222 @@
...
+        Self {
+            status: 0,
+            headers: Headers::new(),
+            data: Value::Static(StaticNode::Null),
+            content_length: 0,
+        }
+    }
+}
+
+#[allow(clippy::unnecessary_wraps)]
```
Licenser, line 120: let's not add allows

tremor-connectors-azure/src/rest.rs
```diff
@@ -0,0 +1,222 @@
...
+pub(crate) fn extract_request_meta(request: &Request) -> Result<RequestMeta, Error> {
+    let mut meta = RequestMeta::default();
+    meta.method = request.method().to_string();
+    meta.url = request.url().to_string();
+    meta.headers = request.headers().clone();
+    Ok(meta)
+}
+
+pub(crate) async fn extract_response_meta(response: Response) -> Result<ResponseMeta, Error> {
```
Licenser, line 129: This seems to be mixing data and metadata. I think we wan't the body data in the event data section not the metadata. that also avoids the casting to static and avoids a bunch of cloning

tremor-connectors/src/impls/http/meta.rs
```diff
@@ -131,7 +131,10 @@ fn extract_headers(headers: &HeaderMap) -> Result<Value<'static>, ToStrError> {
         .collect::<Result<Value<'static>, ToStrError>>()
 }
 /// Extract request metadata
-pub(super) fn extract_request_meta<T>(
+/// # Errors
+/// - if the request metadata can not be extracted
+#[allow(clippy::module_name_repetitions)] // TODO refactor rename
```
Licenser, line 136: Let's do this now and avoid adding allows

Licenser (commented):
Weeh, I just left a few comments

tremor-connectors-azure/Cargo.toml
```diff
@@ -0,0 +1,78 @@
...
+    "async-await",
+    "futures-executor",
+] }
+serde_json = "1.0.117"
+beef = "0.5.2"
+simd-json = "0.13.10"
+mockito = "1.4.0"
+
+[dependencies.tremor-config]
+# version = "0.0.1"
```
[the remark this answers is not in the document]
darach: Fixed. CI failed on '*' anyway so our CI enforces this!

tremor-connectors-azure/src/auth.rs
```diff
@@ -0,0 +1,281 @@
...
+    #[serde(skip)]
+    #[serde(default = "default_base_url")]
+    pub(crate) base_url: String,
+    pub(crate) client_id: String,
+    pub(crate) client_secret: String,
+    pub(crate) tenant_id: String,
+    #[serde(default = "default_scope")]
+    pub(crate) scope: String, // NOTE Injected by specific sources/sinks on construction
+    #[serde(skip)]
+    token: Mutex<Option<Token>>,
```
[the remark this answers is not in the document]
darach: Given we have concurrency cap, maybe the mutex can be entirely eliminated. I should have thought of that

tremor-connectors-azure/src/auth.rs
```diff
@@ -0,0 +1,281 @@
...
+
+impl Clone for Config {
+    fn clone(&self) -> Self {
+        Self {
+            base_url: self.base_url.clone(),
+            client_id: self.client_id.clone(),
+            client_secret: self.client_secret.clone(),
+            tenant_id: self.tenant_id.clone(),
+            scope: self.scope.clone(),
+            token: Mutex::new(None),
```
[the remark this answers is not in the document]
darach: As above. I think its whackable.

tremor-connectors-azure/src/auth.rs
```diff
@@ -0,0 +1,281 @@
...
+            ("scope", &self.scope),
+        ];
+
+        let client = Client::new();
+        let res = client
+            .post(&url)
+            .form(&params)
+            .send()
+            .await
+            .expect("Failed to refresh token");
```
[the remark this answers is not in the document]
darach: Ja. Already murdered!

tremor-connectors-azure/src/rest.rs
darach: I kept this similar to the http connector for now as I think it'd be worth another go at this at some point and perhaps consolidate on reqwest - to simplify dependencies for the most part

tremor-connectors-azure/src/monitor/ingest/writer.rs
darach: I didn't find a way nice way to make it streaming so punted for now. I'll add a TODO as it makes sense to do some live testing for this

tremor-connectors-azure/src/monitor/ingest/writer/sink.rs
Licenser: Makes sense, that was the context I was missing :) perhaps we throw in a quick comment that we always get 204's so don't need the body?

tremor-connectors-azure/src/monitor/ingest/writer/sink.rs
darach: That was debugging I forgot to cleanup. Whacked and well spotted...

tremor-connectors-azure/src/rest.rs
Licenser: I think we use `SerdeishMethod` (what a wonderful name :D) in `http` to avoid dragging strings aroudn

tremor-connectors-azure/src/monitor/ingest/writer/sink.rs
darach: Yes. I really should add a comment. I also need to live test correlation so there's that too ...

tremor-connectors-azure/src/rest.rs
darach: I didn't deviate much from HTTP client structurally so its no different. The azure-core crate basically builds on reqwest and i'd like to see if we can drop straight to reqwest. That way, perhaps we can drop hyper client from the http connector too. If that works, possibly worth revisiting how we do req/resp metadata.

Punting for now.

Licenser: you said that we always get 204's back so we don't need the body at all,  I guess it's fine then. What I want to avoid is that we release the body in the metadata as that would break all conventions and expectations we established in all the other connectors.

tremor-connectors-azure/Cargo.toml
```diff
@@ -0,0 +1,78 @@
...
+name = "test"
+plugin = false
+proc-macro = false
+edition = "2021"
+required-features = []
+
+[package]
+name = "tremor-connectors-azure"
+edition = "2021"
+version = "0.0.1"
```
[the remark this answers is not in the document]
darach: Totally forgot about version numbers there for a second. A fair cop!

tremor-connectors-azure/src/rest.rs
Licenser: just looked it up, yap it's the one that [never gets use](https://github.com/tremor-rs/tremor-runtime/pull/2583/files/4e3a5785b860270c65afd3a1d34a8b3ca8df551e#diff-b35c45032121e1ecb6dcd1eaacadd2ad13a968fd4de464c0d5905847ac357cb9R211)

perhaps just exclude the body from this as it has no effect and we avoid accidentially pulling it into the response later when we forgot about it?

Licenser: I stand corrected, it is used and pulled apart. What confused me here is the naming as it doesn't align with the use, it is used as a mix of respose data and response metadata and get's used in both places :D

tremor-connectors-azure/src/rest.rs
darach: Azure wraps everything http:

```
use azure_core::headers::Headers;
use azure_core::Request;
use azure_core::Response;
```

So i wanted something KISS for experimenting. Ideally, we can get this based on reqwest directly and drop `azure_core` is what I'm thinking...

Licenser (approved):
wooh nice work!

darach:
Well, it might be nice work when I fix that merge conflict that crept in 🤓

Licenser (approved):

Licenser (approved):

Issue #820, opened 2021-03-09
Enhancements for the gelf-chunking postprocessor


anupdhml:
**Describe the problem you are trying to solve**

The [gelf-chunking](https://docs.tremor.rs/artefacts/postprocessors/#gelf-chunking) postprocessor uses a [numeric message id](https://github.com/tremor-rs/tremor-runtime/blob/v0.10.2/src/postprocessor/gelf.rs#L47) that is incremented on each new message. To better guarantee uniqueness when multiple tremor instances are sending chunked gelf messages to the same receiver, we should choose a different scheme for id.

Also, if the total message size is less than the individual chunk size, we can choose to send the message unchunked -- the gelf protocol supports that. For small-enough messages, this will eliminate the chunking overhead both on the client and server side.

**Describe the solution you'd like**

For message id, [gelf docs](https://docs.graylog.org/en/4.0/pages/gelf.html#chunking) mention using a combination of hostname and timestamp. 

If the length of data that the preprocessor gets is less than the chunk size, send the data as is (i.e. do not apply the chunking logic).

**Notes**

Relevant implementations from the official java client:

https://github.com/Graylog2/gelfclient/blob/gelfclient-1.5.1/src/main/java/org/graylog2/gelfclient/encoder/GelfMessageChunkEncoder.java#L82-L91

https://github.com/Graylog2/gelfclient/blob/gelfclient-1.5.1/src/main/java/org/graylog2/gelfclient/encoder/GelfMessageChunkEncoder.java#L131-L134

Php client implementation for the same:
https://github.com/bzikarsky/gelf-php/blob/1.7.0/src/Gelf/Transport/UdpTransport.php#L88-L96

rameels:
Hi @darach could I please be assigned this issue? Decided to pick it up as a way of learning Rust. 😄

mfelsche:
@rameels You are hereby assigned! Thanks for your initiative. Feel free to drop any questions you might have here or on our discord: https://chat.tremor.rs/

BharatKJain:
Possible fix: https://github.com/tremor-rs/tremor-runtime/pull/2662/files

PR #2662, opened 2024-09-29
Fix: message-id in postprocessor/gelf-chunking

BharatKJain:
# Pull request

## Description

Changed message-id from auto-increment ID to randomized ID in postprocessor/gelf_chunking.rs

HELP NEEDED: I have avoided adding hostname while producing message-id, I am not sure how can we handle adding hostname, please suggest.

## Related

- [Issue-820](https://github.com/tremor-rs/tremor-runtime/issues/820)
- [Reference code](https://github.com/osiegmar/logback-gelf/blob/master/src/main/java/de/siegmar/logbackgelf/MessageIdSupplier.java#L61)

## Checklist

* [x] The code is tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
 ### Fixes
 * gbq connector naming is now `gbq_writer` as it is documented
 * fix gbq connector url missing `/`
-
+* fix GELF message-id from auto-increment to random-id in postprocessor/gelf_chunking.rs 
+  
 ## [0.13.0-rc.29]
 
 ### New features
--- a/tremor-interceptor/Cargo.toml
+++ b/tremor-interceptor/Cargo.toml
@@ -26,6 +26,7 @@ thiserror = { version = "1.0", default-features = false }
 
 #gelf
 rand = { version = "0.8", optional = true, default-features = false, features =  ["std_rng"] }
+sha2 = { version="0.10.8", optional = true, default-feature = false, features = ["std"]}
 
 # Compression
 brotli = { version = "6", optional = true, default-features = false, features = [
@@ -61,7 +62,7 @@ proptest = "1.5"
 [features]
 default = ["base64", "compression", "gelf", "length-prefix"]
 length-prefix = ["dep:bytes"]
-gelf = ["dep:rand"]
+gelf = ["dep:rand", "dep:sha2"]
 base64 = []
 compression = [
     "dep:brotli",
--- a/tremor-interceptor/Cargo.toml
+++ b/tremor-interceptor/Cargo.toml
@@ -25,8 +25,7 @@ anyhow = { version = "1.0", default-features = true }
 thiserror = { version = "1.0", default-features = false }
 
 #gelf
-rand = { version = "0.8", optional = true, default-features = false, features = [
-] }
+rand = { version = "0.8", optional = true, default-features = false, features =  ["std_rng"] }
 
 # Compression
 brotli = { version = "6", optional = true, default-features = false, features = [
--- a/tremor-interceptor/Cargo.toml
+++ b/tremor-interceptor/Cargo.toml
@@ -25,8 +25,7 @@ anyhow = { version = "1.0", default-features = true }
 thiserror = { version = "1.0", default-features = false }
 
 #gelf
-rand = { version = "0.8", optional = true, default-features = false, features =  ["std_rng"] }
-sha2 = { version="0.10.8", optional = true, default-feature = false, features = ["std"]}
+rand = { version = "0.8", optional = true, default-features = false, features =  [] }
 
 # Compression
 brotli = { version = "6", optional = true, default-features = false, features = [
@@ -62,7 +61,7 @@ proptest = "1.5"
 [features]
 default = ["base64", "compression", "gelf", "length-prefix"]
 length-prefix = ["dep:bytes"]
-gelf = ["dep:rand", "dep:sha2"]
+gelf = ["dep:rand"]
 base64 = []
 compression = [
     "dep:brotli",
--- a/tremor-interceptor/src/postprocessor/gelf_chunking.rs
+++ b/tremor-interceptor/src/postprocessor/gelf_chunking.rs
@@ -13,12 +13,36 @@
 // limitations under the License.
 
 //! Splits the data using [GELF chunking protocol](https://docs.graylog.org/en/3.0/pages/gelf.html#chunking).
+//!
+//! ## What's the logic for creating messgae id?
+//! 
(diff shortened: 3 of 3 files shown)


Licenser (approved):
👍  looks reasonable nothing to prevent using this I see, great documentation too :)

two things I notice.

1) it would be nice to mention how message id's are generated in the docs (the `//! ` section)

2) if you are up for a challenge: we try to keep all time and randomness out of tremor to allow for deterministic replays. We do this by using `ingest_ns` for as a random seed, and for times, that way a event that is logged with it's ingest ns can be replayed and generate the exact same out put. [The random function](https://github.com/tremor-rs/tremor-runtime/blob/main/tremor-script/src/std_lib/random.rs#L38) is a good example. It'd be interesting to see this same concept re-used for this to allow repeatable yet still random message id's; one way would be to use ingest-ns instead of the current epoch (which would be nice anyway as looking uo time isn't fast), and then seed the RNG somehow (probably not with the ingest ns as that would make it useless) but perhaps with the fist n bytes of the message? or with some bytes of a hash of the message?

BharatKJain:
1. Okay, will do.

2. Trying to think-out-loud, so let's say if we create a hash of a message but repetitive logs will create same message-id which is a problem because message-id has to be unique in nature, collisions can cause problems when we're decoding the message on the server side. I am not completely sure but ideally we would want to have uniqueness in the message-id to make sure that we are not breaking the server GELF-decoding. 

(How it will break server-side due to collision? So message-id is a way of determining if the UDP packet is associated with already existing log or it's for a new log, when we're sending same message-id for multiple logs then server behaviour will be to merge the data-together which will end-up breaking the log)

TBH I am also trying to figure this out, please share any suggestions, am I thinking right? 😅

Licenser:
Ja just the message content would not work, I'm still considering if message content + ingest_ns (nanosecond when the message was registered at tremor) would be enough, if a server produces the same log twice in the same nanosecond that'd be very odd (but not impossible) OTOH having two random generated numbers be the same is also odd (but not impossible) it would also one a more deterministic failure case "When messages with the same content arrive at exactly the same time they will get duplicated message ids" instead of "if the RNG hates you, you'll get duplicated message ids"

Licenser:
Sorry for all the forth and back, so I've been thinking about a good way to make this:

1) fast
2) deterministic
3) non-coliding

and wanted to throw out a suggestion using the ingest_ns + an incremental id as a message ID. This is:

1) fast since we just have to look at two integers, no RNG, no hashing
2) deterministic since the ingest_ns is settable, and the incremental counter is well, incremental so determinstic as well
3) it avoids duplicates by not creating duplicates on the same system due to incremtal id's and makes them extremely unlikely on multiple systems due to nano second timestamps involved.

BharatKJain:
Hmm, I have changed the code to this:
```
(epoch_timestamp & BITS_13) | (data_sha_hash_u64 & !BITS_13) | (thread_id_u64 & BITS_13)
```

epoch_timestamp is `ingest_ns` for process function, for finish function I am taking the current timestamp as there's no `ingest_ns` in the downstream function call.

1. Does the above mentioned approach look good?
2. I didn't understand why are they using `& BITS_13` in the java library, I think if we flip the bits using OR `|`, that should suffice - is that right?
3. I have also added a unit-test as well as removed id from the struct for Gelf - I didn't understand if there's a need to add id in that struct, I cannot find code where we are preserving any message-id that's coming from downstream, please check this as well.

BharatKJain:
> Sorry for all the forth and back, so I've been thinking about a good way to make this:
> 
>     1. fast
> 
>     2. deterministic
> 
>     3. non-coliding
> 
> 
> and wanted to throw out a suggestion using the ingest_ns + an incremental id as a message ID. This is:
> 
>     1. fast since we just have to look at two integers, no RNG, no hashing
> 
>     2. deterministic since the ingest_ns is settable, and the incremental counter is well, incremental so determinstic as well
> 
>     3. it avoids duplicates by not creating duplicates on the same system due to incremtal id's and makes them extremely unlikely on multiple systems due to nano second timestamps involved.

Can I keep thread-id just to reduce the collision probability more? 😅

BharatKJain:
Okay, I have made the changes as suggested in the latest comment.

```
(epoch_timestamp & BITS_13) | (auto_increment_id & !BITS_13) | (thread_id_u64 & BITS_13)
```

- Added a unit-test for it.
- Renamed id to auto_increment_id


Let me know your thoughts!

Licenser:
I would remove the thread ID. It is breaking the distribution and making the date less unique also it prevents event id's to be replayable. Basiclly it will result in the lower 13 bit to ave 3 times as many 1's as 0's making them more likely to collide while also making the id not determinstic.

The second problem I spot is that by removing the lower 13 bit from increment, that part will have no effect for the first 8192 messages and then only change every 8192 messages. My suggestion would be to shift it by 13 bits instead of truncating them.
 
Lastly I'd probably pull in a few more bits form the timestap, 13 bit are only `0,0000082 secs` even if you bump it to 16 that means you get a window of `0,000066 secs` + a 48bit counter (or 281.474.976.710.656 events to count)
 
You'd end up with something like this:


```rust
(epoch_timestamp & 0xFF_FF) | (auto_increment_id << 16 )
 ``
 
 do you think that would solve the original problem?

**Review by Licenser — APPROVED** [review]
wooh wooh

**Comment by BharatKJain** [comment]
I will fix the clippy checks and DCO, please allow me sometime.


---

Honestly I didn't anticipate this to happen but I also had built otel gelf-exporter. :)

[Opentelemetry PR for gelf-exporter (WIP)](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35820)

**Comment by BharatKJain** [comment]
Line 190:

```
let current_epoch_timestamp = u64::try_from(SystemTime::now().duration_since(UNIX_EPOCH).expect("SystemTime before UNIX EPOCH!").as_nanos())?;
```

Not sure how to handle this better, is this okay?

Licenser:
We'd want to avoid the expect. Let me explain why.

`expect` and `unwrap` cause a crash, in some applications that's okay since, say you have a CLI if something goes wrong you might just want to error out and restart from scratch. Now tremor is a bit more complex there might be more then one pipeline running so we don't want one pipeline affect the progress of another. If we'd crash in pipeline A we'd take pipeline B down with it - that's not desirable. So we handle errors and let them bubble up so only the pipeline causing the issue is affected.

To do that there are a few ways `?` in many places work, but if the error type can't be mapped something like `map_err(...)?` is a good alternative.

BharatKJain:
@Licenser @darach Does it look good now?

(I have fixed the DCO, format, clippy-check, code quality checks)

darach (approved):
LGTM 🚀

PR #2671, opened 2024-10-22
Add graphite plaintext codec

darach:
# Pull request

## Description

Adds the graphite line oriented plaintext protocol as used by graphite, the carbon-c-relay and carbon-relay-ng.

## Checklist

* [ x ] The code is tested

## Performance

Minimally functionally tested via `carbon-c-relay` and `carbon-relay-ng`

### DIFF
--- a/tremor-codec/src/codec/graphite.rs
+++ b/tremor-codec/src/codec/graphite.rs
@@ -1,0 +1,188 @@
+// Copyright 2024, The Tremor Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! The `graphite-plaintext` codec supports the [graphite plaintext format](https://graphite.readthedocs.io/en/latest/feeding-carbon.html#the-plaintext-protocol).
+//!
+//! The format is similar to the `influx` line protocol.
+//!
+//! The codec translates a single `graphite` measurement line into a structured event.
+//!
+//! ## Example
+//!
+//! The native format of a single graphite event line is as follows:
+//!
+//! ```text
+//! <metric_path> <value> <timestamp>
+//! ```
+//!
+//! The equivalent representation as a tremor value:
+//!
+//! ```json
+//! {
+//!   "metric": "my.metric.path",
+//!   "value": 7,
+//!  "timestamp": 1620649445 // number of seconds since the epoch
+//! }
+//! ```
+//!
+//! ## Supported types
+//!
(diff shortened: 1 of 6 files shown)


tremor-codec/src/codec/graphite.rs
```diff
@@ -0,0 +1,243 @@
...
+            let decoded = codec.decode(&mut bad, 0, Value::null()).await;
+            assert!(decoded.is_err());
+
+            let mut bad = b"metric.no.measure.no.timestamp".to_vec();
+            let decoded = codec.decode(&mut bad, 0, Value::null()).await;
+            assert!(decoded.is_err());
+
+            let mut bad = b"metric.with.measure.no.timestamp 320.0".to_vec();
+            let decoded = codec.decode(&mut bad, 0, Value::null()).await;
+            assert!(decoded.is_err());
```
Licenser, line 219: Based on the documentation this should parse

> //! If timestamps are not provided, the codec will use the current time as the event's timestamp.

darach: I'll have to reread and check as there's precious little. You might be right though assuming it's the same as -1 and gets current time

Licenser: Ja, at least from the docs I'd expect no timestamp acting the same as -1

darach: I think that's another vestigial thing as I used statsd as a base to get things close. The graphite docs don't have this
behaviour though at all from my reading. So the comment was false, but the test correct. I should have removed the comment when I added the test:

https://graphite.readthedocs.io/en/latest/feeding-carbon.html#step-3-understanding-the-graphite-message-format

I'll update accordingly

Licenser: Agreed, the docs were off the test + code are <3

darach: Pushing now. Well spotted either way!

tremor-codec/src/codec/graphite.rs
```diff
@@ -0,0 +1,188 @@
...
+//! <metric_path> <value> <timestamp>
+//! ```
+//!
+//! The equivalent representation as a tremor value:
+//!
+//! ```json
+//! {
+//!   "metric": "my.metric.path",
+//!   "value": 7,
+//!  "timestamp": 1620649445 // number of seconds since the epoch
```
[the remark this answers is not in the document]
Licenser: looking at the code multiplying makes sense, if -1 is passed it already uses `ns` instead of `s`

```rs
    if "-1" == ts {
        m.insert("timestamp".into(), Value::from(ingest_ns));
    } else {
        let ts = lexical::parse::<u64, _>(ts)?;
        m.insert("timestamp".into(), Value::from(ts));
    }
```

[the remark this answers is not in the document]
darach: I'll do this separately as it deserves changes to the tests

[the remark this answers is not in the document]
darach: I also added comments as this isn't obvious to anyone casually looking at this. Especially our future selves...

Licenser (approved):
<3 awesome addtion!

tremor-codec/src/codec/graphite.rs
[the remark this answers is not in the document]
Licenser: Good call, it'll help us down the road :D

Issue #2701, opened 2024-12-20
Fix broken connectors link in README.md


stan-dot:
Describe the problem you are trying to solve
The link for connectors in the current REAMDE.md is broken. (https://www.tremor.rs/docs/artefacts/connectors)

Describe the solution you'd like
Update the link to the current connectors page.

devskillz47:
i think documentation might have moved to a separate project (https://www.tremor.rs/) and not sure if this file is still maintained any longer. For this issue, please see https://www.tremor.rs/docs/0.12/reference/connectors. 
(on the website, you can also choose different documentation versions depending on what you're interested in)

devskillz47:
this should have now been fixed, together with some other invalid references which have been caught. please verify and close the issue if requirements are satisfied. :p

Licenser:
Yay, thanks!

PR #2702, opened 2024-12-21
Fixing broken links described in issue#2701

devskillz47:
# Pull request
Fix "broken" links included in the Issue#2701(https://github.com/tremor-rs/tremor-runtime/issues/2701)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,8 +70,8 @@ We currently do not recommend tremor where:
 
 - Your event structure is not mappable to a JSON-like data structure.
   - If in doubt, please reach out and create a ticket so we can assist and advice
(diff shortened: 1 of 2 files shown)


Licenser (changes requested):
Can we link this to edge instead of 0.12:

1) edge will not go out of date as a long
2) edge effects what is directly in github so it is the more correct link I think

devskillz47:
will do, but took a little break for the holidays... :p

Licenser:
All good, I just wanted to track it  before I forget :)

devskillz47:
so, i had a more in depth look at this. it seems that currently everything points to v0.12. my only concern in moving it to edge would be that if in the installation instructions we point to the latest release (in this case v0.12) and in the documentation we point it to edge the features described in the documentation might not be there for the users who followed the instructions which could be quite cumbersome.

Licenser:
That's a fair point, then lets just make sure we update the version in the README as we release!

Licenser (approved):

PR #2703, opened 2024-12-31
Fix RUSTSEC-2024-0402

rex4539:
https://rustsec.org/advisories/RUSTSEC-2024-0402


Licenser:
Hi, thanks! CI has some problems we got to resolve before merging this :) but that's a us-problem not yours.

Licenser (approved):

